PR #19, opened 2017-09-14
Initial version of LibOVR/Oculus SDK backend.

shakesoda:
It's enough to work with here. I feel like the GL's a bit uglier than it needs to be, but trying to do it the way OpenVR did wasn't working out for me.

Also includes my attempt to make CMake set the project to Windows subsystem, which seems to get filtered out (!?).

This is disabled by default and requires some manual tweaking of the generated project to build, and downloading your own LibOVR and putting it somewhere. I'd like if someone with better CMake-fu could help make that better.

Tested working:

- Head tracking (position and orientation)
- Touch positions (position and orientation)
- Rendering to headset (converges correctly, latency is good, etc...)
- Rendering to mirror window
- Quits when Oculus Home tells it to
- Recentering
- Clip distances

Not tested, but implemented:

- Controller buttons (isDown) and press/release events
- Controller axes
- Controller touches
- Boundary width/depth
- Headset velocity and angular velocity

NYI:

- MSAA
- Controller add/remove events (currently always reports both, although isPresent is implemented)
- Controller models
- Boundary geometry
- Haptics
- Proper handling of errors such as display lost
- Other supported controller types than Touch (e.g. Oculus remote, xbox controller...)

Known issues:
- Something is weird with transforms causing inconsistent shading with OpenVR (testable with [the example shader](http://lovr.org/docs/Lighting) and a hand model. Help wanted, I am stumped.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -274,8 +274,11 @@ set(LOVR_SRC
   src/util.c
 )
 
+set(LOVR_USE_OCULUS OFF CACHE BOOL "")
 if(EMSCRIPTEN)
   set(LOVR_HEADSET src/headset/webvr.c)
+elseif(LOVR_USE_OCULUS)
+  set(LOVR_HEADSET src/headset/oculus.c)
 else()
   set(LOVR_HEADSET src/headset/openvr.c)
 endif()
@@ -287,6 +290,7 @@ endif()
 
 include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src)
 add_executable(lovr ${LOVR_SRC} ${LOVR_HEADSET})
+
 target_link_libraries(lovr
   ${LOVR_ASSIMP}
   ${LOVR_ENET}
@@ -302,6 +306,11 @@ target_link_libraries(lovr
 
   ${LOVR_EMSCRIPTEN_FLAGS}
 )
+
+if (WIN32)
+  set_target_properties(lovr PROPERTIES LINK_FLAGS "/SUBSYSTEM:WINDOWS")
+  set_target_properties(lovr PROPERTIES LINK_FLAGS "/ENTRY:mainCRTStartup")
+endif()
 
 if(WIN32)
   function(move_dll ARG_TARGET)
--- a/src/headset/oculus.c
+++ b/src/headset/oculus.c
@@ -1,0 +1,531 @@
+#include "headset/headset.h"
+#include "event/event.h"
+#include "graphics/graphics.h"
+#include "graphics/texture.h"
+#include "lib/vec/vec.h"
+#include "math/mat4.h"
+#include "math/quat.h"
+
+#include <stdbool.h>
+#include <OVR_CAPI.h>
+#include <OVR_CAPI_GL.h>
+
+typedef struct {
+  bool isInitialized;
+  bool isRendering;
+  bool isMirrored;
+  bool hmdPresent;
+  headsetRenderCallback renderCallback;
+  vec_controller_t controllers;
+  ovrSession session;
+  ovrGraphicsLuid luid;
+  ovrTrackingState ts;
+  ovrInputState is;
+  int frameIndex;
+  float clipNear;
+  float clipFar;
+  int lastButtonState;
+  struct {
+    ovrSizei size;
+    GLuint fboId;
+    GLuint depthId;
+    ovrTextureSwapChain chain;
+  } eyeTextures[2];
+  ovrMirrorTexture mirrorTexture;
+  GLuint mirrorFBO;
+} HeadsetState;
+
+static HeadsetState state;
+
+void lovrHeadsetInit() {
+  vec_init(&state.controllers);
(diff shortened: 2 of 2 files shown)


src/headset/oculus.c
```diff
@@ -0,0 +1,531 @@
...
+    ovrTextureSwapChain chain;
+  } eyeTextures[2];
+  ovrMirrorTexture mirrorTexture;
+  GLuint mirrorFBO;
+} HeadsetState;
+
+static HeadsetState state;
+
+void lovrHeadsetInit() {
+  vec_init(&state.controllers);
```
bjornbytes: This is here and on 67, can probably just remove this

src/headset/oculus.c
```diff
@@ -0,0 +1,531 @@
...
+      e.type = EVENT_CONTROLLER_RELEASED;
+      e.data.controllerreleased.controller = controller;
+      e.data.controllerreleased.button = target;
+    }
+    lovrEventPush(e);
+  }
+}
+
+void lovrHeadsetPoll() {
+  ovr_SetTrackingOriginType(state.session, ovrTrackingOrigin_FloorLevel);
```
bjornbytes: Any reason this is being done every frame instead of just once?

src/headset/oculus.c
```diff
@@ -0,0 +1,531 @@
...
+  state.ts = ovr_GetTrackingState(state.session, 0, true);
```
bjornbytes: Since we don't use the tracking state to emit any events, is it possible to be more lazy here with respect to the tracking state and fetch it in `lovrHeadsetGetPosition` etc.?

The event pump isn't really meant to be used for general updating (though it's currently the only place to do update-type stuff).

src/headset/oculus.c
```diff
@@ -0,0 +1,531 @@
...
+    lovrEventPush(e);
+  }
+
+  //if (!status.HmdMounted) // update mirror only?
+  //ovr_SetInt(state.session, OVR_PERF_HUD_MODE, (int)ovrPerfHud_PerfSummary);
+
+  state.hmdPresent = (status.HmdPresent == ovrTrue)? true : false;
+
+  Controller* left = state.controllers.data[ovrHand_Left];
+  Controller* right = state.controllers.data[ovrHand_Right];
```
bjornbytes: I think this is only guaranteed to work if `ovrHand_Left` is 0 and `ovrHand_Right` is 1.

src/headset/oculus.c
```diff
@@ -0,0 +1,531 @@
...
+    if ((state & button) > 0) {
+      e.type = EVENT_CONTROLLER_PRESSED;
+      e.data.controllerpressed.controller = controller;
+      e.data.controllerpressed.button = target;
+    }
+    else {
+      e.type = EVENT_CONTROLLER_RELEASED;
+      e.data.controllerreleased.controller = controller;
+      e.data.controllerreleased.button = target;
+    }
```
bjornbytes: Minor, a ternary could be used for `e.type` instead of the if.

src/headset/oculus.c
shakesoda: ovrHand_Left and _Right are defined per the docs to be used as 0/1 for array access

src/headset/oculus.c
shakesoda: I think you're only supposed to call GetTrackingState once per frame since it does some stuff with latency markers, and beyond that I have no idea how many times per frame a user might call the functions for TS, so it'd end up shuffling a bunch of data around needlessly.

src/headset/oculus.c
shakesoda: No particular reason, I'll put this in init.

src/headset/oculus.c
shakesoda: woops

src/headset/oculus.c
shakesoda: I considered it, but that just adds another branch and swaps a trivial line with a big ugly one, since I can't ternary out the whole block.

bjornbytes: Oh derp, I missed the controllerpressed/controllerreleased difference.

src/headset/oculus.c
shakesoda: checked the docs again and I need to call GetTrackingState with the third argument true once per frame for accurate latency measurements. I was actually using it wrong here to begin with, however, the guide says I should be using predicted time instead of real (0). fixed in latest.

CMakeLists.txt
```diff
@@ -303,6 +307,11 @@ target_link_libraries(lovr
   ${LOVR_EMSCRIPTEN_FLAGS}
 )
 
+if (WIN32)
+  set_target_properties(lovr PROPERTIES LINK_FLAGS "/SUBSYSTEM:WINDOWS")
+  set_target_properties(lovr PROPERTIES LINK_FLAGS "/ENTRY:mainCRTStartup")
+endif()
+
```
bjornbytes: These can be removed now, I've got those added to master (with correct flags for debug/release).

PR #21, opened 2017-09-22
cmakelist: fix msvc defs leaking to other platform

bcampbell:
The MSVC_VERSION check was kicking in on Linux (and presumably OSX),
and causing redefines of `inline` and `snprintf`, and failing the
build with  lots of "undefined reference to `_snprintf'" errors.


bcampbell:
Hmm. My first attempt to fix this seems to have borked the windows build... sigh. I just committed a second go at it. Hopefully it'll pass the CI build this time...

bcampbell:
Ah, maybe that wasn't me breaking the build after all!

PR #22, opened 2017-10-06
add basic mouse support

bcampbell:
I've added a couple of basic functions to support mouse input.
It's all in a new `input` module, as I'm planning to add some keyboard functions too.
Mostly self contained, although I had to add a function to graphics to get at the the GLFW window (which I need to use the GLFW mouse functions).


bcampbell:
Background: I don't actually have constant access to a VR headset and controllers, so I want to be able to hack together a stand-in headset and controller using mouse and keyboard (mouselook for orientation, WASD/cursor keys to move and a 'controller' position and orientation locked to the camera).
There's probably an argument for having a fake headset+controller built in for when a real one isn't available (I think the webvr backend does this somewhat).
But I think general mouse and keyboard (and maybe even game controller) support is a good thing - lovr looks pretty appealing for doing non-VR 3D games too :- )

bjornbytes:
Would it work for you if LÖVR just came with a set of default mouse/keyboard controls (maybe controlled via a `lovr.conf` switch), instead of exposing them in the API?  That would let people use them out of the box without having to write it themselves or use a library, and keeps the API smaller and more focused on VR.

bcampbell:
Yep, that'd work great for me. I just want to be able to develop stuff without necessarily having VR gear to hand. A fake Headset (WASD movement and mouselook) and a fake Controller (notionally just glued to my forehead, so it pointed where I was looking ;- )

That said, I think LÖVR makes a great enviroment for coding up 3d stuff, VR or not.
I can definitely see myself using it for doing non-VR games.
I think the API surface area for input support could be kept pretty small, and an `input` module can always be excluded via conf...

bjornbytes:
Been thinking about this a bit, I think for now it would be best to keep it internal and expose some default mouse/keyboard controls for situations where hardware isn't available.  At some point in the future the mouse/keyboard modules will make more sense but right now it has the potential to distract from the VR focus.  I have mixed feelings about this because exposing the mouse/keyboard API to Lua lets people customize things more and create different types of projects with LÖVR, but that comes at a cost of making the project a little more difficult to maintain going forward.

With that said, I'd be happy to collaborate more on the internal mouse/keyboard input!  I think it would help out a lot of people.

bjornbytes:
It's been a while, but I've created some small external libraries for keyboard/mouse support using the LuaJIT FFI that may be of interest: [lovr-keyboard](https://github.com/bjornbytes/lovr-keyboard) and [lovr-mouse](https://github.com/bjornbytes/lovr-mouse).

PR #24, opened 2017-10-22
Fake headset

bcampbell:
OK, I've merged in all the changes on `master` since I started on the fake-headset stuff. No major conflicts popped up, and it all seems pretty solid. So I'm happy for it to be merged in now, if you think it looks all right.

Definitely a couple of rough edges (I definitely need to work on my fake controller), but I think having fallback mouselook support is good thing (and runtime-selectable headset implementation, which will hopefully be useful for #19).


CMakeLists.txt
```diff
@@ -279,9 +281,7 @@ set(LOVR_SRC
 )
 
 if(EMSCRIPTEN)
-  set(LOVR_HEADSET src/headset/webvr.c)
-else()
-  set(LOVR_HEADSET src/headset/openvr.c)
```
bjornbytes: Hm, shouldn't we actually avoid compiling openvr.c when building for WebVR?  It isn't used there.

bjornbytes:
Do you mind commenting in or removing the sections of code that are commented out?  It's hard to know the reason why they're commented out and if/when it would make sense to comment them back in.

src/headset/fake.c
```diff
@@ -0,0 +1,493 @@
+/* vim: set ts=2 sts=2 sw=2: */
```
bjornbytes: Can you remove these comments?  There's an `.editorconfig` already for this sort of thing.  Looks like they're here and in `headset.c`, `headset.h`, and `openvr.c`.

src/headset/headset.h
```diff
@@ -56,6 +58,48 @@ typedef struct {
...
+    void (*controllerGetPosition)(Controller* controller, float* x, float* y, float* z);
+    void (*controllerGetOrientation)(Controller* controller, float* angle, float* x, float* y, float* z);
+    float (*controllerGetAxis)(Controller* controller, ControllerAxis axis);
+    int (*controllerIsDown)(Controller* controller, ControllerButton button);
+    int (*controllerIsTouched)(Controller* controller, ControllerButton button);
+    void (*controllerVibrate)(Controller* controller, float duration, float power);
+    ModelData* (*controllerNewModelData)(Controller* controller);
+    void (*renderTo)(headsetRenderCallback callback, void* userdata);
+    void (*update)(float dt);
+} HeadsetImpl;
```
bjornbytes: Do you mind renaming this to something without an abbreviation?  Maybe `HeadsetInterface` or something.

src/headset/fake.c
```diff
@@ -0,0 +1,493 @@
...
+  if (glfwGetKey(w, GLFW_KEY_A)==GLFW_PRESS || glfwGetKey(w, GLFW_KEY_LEFT)==GLFW_PRESS) {
+    v[0] = -k;
+  }
+  if (glfwGetKey(w, GLFW_KEY_D)==GLFW_PRESS || glfwGetKey(w, GLFW_KEY_RIGHT)==GLFW_PRESS) {
+    v[0] = k;
+  }
+  if (glfwGetKey(w, GLFW_KEY_SPACE) == GLFW_PRESS) {
+    v[1] = k;
+  }
+  if (glfwGetKey(w, GLFW_KEY_LEFT_SHIFT) == GLFW_PRESS) {
```
bjornbytes: I wonder if this will be annoying in windows due to all the weird sticky keys/filter keys stuff.  If so, maybe it could be switched to Q/E or something.

src/data/boot.lua
```diff
@@ -187,6 +187,9 @@ function lovr.step()
   end
   local dt = lovr.timer.step()
   if lovr.audio then
+    if lovr.headset and lovr.headset.isPresent() then
+      lovr.headset.update(dt)
+    end
```
bjornbytes: I was wondering why mouselook wasn't working on the no game screen, and realized that it's because this is inside the `if lovr.audio` condition, and the nogame screen has the audio module disabled!

bcampbell:
Ha! Sorry about that one.
I got rid of the `isPresent()` check too.

Not sure it's obvious enough in the code, but the new `isAvailable` function is designed to be called to determine if an implementation can be used,  without initialising it to find out. In theory that's separate from whether or not `isPresent` is set... but I'm not sure if `isPresent` would ever return false now. (hot-pluggable headsets maybe? Seems a bit of a fringe case...).

bjornbytes:
Yeah, `isPresent` is getting really crusty.  I think it may end up being renamed to `isTracking` or something, which is what it's more useful for.

bjornbytes:
I'll go ahead and merge this once that last round of conflicts is resolved.  I think (as you hinted at) some sort of driver system in conf.lua that lets you choose which driver(s) you want makes sense, but as-is this is a valuable change that makes it a lot easier to get started with LÖVR!

bcampbell:
Oh - are there still conflicts? There's that message that it can't be rebased, but I think that just means that the two branches touched some of the same stuff and can't be reordered into a nice linear history...

bjornbytes:
Oh, I see.  Merging!

PR #26, opened 2017-10-29
Fake headset - controller button event

bcampbell:
Forgot to send events for the fake controller button.
Also moved the offset for the fake controller, so it's not situated at your feet.

### DIFF
--- a/src/headset/fake.c
+++ b/src/headset/fake.c
@@ -118,11 +118,41 @@ static void cursor_position_callback(GLFWwindow* window, double xpos, double ypo
 
 }
 
+
 static void mouse_button_callback(GLFWwindow* window, int button, int action, int mods)
 {
+
+  // clicking in window grabs mouse
   if (!state.mouselook) {
     if (button == GLFW_MOUSE_BUTTON_LEFT && action == GLFW_PRESS) {
       enableMouselook(window);
+    }
+  }
+
+  // now generate events on fake controllers
+  {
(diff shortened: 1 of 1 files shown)


src/headset/fake.c
```diff
@@ -336,7 +366,14 @@ static ControllerHand fakeControllerGetHand(Controller* controller) {
 
 static void fakeControllerGetPosition(Controller* controller, float* x, float* y, float* z) {
   // for now, locked to headset
-  fakeGetPosition(x,y,z);
+  
+  float offset[3];
+  vec3_set(offset, 0, 0,-1.0f);
```
bjornbytes: Does this put the controller 1 meter in front of you?

bcampbell: yes, it does. I figured that made for more-or-less arms length, depending on your arms :- )

bjornbytes: Okay!  Just checking.  I'm a little weary at how rigid some of the fake headset stuff is, but I guess that was one of the tradeoffs of doing it as opposed to the raw keyboard/mouse APIs...

bcampbell: Yes. it'd be nice to be able to configure some of this stuff in `conf.lua`, but I'm wary of adding too much noise.
(eg I'd like to be able to set up two fake controllers - one offset to the left and one offset to the right, but it seems like a pretty niche requirement)

PR #29, opened 2018-01-03
--live argument automatically reruns when lua files change [rfc]

mcclure:
This patch adds a --live argument to lovr (like, `lovr --live path/to/module`) which places all `require`d files, as well as any files designated `lovr.filesystem.reloadAddWatch`, on a watch list. If any file on the watch list changes, or `lovr.filesystem.reloadReset` is called, the currently running lua script shuts down and is restarted from the beginning, but the executable doesn't close and the window stays open. I find this mode of working much more convenient— i can just change a couple lines of code, save, pull my headset back over my head and see my change.

Current limitations/work left undone:
- This touches a LOT of different parts of the code!! It even changes main.c. I'm not sure I architected this the way the project would prefer. Bjorn implied he would rather do this in-lua than use the C code (although one thing I like about this approach is the old and new lua are environments are isolated)
- Not documented
- It doesn't work in emscripten mode, so it should print an error if you try to run it in emscripten mode.
- I would like to have there be `lovr.willReload()` and `lovr.didReload(state)` callbacks that can be used to pass data from one run to the next. For example if the user has warped to coordinate <3,4,24> it should be possible to have `willReload` return `{x=3, y=4, z=24}` and `didReload` receive this coordinate so that when the world restarts the user is in the same place they were before.
- It does not work when an error occurs, when you enter the handleError screen . The problem is there are a couple ways to get into `handleError` so I'm not quite sure yet how to best architect getting back out again.

I am not sure this is ready to be checked in but I thought I'd post it to see what you think of the general direction. What do you think?


bjornbytes:
There's a lot of interesting stuff here!  My initial reaction: LÖVR should expose a way to restart itself quickly and gracefully, that's super useful.  The "when to reload and how to watch files" part is trickier, and I think we should be as unopinionated as possible about it.  Need to let it simmer more though.  I :purple_heart: the `lovr.willReload` and `lovr.didReload` callbacks, those are a neat idea.

mcclure:
The way I have implemented this definitely is based around the assumption of a specific use case (you are debugging, and you want to forcibly reload everything cleanly). This leads to things like emscripten not being supported, `reloadReset()` refusing to work when called but you're not in `--live` mode, etc.

This said, if you were restarting for some reason *other* than debugging, you might not want it to work the same way… this code goes to a great deal of trouble to make the environments pre- and post- restart be totally isolated, but if you are restarting for a non-debugging reason this might not be what you want.

bjornbytes:
Superseded by https://github.com/bjornbytes/lovr/pull/32

PR #31, opened 2018-01-13
Set glfw time to 0 during init.

mcclure:
This was something I did in my live-reload branch (where prevents a large erroneous negative dt on first update). However, it also helps if lovr is launched in a debugger such as lldb which pauses at boot (it prevents a large erroneous positive dt on first update).


mcclure:
Wait, there's already a SetTime(0)

PR #32, opened 2018-01-21
Restart-without-quit support: lovr.event.quit("restart")

mcclure:
This is an alternative to #29. lovr.event.quit and return-from-run() both support the string "restart" in place of an exit code. When this happens, the program reruns itself in a new, isolated environment. In Emscripten, no reboot occurs and the program only quits. If either receive an unexpected value (something other than a number, nil or a string other than "restart") there is an error.

I am making this change with the intent of enabling a feature where lovr auto-restarts on file changes (either in a future PR or in a personal fork I use).

Here is a test program I use to verify it is working.

```
local total = 0

print("BOOT")

function lovr.update(dt)
	total = total + dt
	if total > 5 then
		print("QUIT")
		lovr.event.quit("restart")
	end
end

function lovr.draw()
  lovr.graphics.print('oh wow', 0, 2, -3, .5)
end
```

Concerns I have about the current code:

- There is a mix of init functions handling the reboot by simply tolerating multiple calls and discarding repeat calls; and special-checking the `lovrReloadPending` variable. I am uncertain the `lovrReloadPending` variable should exist (it is an artifact of how #29 did things) and maybe it should be removed.
- Some code duplication between lovrRun and l_lovrEventQuit
- I think there should be some way for a lovr builder/invoker to disable restarts (in case it is some sort of security risk for example?)
- In my recent tests on Windows, when I quit I get an "openAL not shut down!" error. I do not know if this is because of this patch or not and I do not have access to my Windows machine today.

### DIFF
--- a/src/api/event.c
+++ b/src/api/event.c
@@ -15,7 +15,10 @@ static int nextEvent(lua_State* L) {
   switch (event.type) {
     case EVENT_QUIT: {
       lua_pushstring(L, "quit");
-      lua_pushnumber(L, event.data.quit.exitCode);
+      if (event.data.quit.restart)
+        lua_pushliteral(L, "restart");
+      else
+        lua_pushnumber(L, event.data.quit.exitCode);
       return 2;
     }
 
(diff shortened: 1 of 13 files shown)


bjornbytes:
It would be great if audio, filesystem, graphics, and luax didn't need to include `lovr.h`.  I've been trying to avoid that so that they are more modular and able to exist on their own.

bjornbytes:
Putting `lovrReloadPending` in `util` might be enough for this, but like you said it might be better to just have all the init functions consistently store that state in a static bool local to the file.

mcclure:
I think the only thing they use lovr.h for is `lovrReloadPending`. `lovrReloadPending` could be removed or moved.

bjornbytes:
Thanks!

PR #33, opened 2018-01-26
remove webvrGetBoundsGeometry reference from webvr.c

esno:
getBoundsGeometry implementation was dropped with 11586e8
but not in HeadsetInterface.


bjornbytes:
Thanks

PR #34, opened 2018-01-28
Live-reload mode: Restart automatically triggers on source changes

mcclure:
Last I talked to Bjorn this was not his preferred approach to get this functionality. The solution here is pretty hardwired to this specific feature. Another possibility we discussed was to put this entirely in boot.lua and use package.loaded instead of the require() hook, and possibly have some kind of a lovr.filesystem hook for watching a pool of files for changes efficiently (say using OS-specific features like fsevents instead of a linked list as is the case here).

Anyway, here's a branch with a `--live` feature for autoreloading sources when you edit them.


mcclure:
Just updated this PR with (1) rebased against lovr master as of today (2) autoreload works even at the errhand() screen

bjornbytes:
Hey what do you think about putting this feature in a new `lovr.dev` module?  I'd be up for merging it like that, can also help with creating the module, rebasing, etc. if you want.

bjornbytes:
Should this be closed now that lodr exists?

mcclure:
Yeh

PR #35, opened 2018-01-31
Error message instead of crash on invalid audio.newSource call

mcclure:
Checks for file not found and file-not-understood cases.

### DIFF
--- a/src/api/audio.c
+++ b/src/api/audio.c
@@ -70,6 +70,10 @@ int l_lovrAudioNewSource(lua_State* L) {
     Blob* blob = luax_readblob(L, 1, "Source");
     stream = lovrAudioStreamCreate(blob, 4096);
     lovrRelease(&blob->ref);
+    if (!stream) {
+      luaL_error(L, "Could not decode Ogg audio source at '%s'", luaL_checkstring(L, 1));
+      return 0;
+    }
   }
 
   Source* source = lovrSourceCreate(stream);


src/api/audio.c
```diff
@@ -68,8 +68,14 @@ int l_lovrAudioNewSource(lua_State* L) {
     stream = *type;
   } else {
     Blob* blob = luax_readblob(L, 1, "Source");
+    if (!blob) // readblob responsible for setting Lua error
+      return 0;
```
bjornbytes: I don't see how this is ever true?  Looking at readblob, it should either return a Blob or throw an error.

mcclure: So I guess the correct thing to check is if (!blob->data) ?

src/api/audio.c
```diff
@@ -68,8 +68,14 @@ int l_lovrAudioNewSource(lua_State* L) {
     stream = *type;
   } else {
     Blob* blob = luax_readblob(L, 1, "Source");
+    if (!blob->data) // readblob (1) never returns NULL (2) is responsible for setting Lua error
+      return 0;
```
bjornbytes: I'm still confused: given `readblob` either returns a Blob or throws an error, we shouldn't need to check the return value and could just omit this `if`.

mcclure: You are right, I messed up in my initial testing and thought this was necessary.

PR #36, opened 2018-02-01
Support for 3DConnexion Spacemouse

mcclure:
…which is this thing https://www.youtube.com/watch?v=GIi1s1dNaCg&t=198s

Run cmake with -DSPACEMOUSE_SUPPORT=1 and the fake.c simulator will use the 3DConnexion SpaceMouse in addition to KBAM.

Problems:
1. Currently only works on mac
2. When enabled, will only compile if you have the SpaceMouse drivers installed (ie 3DconnexionClient.framework).
3. I *think* this will produce binaries that only work on systems that have the SpaceMouse drivers installed
4. The way spacemouse.c integrates into fake.c is a little ad hoc? Not sure this was the best way to do it.
5. -DSPACEMOUSE_SUPPORT builds are not tested in appveyor

I will look at Windows support later but it's a pain because the Mac, Windows and Linux SDKs for the SpaceMouse are all *completely* different. I would also like to add support for controlling fake handsets.

I think if we can fix 3 and 5 and get the SpaceMouse SDKs put on your builders, it might be worth enabling this in the release builds of Lovr.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -343,3 +343,11 @@ if(WIN32)
   move_dll(${LOVR_OPENVR})
   move_dll(${LOVR_PHYSFS})
 endif()
+
+# 3DConnexion SpaceMouse support-- build if you want to locally debug with the fake-movement simulator using a SpaceMouse
+# Currently this requires you to be on Mac OS X, and to have installed the SpaceMouse drivers before you build.
+if(SPACEMOUSE_SUPPORT)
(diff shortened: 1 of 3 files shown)


bjornbytes:
I think this would be a good candidate for an external LuaJIT FFI module, given that it seems like a pretty niche use case and depends on (I think) a closed source library.  There's an example of the FFI stuff [here](https://github.com/bjornbytes/lovr-mouse).

PR #37, opened 2018-02-02
Export texture to PNG [rfc/dubious]

mcclure:
Adds a texture:exportFile(tagString) method. Also works on canvases. The file name is tagString plus ".png", with a number before .png if necessary to avoid overwriting an existing file.

Image export handled by STB. This library involves no license requirements.

This is probably not the right way to do it:
- Operating systems have preferred locations for saving data. Saving to PWD on Mac OS X instead of ~/Documents or whereever may result in the file just being deleted
- This should probably (?) be in data and not in the texture object
- It would be good to have a way to do this for "last rendered frame" in addition to a canvas/texture
- STB may not be the best way to do image export (the indirect dependencies for lovr already contain AssIMP and LodePNG)

### DIFF
--- a/src/api/types/texture.c
+++ b/src/api/types/texture.c
@@ -62,6 +62,14 @@ int l_lovrTextureSetWrap(lua_State* L) {
   return 0;
 }
 
+int l_lovrTextureExportFile(lua_State* L) {
+  Texture* texture = luax_checktypeof(L, 1, Texture);
+  const char *str = luaL_checkstring(L, 2);
+  if (!lovrTextureExportFile(texture, str))
+    luaL_error(L, "Failed to write PNG file");
+  return 0;
+}
+
 const luaL_Reg lovrTexture[] = {
   { "getDimensions", l_lovrTextureGetDimensions },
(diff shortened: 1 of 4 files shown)


mcclure:
Build failing because my method of checking "does file exist" apparently isn't windows portable

bjornbytes:
- For save location, I think the save directory is the best place for this.  It'll handle all the cross platform stuff for you, and you can offload a lot of work over to lovrFilesystem.
- I do think this should go in the `lovr.data` module.  It feels more data-y than graphics-y, and having it in `lovr.data` has the added bonus that we can export textures from background threads.  Example API: `blob = TextureData:encode('filename.png' | nil)`.
- The "last rendered frame" feature could be `lovr.graphics.newScreenshot` like LÖVE has.  Could make more sense for it to go in `lovr.headset` and have more options for stereo/mono/cubemap screenshots though.
- STB seems good for this, but please put it in `src/lib/stb`.

bjornbytes:
There's now a `TextureData:encode(filename)` function.

mcclure:
Cool! I'm gonna close this until I can work out the best way to proceed.

PR #40, opened 2018-03-03
Fix linux build

incinirate:
On (arch) linux, compilation would not complete due to the following error:
```/usr/bin/ld: CMakeFiles/lovr.dir/src/lib/tinycthread/tinycthread.c.o: undefined reference to symbol 'pthread_mutexattr_settype@@GLIBC_2.2.5'
/usr/lib/libpthread.so.0: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status```

Adding `-lpthread` to the linker flags fixes this


bjornbytes:
Thanks!

PR #45, opened 2018-04-22
Fix controller button detection for linux and mac

johnoliverdriscoll:
There seems to be a memory alignment issue on linux and mac that causes every button press to be reported as 'system'.
Please see this SteamVR issue for reference:
https://github.com/ValveSoftware/SteamVR-for-Linux/issues/35


bjornbytes:
Thanks

PR #52, opened 2018-08-20
Introduce -DUSE_OPENAL=0 CMake option which prevents inclusion of OpenAL

mcclure:
When cmake is run with -DUSE_OPENAL=0 with this patch, OpenAL is not linked and all audio functions are replaced with dummy versions which compile but do not work. This is handy for me because I like to link in PortAudio in place of OpenAL and write custom audio code against it. It could also be a useful stopgap in the unlikely circumstance there is a platform where OpenAL does not work.

This may not be the best way to handle this because when this is set, the results of calls to lovr.audio could possibly crash (the stubs sometimes return NULL). Other possible solutions include
- Remove lovr.audio from the lua namespace so that audio-using programs fail cleanly
- Do something elaborate so that the lovr.audio calls always appear to succeed but no audio plays— no idea how you'd test this
- Bjorn has selected making some of the lovr modules, such as audio, separate DLLs or something
What do you think?


bjornbytes:
I had an idea about this: What if all the dummy audio functions were put into a file called `audio/null.c` or something, and then in the CMakeLists we just said something like

```
if(USE_OPENAL)
  set(LOVR_SRC ${LOVR_SRC} src/audio/audio.c src/audio/source.c src/audio/microphone.c)
else()
  set(LOVR_SRC ${LOVR_SRC} src/audio/null.c)
endif()
```

This is similar to how we conditionally include C files for the different headset implementations (see "Conditional sources" section of CMakeLists).  I think that would help avoid ifdefs in the audio code.  Thoughts?  I could also merge now and make a todo task for it.

mcclure:
That would be a perfectly good solution BUT, there would still need to be #if USE_OPENAL switching in the header files because openal types are used in the header. Is that ok?

mcclure:
Updated with bjorn's suggestion

PR #53, opened 2018-09-19
Make CMakeLists more modular;

bjornbytes:
- Enable/disable individual modules, including enet/json
- Enable/disable different "backends" (audio/headset libraries, lua vs. luajit)
- Choose to use system vs. source versions of different libraries

Trying to find ways to avoid so many ifdefs.

Still need to test on Windows.

Also want to make sure this is actually useful and not overly modular.


mcclure:
I like this change... how hard would it be to have a switch that would *exclude* assimp? (It's big and it's hard to build... well, I had trouble getting it to build on Android, anyway)

mcclure:
Awesome

PR #61, opened 2018-09-25
Replace luadist luajit with luajit from luvit (DO NOT MERGE)

mcclure:
The benefits of this are (1) easier to patch luajit cmake file if there are problems with it; (2) luadist luajit is kind of old and has known bugs

Source https://github.com/luvit/luv commit 26cb7dc50da, cmake file hosted in-repo, luajit itself from main tree commit b025b01

Outstanding matters before merge:
- Should I do this for Lua too?
- What does EXCLUDE_FROM_ALL do? I removed some targets from the EXCLUDE_FROM_ALL list that aren't in the new luajit CMakeLists, BUT i didn't add any new targets that new-luajit CMakeLists happens to include.


mcclure:
Also probs should be tested on mac before merge (i have tested on windows)

mcclure:
Note, this is currently blocked on a request by bjorn that instead of changing the directory structure we create a luajit.cmake. This is actually a very good idea because it reduces weird merge conflicts when changing commits…

bjornbytes:
On a different LuaJIT fork now that supports Android, this can be closed now.

PR #62, opened 2018-09-27
Dynamic modules

bjornbytes:
With this change the main runtime doesn't know anything about the Lua API or the set of modules and instead loads things dynamically using conventions similar to Lua.

- The runtime is now able to find and load modules when they're required instead of loading a hardcoded list at startup.
- Modules destroy themselves using the new `luax_atexit` to avoid a hardcoded destroy-list.
- Rename module entry points to be of the form `luaopen_lovr_<module>`.
- Separate common api helper functions into individual headers.
- Move lovr.c stuff into main.c since it's rather small now.


bjornbytes:
Smart people have informed me that this approach won't work on iOS or consoles.

PR #65, opened 2018-10-09
Curve;

bjornbytes:


bjornbytes:
It may be useful to add the ability to get normal/tangent for a point on the curve.

PR #66, opened 2018-10-25
Use alpha to coverage and alpha test for text;

bjornbytes:
This solves problems with transparency and depth testing when rendering fonts, but subtly changes the appearance of rendered text.  I would say that text seems to be a little more "shimmery" when viewed from far away now.  My guess is that the font shader is not antialiasing properly.


bjornbytes:
Should this be exposed as graphics state?

PR #68, opened 2018-10-26
Add LOVR_USE_LOGO option;

bjornbytes:
Saves 20kb, may be worth for web build, or for "release" builds that don't need the logo.


bjornbytes:
Closing this, I think the nogame screen is a little too ugly without the logo.  It would be cool if it was rendered with a shader instead of a raster image.

PR #69, opened 2018-10-28
Oculus Mobile support

mcclure:
Supports Oculus Mobile SDK (ie: Samsung Gear, Oculus Go, possibly Oculus Quest…?).

You build with `LOVR_USE_OCULUS_MOBILE` in CMake for this support, but, realistically you cannot build the oculus mobile support by itself. Instead, you want to build an Android shim app, which I am calling LovrApp, and which you can find here https://github.com/mcclure/lovr-oculus-mobile . In fact, two files and an include are expected to be found OUTSIDE of the Lovr directory (../BridgeLovr.cpp -- LovrApp includes Lovr as a git submodule). Because of the oddness of the Android app build system and the Oculus libraries, I think it makes the most sense to use this "separate app repo, submodule" architecture and not attempt to put the Android app inside the lovr repo-- **but** it would be possible to move BridgeLovr into the lovr repo, if that would make more sense.

This is a first pass (I will do a line-by-line self-audit tomorrow) and has various outstanding known issues:

## TODO — MUSTFIX BEFORE MERGE

* ~~Something is wrong with the Android lifecycle management. If you quit and then reopen it, on the second pass it crashes. (It then offers to let you reboot it, and on the pass after THAT it succeeds). I think the fix for this is to support something like the "restart" mode.~~

* ~~Controller support~~

* ~~OpenAL not tested~~

* ~~Player "tallness" is wrong?~~

* Syntax error during draw() currently results in hard crash

## TODO — NICETOHAVE, COULD DO AFTER MERGE

* PhysFS is doing something "wrong" currently. Right now, when the app wants to run it copies all of its contents out of the APK into internal storage. This serves two purposes (1) PhysFS does not currently support mounting a subdirectory of a zip so this is necessary to use PhysFS while preserving the paths lovr expects (2) this makes it possible to `adb push` files into the directory for faster debugging (the apk-copy saves a timestamp and only copies when the apk changes). This is inadequate for two reasons. (1) In the two weeks since I wrote this code, PhysFS added the missing feature (because I asked for it).  (2) It turns out the permissions in Android are really weird, and you can't `adb push` to internal storage, only to `/sdcard`. You have to do some annoying thing like `adb push` to /tmp and then copy in an `adb shell`. So a better way to do all this would be (1) Use physfs instead of copying out (2) make the `adb push` work via something like automatically checking `/sdcard/lovr-overwrite/com.lovr.appsample` for existence (maybe only when it is a debug build?)

* ~~OpenAL should be switched to https://github.com/takeitallsource/openal-android-sample , I used Apportable's old Android OpenAL but this is abandoned and this other openal is newer~~

* Currently uses Lua rather than Luajit ( this is nontrivial to fix and might require https://github.com/bjornbytes/lovr/pull/61 -- the problem is Luajit builds two helper tools during its own build, but this does not work in a cross compile environment)

* `getEyePose` in oculus mobile driver is wrong (just redirects to `getPose`)

* Avatar SDK support (to display controller) (this would work in Oculus desktop too!)

* Would be nice if liblovr.so were a static rather than a dynamic lib

## TODO — MINOR

* The oculus sample app I based LovrApp on only supports one-eye-at-a-time drawing OR multiview, so right now it draws one eye at a time. It should either draw stereo, or we should add multiview support to Lovr and use multiview.

* Currently it doesn't/can't build in Android Studio?

* No debugging (that is, lldb) instructions in LovrApp README

* No Windows build instructions in LovrApp README

* It would be SUPERCOOL if there were an official lovrapp-debug binary that you are intended to `adb push` something into `/sdcard/`, so that people could do development without having to build anything

* lovr.conf msaa is ignored

* Issue "focus" events on Android pause/resume?


src/graphics/opengl.c
```diff
@@ -135,6 +135,7 @@ struct Canvas {
   int attachmentCount;
   bool needsAttach;
   bool needsResolve;
+  bool immortal;
```
mcclure: This is meant to communicate "If true, framebuffer resource lifetimes are externally managed, if false, framebuffer resources are managed by Lovr and Lovr will delete framebuffers when the object is deleted".

mcclure:
"postprocessing" demo from examples does not work

mcclure:
ran a sound test, OpenAL appears to not work

src/headset/oculus_mobile.c
```diff
@@ -0,0 +1,231 @@
...
+
+  renderCallback(renderUserdata);
+
+  lovrGraphicsSetCamera(NULL, false);
+  lovrRelease(canvas);
+}
+
+// TODO
+void lovrHeadsetUpdate(float dt) {
+}
```
bjornbytes: This isn't doing anything right?

CMakeLists.txt
```diff
@@ -532,7 +564,8 @@ if(WIN32)
   move_dll(${LOVR_GLFW})
   move_dll(${LOVR_LUA})
   move_dll(${LOVR_ODE})
-  move_dll(${LOVR_OPENAL})
-  move_dll(${LOVR_OPENVR})
+  if (LOVR_USE_OPENVR)
+    move_dll(${LOVR_OPENVR})
+  endif()
```
bjornbytes: This is removing the OpenAL .dll on windows builds?  The `if` shouldn't be necessary either since `move_dll` checks to see if the target exists before doing the copy.

src/headset/headset.h
```diff
@@ -27,6 +28,7 @@ typedef enum {
   HEADSET_UNKNOWN,
   HEADSET_VIVE,
   HEADSET_RIFT,
+  HEADSET_OCULUS_MOBILE,
```
bjornbytes: Any way this can be more specific, like "OCULUS_GO"?  Assuming Oculus is exposing something like that...

bjornbytes:
Another thing I would consider a MUSTFIX: `lovr.getOS` should return "android".  There's a giant chain of `#ifdef`s in `api/lovr.c`.

CMakeLists.txt
```diff
@@ -319,6 +334,14 @@ add_executable(lovr
   src/lib/vec/vec.c
 )
 
+if (ANDROID)
+  # For Android only we build as a library.
```
bjornbytes: I would prefer a more semantic condition like "build as (shared) library" rather than "on android we happen to build as a lib".

src/graphics/graphics.c
```diff
@@ -53,13 +60,16 @@ void lovrGraphicsDestroy() {
...
 void lovrGraphicsPresent() {
+#ifndef NO_WINDOW
   glfwSwapBuffers(state.window);
+#endif
   lovrGpuPresent();
 }
 
 void lovrGraphicsCreateWindow(int w, int h, bool fullscreen, int msaa, const char* title, const char* icon) {
-  lovrAssert(!state.window, "Window is already created");
+  lovrAssert(!state.window && !state.initialized, "Window is already created");
```
bjornbytes: I'm confused about this.  Wouldn't it make all calls to `lovr.graphics.createWindow` cause an error if `require('lovr.graphics')` has already been called?

src/headset/oculus_mobile.c
```diff
@@ -0,0 +1,231 @@
...
+static void (*renderCallback)(void*);
+static void* renderUserdata;
+
+static float offset;
+
+void lovrOculusMobileDraw(int framebuffer, int width, int height, float *eyeViewMatrix, float *projectionMatrix) {
+  lovrGpuDirtyTexture();
+
+  CanvasFlags flags = {0};
+  Canvas *canvas = lovrCanvasCreateFromHandle(width, height, flags, framebuffer, 0, 0, 1, true);
```
bjornbytes: Leaving a note about the memory allocation here (can be fixed later post-merge).

src/headset/oculus_mobile.c
```diff
@@ -0,0 +1,231 @@
...
+  oculusMobileControllerGetAxis,
+  oculusMobileControllerIsDown,
+  oculusMobileControllerIsTouched,
+  oculusMobileControllerVibrate,
+  oculusMobileControllerNewModelData,
+  oculusMobileRenderTo,
+  oculusMobileUpdate
+};
+
+// Pseudo GLFW
```
bjornbytes: Adding note to add `platform.h` to handle platform-specific code more elegantly (not necessary for merge).

src/headset/oculus_mobile.c
mcclure: Yup, accidental leftover from an earlier version of the code. Will remove

src/headset/headset.h
mcclure: "Oculus Mobile" is the name Oculus gives to the platform which comprises the Samsung Gear, Oculus Go, and (I assume) Oculus Quest platforms.

What I am 98% sure I could do is use DRIVER_OCULUS_MOBILE for the driver, and then for the headset flag have HEADSET_GEARVR/'gearvr', HEADSET_OCULUS_GO/'oculusgo', and HEADSET_OCULUS_QUEST/'oculusquest' flags. (HEADSET_GEAR/'gear', HEADSET_GO/'go', and HEADSET_QUEST/'quest' would probably also be options). That work?

CMakeLists.txt
mcclure: on Android it's impossible to build as anything other than a shared library. there's no concept of an executable. I'l switch this to LOVR_BUILD_SHARED and have this set to ON when if (ANDROID)

CMakeLists.txt
mcclure: Wait, now I see the problem. I will remove this change entirely.

src/graphics/graphics.c
mcclure: Possible. This could be related to the thing where the oculus mobile always crashes on second launch >_> i'll look into it

src/headset/oculus_mobile.c
mcclure: That might also be a good time to put a more coherent abstraction over what this patch calls "NO_WINDOW".

mcclure:
did further sound testing and made some changes in the other repo. openal works great now

mcclure:
> Another thing I would consider a MUSTFIX: `lovr.getOS` should return "android". There's a giant chain of `#ifdef`s in `api/lovr.c`.

Addressed btw

mcclure:
Just a note that in the current version openal sounds that do not have a position set come out of the left ear only. It is believed this could possibly be fixed by upcoming lovr.headset.getPosition() fixes and this should be tested after merge

src/headset/headset.h
cvan: yep, that is correct and sounds like the right approach for platform (driver) vs. headset. as you said, oculusvr is the platform name for all those mobile Oculus headsets.

src/graphics/graphics.c
mcclure: state gets cleared on Destroy. this is the same as on PC, since initialized and window are both set in this function

bjornbytes: I see, this works because "initialized" is only set in `lovrGraphicsCreateWindow` here.

PR #70, opened 2018-11-12
Support errors in lovr.draw on Oculus Mobile, and other fixes

mcclure:
This began as an attempt to make errors in lovr.draw() on Oculus Mobile appear in the lovr.errhand screen instead of a printf-and-quit. It spiraled out of control and became the following:

General changes (all drivers):

- Amended the boot.lua error handling so when an error occurs in the error handler, the inner error is printed before quitting
- Silent quit instead of crash if a user implements lovr.errhand but it doesn't return a function

Oculus Mobile changes:

- The lovr.errhand screen is now correctly invoked for errors that occur inside lovr.draw. Multiple changes were needed to make this work:
    - Instead of calling renderHelper, which uses lua_call (unsafe as Oculus Mobile does not call renderHelper) the oculus driver gets hold of the Lua ref and lua_pcalls itself. A new lovrHeadsetExtractRenderFn is added to make this possible.
    - A mechanism is added where if the coroutine resume in boot.lua returns a value, boot.lua treats this as the string returned from luax_getstack and invokes lovr.errhand.
- Added a custom atpanic that routes through lovrThrow (since stderr gets eaten). With the draw() changes this should never be encountered, but it's good just in case. In current testing the tracebacks this prints don't seem to be right.
- Fix major bug in android_vthrow that meant % codes didn't work in lovrThrow on Android
- Nothing to do with errors, but fix getAxis("trigger")


src/headset/oculus_mobile.c
```diff
@@ -566,12 +560,45 @@ void bridgeLovrUpdate(BridgeLovrUpdateData *updateData) {
...
+  lovrGraphicsSetCamera(&camera, true);
+
+  if (renderUserdata && renderErrorRef == LUA_NOREF) { // Don't render if a render error occurred
+    int fn = lovrHeadsetExtractRenderFn(renderUserdata);
+    lua_pushcfunction(L, luax_getstack);
+    lua_rawgeti(L, LUA_REGISTRYINDEX, fn);
+    if (lua_pcall(L, 0, 0, -2)) {
+      renderErrorRef = luaL_ref(L, LUA_REGISTRYINDEX); // save luax_getstack output
+    }
+  }
```
bjornbytes: I don't think the "core headset.renderTo" code should know about Lua details like this.  It's a little more painful, but this sort of thing should go in `api/headset.c`.  Two ideas for this:

- The error could be stored in a known key in the registry like `_lovrRenderError` or something that the lua_resume code could extract later, or
- The render callback could have its signature modified to return an error message/value that `bridgeLovrDraw` could maybe store for later.

For more background on this motivation: my goal is to only have Lua code/headers in `main.c`, `luax.c`, and the `api` folder.  So far it's been a decent way to separate things and it makes it easier to add other language bindings in the future.

src/headset/oculus_mobile.c
```diff
@@ -483,6 +475,7 @@ void bridgeLovrInit(BridgeLovrInitData *initData) {
   // Copypaste the init sequence from lovrRun:
   // Load libraries
   L = luaL_newstate(); // FIXME: Just call main?
+  lua_atpanic(L, luax_custom_atpanic);
```
bjornbytes: I was getting panics for a while and was able to fix it in https://github.com/bjornbytes/lovr/commit/185bb8438e9a14f79bba40b9b8de3478d4157200.  Does changing `lovrSetErrorCallback` to error on `Lcoroutine` instead of `L` fix any of the panic issues you were having?  I'd sorta prefer it if panicking was not a normal error handling mechanism.

src/luax.h
```diff
@@ -20,6 +20,7 @@ void* _luax_checktype(lua_State* L, int index, const char* type);
 void luax_pushobject(lua_State* L, void* object);
 void luax_vthrow(lua_State* L, const char* format, va_list args);
 int luax_getstack(lua_State* L);
+int luax_getstack_panic(lua_State *L);
 void luax_pushconf(lua_State* L);
 int luax_setconf(lua_State* L);
-Color luax_checkcolor(lua_State* L, int index);
+Color luax_checkcolor(lua_State* L, int index);
```
bjornbytes: This is really silly but this change gives me a compile error when using `-Werror -Wall -Wextra` (I use those flags when building with tup).

mcclure: This was a mistake >_>

src/headset/oculus_mobile.c
mcclure: I'm not currently getting panics; the panics were in lovr.draw and the rest of the patch fixes them. The problem is that without this patch a panic results in *no* debug output— actually, on oculus mobile it causes a crash in a different thread (because the main thread tries to crash and some internal render thread then crashes). So this is needed because without it, development on oculus mobile becomes very difficult (because the act of development can introduce panics, and this patch makes it possible to distinguish a panic from heap corruption).

src/resources/boot.lua
```diff
@@ -169,17 +173,55 @@ function lovr.threaderror(thread, err)
...
-    if result and ok then return result
-    elseif not ok then thread = result end
-    coroutine.yield()
-  end
+    local ok, result = xpcall(continuation, onerror)
+    if result and ok then return result          -- Result is value returned by function. Return it.
+    elseif not ok then continuation = result end -- Result is value returned by error handler. Make it the new error handler.
+
+    if type(continuation) == "string" then -- LuaJIT returns a fixed string if an error occurs in an xpcall error handler
+      print("Error occurred while trying to display another error.")
```
bjornbytes: Could you use single quotes for the string literals added to this file?

src/luax.c
```diff
@@ -215,6 +215,28 @@ int luax_getstack(lua_State* L) {
...
+  if (!lua_isfunction(L, -1)) {
+    lua_pop(L, 2);
+    return 0;
+  }
+  lua_remove (L, -2);     // Pop debug object
+  lua_insert(L, -2);      // Move message to top
+  lua_pushinteger(L, 0);
+  lua_call(L, 2, 1);      // Call debug.traceback
+  return 1;
+}
```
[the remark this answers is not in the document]
mcclure: Fixed the various style changes. Your suggestion of merging the two getstacks seems reasonable but I think you have something kinda specific in mind so I'd prefer to let you handle that after merge.

Since we only have two levels we want (2 or 0) another way to do this without relying on closures would be to have one `lua_getstack(lua_State *L, int errorLevel)` and then have two hardcoded `lua_getstack_2` and `lua_getstack_0` C functions.

PR #71, opened 2018-11-13
"Fix" arg standard and allow mounting inside a zip file

mcclure:
The intent of this PR is to allow a --inside or -i argument, which specifies a path "inside" the gamepath which is actually mounted. This is pointless if the gamepath is a directory, but useful if the gamepath is a .zip file, and required if the gamepath is an .apk file on android.

I want to stress I have no opinions what to name this argument. Other options would be --innerPath or --nested.

As this is the first/only argument taken by lovr other than the gamepath, major changes were made to argument parsing to make this possible.

Please see commit messages for more detail.


src/main.c
```diff
@@ -101,19 +101,71 @@ static void onGlfwError(int code, const char* description) {
...
 lua_State* lovrInit(lua_State* L, int argc, char** argv) {
   glfwSetErrorCallback(onGlfwError);
   lovrAssert(glfwInit(), "Error initializing GLFW");
   glfwSetTime(0);
 
-  // arg
+  // arg table
+  // Args follow the lua standard https://en.wikibooks.org/wiki/Lua_Programming/command_line_parameter
+  // * The lovr interpreter and its arguments are in successive negative indexes starting at -1.
+  //   The interpreter will always be at arg[-1] and always non-nil.
```
bjornbytes: I tested this script:

```lua
for i = -3, 3 do
  print(arg[i])
end
```

like this:

```sh
lua -E test.lua
```

and got

```
nil
lua
-E
test.lua
hi
nil
nil
```

So I think it's reversed and the interpreter is at the "lowest" index.  I haven't yet read on enough to see if this impacts the parsing.

src/headset/oculus_mobile.c
```diff
@@ -416,64 +359,6 @@ void bridgeLovrInit(BridgeLovrInitData *initData) {
...
-      // Save timestamp in a new lastprogram.dat file
-      timeFile = fopen(timePath, "w");
-      fwrite(&apkstat.st_mtim.tv_sec, sizeof(apkstat.st_mtim.tv_sec), 1, timeFile);
-      fwrite(&apkstat.st_mtim.tv_nsec, sizeof(apkstat.st_mtim.tv_nsec), 1, timeFile);
-      fclose(timeFile);
-    }
-
-    sdsfree(timePath);
-  }
-
```
bjornbytes: 🎉

bjornbytes:
How about `--archive-root` for the argument name?  It's long, but unambiguous.

src/main.c
```diff
@@ -101,19 +101,71 @@ static void onGlfwError(int code, const char* description) {
...
+  argparse_exe, // Interpreter
+  argparse_flags,
+  argparse_game // Game path and arguments
+} ArgParseState;
+
+typedef enum { // In state argparse_flag, what flag is being searched for?
+  argflag_none, // Not processing a flag
+  argflag_inside
+} ArgFlag;
+
```
bjornbytes: Enums usually use `SCREAMING_CASE`, should be consistent here.

src/api/filesystem.c
```diff
@@ -263,7 +263,8 @@ static int l_lovrFilesystemMount(lua_State* L) {
   const char* path = luaL_checkstring(L, 1);
   const char* mountpoint = luaL_optstring(L, 2, NULL);
   bool append = lua_isnoneornil(L, 3) ? 0 : lua_toboolean(L, 3);
-  lua_pushboolean(L, !lovrFilesystemMount(path, mountpoint, append));
+  const char* mountpointInner = luaL_optstring(L, 4, NULL);
```
bjornbytes: minor: I think this would be easier to understand if it was named "root" (it seems to also match up with the PhysFS terminology).

PR #72, opened 2018-11-14
Android cleanup/improvements

bjornbytes:
Some smaller Android cleanup tasks.  I can't test these and don't know much about Android so it's entirely possible that some of this needs to be changed or reverted.

- I made the lovrLog functions macros for all platforms, and used them in oculus_mobile.c.  It seems like the biggest downside about this is going from N log levels to 2 log levels.  If that's a problem, maybe a log level concept can be added to platform.h.
- Ditched dynamic module loading and use a hardcoded list everywhere so Android doesn't need to special case things as much (commit message has more info).
- Moved print override into luax
- Tried to combine the two getstack functions but ended up porting over luaL_traceback and using that for the panic error handler.


mcclure:
I tested this and it works on the headset. It looks good to me except
- I actually did like the dynamic-load because it makes it easier to casually hack additions to the engine*— would it be possible to keep it in as a fallback, at least, *after* the preloads, even if it only activates on desktop platforms?
- Any __android_log_prints that hadn't been changed to lovrLog/lovrWarn were unintentional— EXCEPT LOG_FATAL, I think we actually have lost something by dropping that? But this is minor because some of the fatal logs are currently being printed in lua with print() so we're already not reliably doing FATAL. So… I think it's ok to go with this patch for now and revisit later.
- Travis failure of course

bjornbytes:
Dynamic load is still possible via the lovr.filesystem loader (you can `require('lib.dll')` and it'll load `luaopen_lib`) so that should still make the extension use case possible.  The only thing that's being removed is loading modules from _within_ the executable, which seems alright for now.

Thanks for testing this!  Merging...

PR #75, opened 2018-11-23
Fixes to error handling in boot.lua

mcclure:
- The pcalls to run conf were triggering their error very late, and were detecting error based on the second return value of pcall being non-nil. In fact the second return value of pcall is often non-nil in the case of *success*. We should check the first parameter, success/failure, instead.

- It is possible in principle to call error() with a non-string value, so errhand should sanitize its values with tostring.

- There is a check in the main boot.lua loop whether xpcall returned a string, which is the sign luajit hit an error in an error handler. This check was occurring in the wrong place (my fault; merge error when preparing PR) so if lovr.load() hit this case we were falling back to the horrible “could not call a string” error.


mcclure:
Tests:

- Added `error(true)` to lovr.draw of a sample program
    - Pre patch: Program quits, prints to terminal:

            Error occurred while trying to display another error.
    - Post patch: Normal lovr error screen with message “Error: True”

^ tests added tostrings

- Added an empty conf.lua to a sample project
    - Pre patch: Program quits, prints to terminal:

                Error occurred while trying to display another error:
                attempt to call a string value

                Stack:
                [C]: in function 'xpcall'
    - Post patch: Program runs normally

^ tests both of the other two changes. the conf change fixes conf.lua and the reordering change results in "attempt to call a string value" as an error message instead of the more correct "Error occurred while trying to display another error." which is what I get if I back out the conf and tostring changes only and rerun this test.

PR #76, opened 2018-11-23
Android lua improvements

mcclure:
Currently provides only a lovr.android.getApplicationId(). This returns an Android-specific identifier that doesn't cleanly map to anything specific in other OSes.

~This hasn't been tested, so please don't merge it yet~


mcclure:
Commit "android lua module" is the lua module

The next three fix the android build (it is currently broken on master)

"Fix lovr.event.quit("restart") on Android" is a separate fix

bjornbytes:
Thanks, merging!  Note: I might be adding newlines to the ends of files again if clang complains.

PR #82, opened 2018-11-30
Awkward hotfix to tinycthread.h

mcclure:
Becuase of https://github.com/tinycthread/tinycthread/issues/47 . Only needed on Android. Not needed in the glfw tinycthread because we don't use glfw on Android.


bjornbytes (approved):

mcclure:
Note per this comment

https://github.com/r-lib/later/pull/129#issuecomment-3149584465

It's possible you can remove this kludge now. But I don't know which versions of Android you need to support for present-day Oculus.

PR #99, opened 2019-01-08
ShaderBlock:getValue; Shader:send vector fixes;

bjornbytes:
Allows you to read back values from a ShaderBlock.

- This is also probably useful for Shaders and their uniforms.
- This creates SO MANY tables.  It feels so bad to expose this.  But
  also it's the only way to read values back from a compute operation,
  so it's pretty indispensable.  In the future there could be a Blob
  variant.
- readable and writable are confusing now, since writable means "write from shader" and readable means "read from Lua".  Proposal to change writable to shaderwritable?  Anything better?

### DIFF
--- a/src/api.h
+++ b/src/api.h
@@ -61,6 +61,7 @@ extern const char* ArcModes[];
 extern const char* AttributeTypes[];
 extern const char* BlendAlphaModes[];
 extern const char* BlendModes[];
+extern const char* BlockTypes[];
 extern const char* BufferUsages[];
 extern const char* CompareModes[];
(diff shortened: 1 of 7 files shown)


shakesoda:
how about "readback"

bjornbytes:
I like readback because that makes it more clear that it's a slow operation

bjornbytes:
- Block type needs to be explicitly defined and it is either "uniform" or "compute"
- Change :getValue to :read
- Keep `readable` flag, remove `writable` flag

PR #102, opened 2019-01-24
 Attempt to make a lovr.mirror() for Window purposes

mcclure:
As discussed in chat.

Adds lovr.graphics.hasWindow()


mcclure:
With the second commit, this version has now been tested on fake, openvr, and oculusvr. i have not tested webvr and oculus-mobile needs fixes to compile with master right now anyway. I have not tested the full matrix of edge cases such as building without headset, disabling lovr.headset or alternative mirroring options.

src/headset/oculus.c
```diff
@@ -449,6 +449,11 @@ static void oculusRenderTo(void (*callback)(void*), void* userdata) {
   }
 }
 
+static Texture* oculusGetMirrorTexture() {
+  ovr_GetMirrorTextureBufferGL(state.session, state.mirror, &handle);
```
bjornbytes: I am confused, where is `handle` declared?

src/resources/boot.lua
```diff
@@ -109,6 +109,7 @@ function lovr.boot()
 end
 
 function lovr.run()
+  local hasWindow = lovr.graphics.hasWindow()
```
bjornbytes: It is possible to delay window creation until after boot time, so `lovr.graphics.hasWindow` can return different results.  For this reason I think it should not be cached and should be invoked down below on every frame.

src/api/headset.c
```diff
@@ -292,6 +293,18 @@ static int l_lovrHeadsetUpdate(lua_State* L) {
...
 }
 
+static int l_lovrHeadsetGetMirrorTexture(lua_State* L) {
+  Texture *texture = NULL;
+  if (lovrHeadsetDriver->getMirrorTexture)
+    texture = lovrHeadsetDriver->getMirrorTexture();
+  if (texture)
+    luax_pushobject(L, texture);
+  else
+    lua_pushnil(L);
```
bjornbytes: Small nit, `luax_pushobject` handles NULL and will push nil in that case, so the `if` here isn't really necessary.

src/api/graphics.c
```diff
@@ -339,6 +339,15 @@ static int l_lovrGraphicsGetDimensions(lua_State* L) {
   return 2;
 }
 
+static int l_lovrGraphicsGetHasWindow(lua_State *L) {
```
bjornbytes: This should be called `l_lovrGraphicsHasWindow` (same with the platform function name) to stick with the existing naming conventions.

src/api/graphics.c
```diff
@@ -339,6 +339,15 @@ static int l_lovrGraphicsGetDimensions(lua_State* L) {
...
+  bool window = false;
+#ifndef NO_WINDOW
```
bjornbytes: I ended up folding the NO_WINDOW define into the platform layer, so in this case I think `platform/android.c` should implement `lovrPlatformHasWindow` and always return `NULL`, instead of using an ifdef here.

src/headset/oculus.c
mcclure: That looks like an error. I don't think this should compile.
My test did compile against the oculusvr sdk, so that tells me my test must have been testing the wrong thing.
I can't test again today but tomorrow I should be able to do a more rigorous test.

mcclure:
New commit addresses all above issues but has not been tested on every platform

Issue #58, opened 2018-09-21
Animator:getAABB


bjornbytes:
Need a way to get the 'animated bounding box' of a Model.  Either Model:getAABB or Animator:getAABB could work.  Performance is definitely a concern for huge models, so there are other techniques like threads/heuristics/lower-poly model approximations that could be used, but those could be explored in Lua-land.

bjornbytes:
Fixed in #103

PR #103, opened 2019-01-29
glTF

bjornbytes:
This isn't 100% done yet but the PR helps me visualize the changes better, run the builds, and leave notes as I prepare for merge.


bjornbytes:
This closes #58, `Model:getAABB` now respects the animated transform of nodes.  Note that it still doesn't take skinning/joints into account, I think it's unreasonable for LÖVR to try to compute this and it doesn't seem like other engines try to do it either.

bjornbytes:
Everybody hold onto something.

PR #106, opened 2019-01-29
Fix primitive restart on GLES;

bjornbytes:


mcclure:
This looks fine and mergeable to me, although I still believe it would be better to have a big GL_ES_PLATFORM switch of some sort (or switch on this via checking GLAD gl version) to more comfortably accommodate hypothetical future GLES2.0 platforms.

bjornbytes:
Something interesting is in this case with primitive restart, android and emscripten have different behavior, we're really dealing with WebGL and GLES, not just GLES, and they're very similar but not always the same.

So if we added a define for GL_ES, we'd need to have weird exceptions for it, like: "if it's GLES (AKA android or emscripten) and it's also not emscripten then do this special thing", and at that point I'd rather reason about the "raw" values of the defines.

Though I wouldn't mind having something like LOVR_GL, LOVR_GLES, and LOVR_WEBGL.  Would that be helpful?

mcclure:
(But, yes, I'd be happier forking on WEBGL vs GLES than I would on EMSCRIPTEN vs ANDROID.)

mcclure:
Thanks! I'll test the patch on oculus go tonight

PR #108, opened 2019-02-01
Fix compile with -DLOVR_ENABLE_HEADSET=NO

mcclure:
The event code was using one of the headset string tables, so there was a link error.
Tested this on mac and it seems to work.


bjornbytes:
Thanks!

PR #111, opened 2019-02-08
Fix Oculus driver compilation

mcclure:
See ongoing slack discussion


mcclure:
Outstanding problems:
- Mirror window is broken in oculus
- state.offset (y standing offset) is being totally ignored in oculus

PR #112, opened 2019-02-08
Don't copy unused DLLs on Windows

mcclure:
At the end of the Windows build we copy all the built DLLs into place. However, in some builds not all DLLs are built, and currently we try to copy these DLLs even though they do not exist. This then fails.


bjornbytes:
I semi-recently (September) added `if(TARGET ${ARG_TARGET})` to move_dll to try to fix this problem.  I may try to see if something is going wrong with that, since handling it generally in one place would be preferable to having an `if` around each invocation.

mcclure:
That seems like it *should* be enough, yeah. Maybe I added this to my CMakeLists before I merged the September changes??

mcclure:
I JUST REMEMBERED WHY I WANTED THIS

Using the normal code, if LOVR_OPENAL is empty, then building this on my system results in

    CMake Error at CMakeLists.txt:664 (move_dll:
      move_dll Function invoked with incorrect arguments for function named:
      move_dll

So yes, move_dll can correctly handle it being null
but we never get inside of move_dll

PR #115, opened 2019-02-18
Configurable singlepass;

bjornbytes:
Adds a `graphics.singlepass` conf flag so you can turn off singlepass stereo, and GPU separates the concepts of "is singlepass supported" vs. "is singlepass enabled".  I don't know if it's useful.  It's kind of annoying to add `singlepass` variables to both graphics and gpu for this.  There should probably be `lovr.graphics.isSinglepass` that asks GPU if singlepass is active.


src/resources/boot.lua
```diff
@@ -61,6 +61,9 @@ function lovr.boot()
       thread = true,
       timer = true
     },
+    graphics = {
+      singlepass = true
+    },
     gammacorrect = false,
```
bjornbytes: At some point I want to move gammacorrect into `graphics`, but it's a breaking change.

bjornbytes:
Would this be helpful for multiview in android/web?

PR #116, opened 2019-03-06
Refactor asynchronous headset render errors;

bjornbytes:
It's now a more general error stored in the registry instead of
in a global variable.  There are more helper functions.

I need to do a fair amount of testing on this.


bjornbytes:
Tested the error screen on Android (both synchronous and asynchronous errors).

PR #117, opened 2019-03-13
Update type system;

bjornbytes:
This is an update to the type system.  It's meant to reduce the size of the "Ref" object header and improve the speed of Lua/C interop.  Summary of changes:

- Ref only stores a refcount now, instead of a destructor and a type name.
  - The struct goes from 24 bytes to 1 byte.
  - In most places where Refs are used, we already know their type, don't need to store it.
- lovrRelease takes a destructor to use to destroy the object.
  - Lua stores the destructor as an upvalue on the __gc metamethod.
- Userdata allocated by Lua store the type as an enum along with the object pointer.
  - This is where type information is actually needed.
  - Using an enum is faster than doing a string comparison.
  - Creating new userdata requires an explicit type since we can't look it up on the Ref.
- Variants become a little more complicated.  They are the only place where we truly have a generic reference.  So for this case specifically, we store the type in the Variant now.
  - Still working on this.
- Use a macro to simplify luax_registertype and luax_extendtype.
- Correctly error when refcounts overflow/underflow.

Stuff I'm still thinking about:

- Is this actually good.
- Our macro game is getting a bit too strong, especially in luax.
- Should type stuff go in a separate types.h or still be in util.h?
- Try to standardize more on terminology between Ref, Object, Type, Proxy.
- Maybe lovrSupertypes can go away, we only use it for 3 types...could just be a switch.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -300,6 +300,7 @@ endif()
 set(LOVR_SRC
   src/main.c
   src/util.c
+  src/types.c
   src/luax.c
   src/api/lovr.c
   src/lib/math.c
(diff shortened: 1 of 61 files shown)


bjornbytes:
Rough benchmark: 10% improvement in total time that it takes to call a simple function on an object 100000000 times.

bjornbytes:
This was reimplemented on the `dev` branch, closing.

PR #120, opened 2019-05-24
Oculus quest support

mcclure:
As discussed in Slack, there are two problems with this which appear to be on the lovr side:

- Driver isTouched is not being called when lua calls isTouched. It appears isDown is being called instead.
- Values returned from axis are being absolute-valued.

There is one limitation of the patch itself:

- Acceleration is not yet supported

This patch requires a patch on the "lovrapp" side.

### DIFF
--- a/src/core/platform_android.c.h
+++ b/src/core/platform_android.c.h
@@ -33,7 +33,8 @@ bool lovrPlatformCreateWindow(WindowFlags* flags) {
 }
 
 void lovrPlatformGetWindowSize(int* width, int* height) {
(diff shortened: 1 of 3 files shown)


mcclure:
- I did a rebase; with Bjorn's fixes touched is now fixed
- The axis problem was in fact a problem with lovr.graphics.print (as discussed in slack)

I think this might be ready to merge but I gotta do another test with the Quest on monday

mcclure:
This has now been tested on both Go and Quest and is ready to merge.

PR #124, opened 2019-06-29
Fix -DLOVR_ENABLE_THREAD=OFF compile

mcclure:
Note: I **DO NOT KNOW** if this is the correct fix, bjorn would know better.

l_event.c was processing a thread-related event and in the process using a thread struct, even when LOVR_ENABLE_THREAD is undefined and threads do not exist. I instead made receiving that event an assert failure.

### DIFF
--- a/src/api/l_event.c
+++ b/src/api/l_event.c
@@ -9,7 +9,9 @@
 const char* EventTypes[] = {
   [EVENT_QUIT] = "quit",
   [EVENT_FOCUS] = "focus",
+#ifdef LOVR_ENABLE_THREAD
   [EVENT_THREAD_ERROR] = "threaderror",
+#endif
 };
 
 static LOVR_THREAD_LOCAL int pollRef;
@@ -103,11 +105,13 @@ static int nextEvent(lua_State* L) {
       lua_pushboolean(L, event.data.boolean.value);
       return 2;
 
+#ifdef LOVR_ENABLE_THREAD
     case EVENT_THREAD_ERROR:
       luax_pushtype(L, Thread, event.data.thread.thread);
       lua_pushstring(L, event.data.thread.error);
       lovrRelease(Thread, event.data.thread.thread);
       return 3;
+#endif
 
     case EVENT_CUSTOM:
       for (uint32_t i = 0; i < event.data.custom.count; i++) {
--- a/src/modules/event/event.c
+++ b/src/modules/event/event.c
@@ -37,9 +37,11 @@ void lovrEventPump() {
 }
 
 void lovrEventPush(Event event) {
+#ifdef LOVR_ENABLE_THREAD
   if (event.type == EVENT_THREAD_ERROR) {
     lovrRetain(event.data.thread.thread);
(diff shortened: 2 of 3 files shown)


bjornbytes:
I would rather have `#ifdef LOVR_ENABLE_THREAD`s around _anything_ related to that event (in `l_event` and `event.{c,h}`).  That way the concept of a thread error will just never exist if threads are turned off.

mcclure:
That seems fine, but one thing I'd worry about a *little* is maybe the value should stay in the EventType enum because otherwise there is an "ABI incompatibility" for any dynamically loaded library that uses EventType. Like you keep talking about the possibility of binary plugins, what if a binary plugin is compiled with a different LOVR_ENABLE_THREAD value from the lovr that loads it (this might create problems in other ways I guess)

At the moment, the EVENT_THREAD_ERROR event is literally never emitted except in l_thread_module.c, which isn't present with threads off, but it's referenced in a couple other places (like, it's in the EventTypes string table).

bjornbytes:
This is how we were doing it for the controller events:

https://github.com/bjornbytes/lovr/blob/v0.12.0/src/event/event.h#L15
https://github.com/bjornbytes/lovr/blob/v0.12.0/src/api/event.c#L102

I thought about "ABI compatibility".  I decided that a way to handle this would be to encourage/require non-core modules to emit custom events (EVENT_CUSTOM) instead of relying on event types that might not exist.

Even if this PR were merged, the `threaderror` event would be available but it still wouldn't work because of the assert (the alternative is maybe to ignore the event but that's not great either), so it's probably just best to not rely on event types that depend on a different module and instead use custom events.  If this were the case, it would need to be documented somewhere, probably on the C side of things.

Maybe this means that having a static set of events (or module-specific events) is a bad idea and all events should be custom events.  I don't really like custom events right now because A) they take up way more memory than an event with a known payload B) they support generic references to refcounted objects and that's actually been really challenging to support in the object system.  But that's a separate issue.

mcclure:
Another, possibly silly but simple solution would be to just make it be a requirement that all non-universal event types (ie anything whose presence can vary depending on compile flags) should come after _CUSTOM in the enum. Then _CUSTOM always has the same value and the "ABI" is always compatible.

mcclure:
Or if you worry about what happens once you have multiple module-specific enum values, you could do something like

```
#if LOVR_MODULE_THREADS
    EVENT_THREAD_ERROR,
#else
    EVENT_RESERVED1,
#endif
```

I think it's okay to put Lovr-privileged module values into the enum instead of using _CUSTOM.

Also, honestly, it's okay to throw out "ABI compatibility" as a goal, it just might make certain things a little easier in 18 months hypothetically.

mcclure:
I made a git mistake and this autoclosed.

mcclure:
Bjorn, this follows your suggested approach of removing `EVENT_THREAD_ERROR` entirely when there are no threads instead of using the assert. I did put `EVENT_THREAD_ERROR` at the end of the event enum.

PR #125, opened 2019-06-29
t.headset.fakeKbamBlock = true in conf.lua

mcclure:
Set this to disable ("block") the controls of the desktop driver. The desktop driver will still run, but the keyboard and mouse controls will not activate.

This is of pretty narrow usage, but has been useful to me. Situations you might want this:
- You are using lovr-mouse or lovr-keyboard, and don't want the act of typing to move the camera around. (My primary use)
- You are shipping a release build of an application and decide you don't want to leave in (effectively) "debug controls"

### DIFF
--- a/src/api/l_headset.c
+++ b/src/api/l_headset.c
@@ -642,6 +642,15 @@ int luaopen_lovr_headset(lua_State* L) {
     lua_pop(L, 1);
   }
 
+#if LOVR_USE_DESKTOP_HEADSET
+  // Special conf only used by desktop driver
+  lua_getfield(L, LUA_REGISTRYINDEX, "_lovrconf");
+  lua_getfield(L, -1, "headset"); // FIXME: Could be nil?
+  lua_getfield(L, -1, "fakeKbamBlock");
+  lovrHeadsetFakeKbamBlock(lua_toboolean(L, -1));
+  lua_pop(L, 3);
+#endif
+
   if (lovrHeadsetInit(drivers.data, drivers.length, offset, msaa)) {
     luax_atexit(L, lovrHeadsetDestroy);
   }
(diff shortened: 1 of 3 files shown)


bjornbytes:
Is this (or should this be) any different from either

A) Removing `desktop` from the `t.headset.drivers` list.
B) Disabling the headset module completely.

mcclure:
I don't know. If you don't have desktop in the drivers list, and there is NO driver, and so no driver at all is selected… what happens?

With this mode it can go into desktop mode and still *render*, and possibly the program can then detect the desktop driver is selected and do something case-specific (like, a "no headset found!" screen).

A note: In my personal fork, where I first introduced this, part of why I did it this way is I call the lovrHeadsetFakeKbamBlock() C function manually, in my debug tools— when I launch my debug menu I call lovrHeadsetFakeKbamBlock(true), and when the menu is closed I call lovrHeadsetFakeKbamBlock(false) so I can move the camera around if I want. Maybe it would be better to expose that directly than rely on the t.headset.fakeKbamBlock setting.

bjornbytes:
If there's no driver, it looks like requiring the headset module will error, telling you that it couldn't find a driver.  So that won't work.

Another idea: `lovr.headset.update` is the only thing in the desktop driver that handles keyboard/mouse input and updates the headset transform.  So you could have a custom `lovr.run` that's like:

```lua
function lovr.run()
  return function()
    -- ...

    if lovr.headset and (lovr.headset.getDriver() ~= 'desktop' or not debug.active) then
      lovr.headset.update(dt)
    end

    -- ...
  end
end
```

It's a bit more painful, but I consider this to be a somewhat exceptional use case, and most weird/advanced projects will end up with their own `lovr.run` at some point anyway.

Another thing I'm thinking about is splitting `desktop` into one "vr simulator" backend and one "keyboard/mouse input" backend (I may have already mentioned this?).  I'm not yet sure if this would provide a solution to the problem here.

bjornbytes:
Closing this, I would prefer to find a way to do this that doesn't require a special driver-specific flag.

PR #126, opened 2019-06-29
mat4_transform_project in maf.h

mcclure:
I'm not sure if you actually want this in master or not. but just in case:

mat4_transform ignores the final row of the matrix, which normally contains scale data. This is fine for modelview matrices, which is all lovr currently uses mat4_transform for. However it fails if you ever use mat4_transform on a projection matrix.

mat4_transform_project takes the final row into account, so you can use it with projection matrices. This was useful in my fork and might be useful to have around in lovr someday later.

### DIFF
--- a/src/core/maf.h
+++ b/src/core/maf.h
@@ -675,6 +675,17 @@ MAF void mat4_transform(mat4 m, vec3 v) {
   );
 }
 
+MAF void mat4_transform_project(mat4 m, vec3 v) {
+  vec3_set(v,
+    v[0] * m[0] + v[1] * m[4] + v[2] * m[8] + m[12],
+    v[0] * m[1] + v[1] * m[5] + v[2] * m[9] + m[13],
+    v[0] * m[2] + v[1] * m[6] + v[2] * m[10] + m[14]
+  );
+  float w =
+    v[0] * m[3] + v[1] * m[7] + v[2] * m[11] + m[15];
+  vec3_set(v, v[0]/w, v[1]/w, v[2]/w);
+}
+
 MAF void mat4_transformDirection(mat4 m, vec3 v) {
   vec3_set(v,
     v[0] * m[0] + v[1] * m[4] + v[2] * m[8],


bjornbytes:
Any conceivable downside to just doing `mat4_transform_project` everywhere?  This could replace the existing `mat4_transform`.  It would also be 4-wide so it could more easily take advantage of SIMD.

mcclure:
`mat4_transform_project` is inherently slower, because it does more work; `mat4_transform` is effectively an optimization when the final row is known to be trivial. I'm not sure I understand the SIMD idea. You mean, you are thinking of taking a vec4 as an input? That's interesting, but if we don't have a **use** for vec4s I don't know if it's a real benefit. Maybe you know something I don't.

What **might** make sense is renaming `mat4_transform` to `mat4_transform_model` or `mat4_transform_unscaled` or something, to reflect it's the "special, optimized" version, use `mat4_transform_unscaled` in place of all current uses, and make `mat4_transform_project` the new `mat4_transform`.

bjornbytes:
vec3s always store 4 floats now, the last one is just unused.  When doing operations on vec3s, the operation is performed on all 4 elements, including the final unused one.  This causes the compiler to emit a single 4-float SIMD instruction instead of three single-float instructions.

This transform_project function could set v[3] to 1.0 and then do its dot products and the w divide, and that would provide the correct answer for `mat4_transform` and `mat4_transform_project`, and also (likely) be _faster_ than `mat4_transform` because it's doing work on 4 things instead of 3.

I think that's preferable because then there's a single `mat4_transform` function and it's (likely) faster.

Let me know if you wanna work on this or if I should merge and work from there!

mcclure:
OK! I think I get the idea, but it looks like you have something specific in mind that I don't. I would like to request you handle it because you have a clear idea of how to proceed.

If you're using dot product operations, it might be better if you start from your own code rather than using this patch as a basis.

Also, a request: As an "external user" of maf.h (ie someone branching the code), I passed 3-value arrays into maf.h functions because the type is "vec3" and I assumed I only needed 3 values in a vec3. If vec3 is actually 4 values it should either be named vec4 or there should be a comment warning that vec3 is assumed to be 4*4 bytes wide.

bjornbytes:
Ok, going to merge and work on a hopefully better version of mat4_transform based on mat4_transform_project.  I think at some point you will be able to switch to `mat4_transform` instead of `mat4_transform_project` and it will do the W divide.

Will add comment mentioning that the vec3 pointer needs to have 4 elements (I don't think it can be named vec4 because at some point there might be a real vec4 and, e.g. `vec3_length` and `vec4_length` should behave differently even if they're both stored as float[4]).

PR #127, opened 2019-06-29
Make it possible to include lovr headers from C++

mcclure:
Although Lovr is pure C, users might fork Lovr or attempt to load binary Lua plugins that reference the header files. My own fork of Lovr contains a number of C++ files. I think it's worth making the Lovr headers C++-compatible as long as the changes do not create an inconvenience for mainline Lovr dev.

This PR contains three patches to make headers C++ friendly:
- Patch SDS to detect __cplusplus and not use empty arrays when present
- Patch a line in maf.h to use a variable instead of an inline array
- Put an #error in core/ref.h detecting __cplusplus and explaining that we are using C++-incompatible C11 atomics-- I have some ideas about how to "fix" this properly at some future point but for now I think a useful error message is good enough

See commit messages for more detail

All of these changes should be total noops for the normal C build

### DIFF
--- a/src/core/maf.h
+++ b/src/core/maf.h
@@ -510,12 +510,13 @@ MAF mat4 mat4_translate(mat4 m, float x, float y, float z) {
 
 MAF mat4 mat4_rotateQuat(mat4 m, quat q) {
   float x = q[0], y = q[1], z = q[2], w = q[3];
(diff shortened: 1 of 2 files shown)


mcclure:
Removed the ref.h commit because that is handled by the other PR.

PR #128, opened 2019-06-29
Y-flip mirror texture on Oculus

mcclure:
Oculus gives us the mirror texture "upside down" compared to every other driver, so flip it rightside up before drawing in boot.lua.

**WARNING: I HAVEN'T TESTED THIS ON MASTER YET**. It worked a few months ago. I'll have time to test it next week. EDIT: I tested it, it worked as expected


bjornbytes:
Holding off until we make a decision on potentially switching mirror texture to use Canvas instead of oculus mirror texture.

bjornbytes:
Oh, I didn't see that this was tested.  I can merge it if the `boot.lua.h` change gets edited out (since that file was removed).

mcclure:
Fixed?

PR #131, opened 2019-07-07
ref.h compiles in C++

mcclure:
This commit is pretty out there and if you want to deny it I'd be fine with that. This replaces the ref.h solution in my other PR.

It turns out there are a set of GCC atomics that are supported by every compiler I can find that supports C11 atomics, but which do not have the C++ incompatibility. This patch uses the GCC atomics where possible but keeps C11 as a fallback (with a helpful error message if the C11 header is illegally included in C++).

Check for GCC version >=4.7 [covers GCC or, theoretically, Intel C Compiler] or __has_builtin(__atomic_add_fetch) [covers Clang]. If either of these are found, use the GCC atomic builtins instead of the C11 builtins.

Also: Gates Microsoft atomics on _MSC_VER, not _WIN32, and defines one of LOVR_HAS_MSVC_ATOMICS, LOVR_HAS_GCC_ATOMICS or LOVR_HAS_C11_ATOMICS where appropriate.

### DIFF
--- a/src/core/ref.h
+++ b/src/core/ref.h
@@ -3,25 +3,45 @@
 
 #pragma once
 
-#if defined(LOVR_ENABLE_THREAD) && defined(_WIN32)
+#ifndef LOVR_ENABLE_THREAD
+
+// Thread module is off, don't use atomics
+
+typedef uint32_t Ref;
+static inline uint32_t ref_inc(Ref* ref) { return ++*ref; }
+static inline uint32_t ref_dec(Ref* ref) { return --*ref; }
+
+#elif defined(_MSC_VER)
+
+// MSVC atomics
 
 #include <intrin.h>
 typedef uint32_t Ref;
 static inline uint32_t ref_inc(Ref* ref) { return _InterlockedIncrement(ref); }
 static inline uint32_t ref_dec(Ref* ref) { return _InterlockedDecrement(ref); }
 
-#elif defined(LOVR_ENABLE_THREAD)
+#elif (defined(__GNUC_MINOR__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 7))) \
+   || (defined(__has_builtin) && __has_builtin(__atomic_add_fetch) && __has_builtin(__atomic_sub_fetch))
+
+// GCC/Clang atomics
+
+typedef uint32_t Ref;
+static inline uint32_t ref_inc(Ref* ref) { return __atomic_add_fetch(ref, 1, __ATOMIC_SEQ_CST); }
+static inline uint32_t ref_dec(Ref* ref) { return __atomic_add_fetch(ref, 1, __ATOMIC_SEQ_CST); }
+
+#else
+
+// No known compiler-specific atomics-- fall back to C11 atomics
+
+// stdatomic.h doesn't work in c++ (except on Android, where it is fine)
+#if !defined(__ANDROID__) && defined(__cplusplus)
+#error "The header core/ref.h cannot currently be included from C++ when threading is enabled with this compiler. Either remove your ref.h include from any C++ files, or rebuild LOVR with -DLOVR_ENABLE_THREAD=OFF"
+#endif
 
 #include <stdatomic.h>
 typedef _Atomic(uint32_t) Ref;
 static inline uint32_t ref_inc(Ref* ref) { return atomic_fetch_add(ref, 1) + 1; }
 static inline uint32_t ref_dec(Ref* ref) { return atomic_fetch_sub(ref, 1) - 1; }
-
-#else
-
-typedef uint32_t Ref;
-static inline uint32_t ref_inc(Ref* ref) { return ++*ref; }
-static inline uint32_t ref_dec(Ref* ref) { return --*ref; }
 
 #endif
 


bjornbytes:
Curious about some stuff: Why `_MSC_VER` instead of `_WIN32`?  Everywhere else in the codebase uses `_WIN32`.  Is there a practical reason for the `LOVR_HAS_X_ATOMICS` defines?

mcclure:
> Curious about some stuff: Why _MSC_VER instead of _WIN32?

As I understand, Interlocked* are not Windows features. They are [MSVC compiler intrinsics](https://docs.microsoft.com/en-us/cpp/intrinsics/intrinsics-available-on-all-architectures?view=vs-2019). Hypothetically, someone could compile for Windows using GCC (mingw or MSYS2) or clang. In this case, the compiler would support the GCC/clang atomic intrinsics, but would not necessarily support Interlocked*. (I asked around on Twitter about this and I have run across a couple signs it might be *possible* to make clang/mingw support Interlocked*, but you might have to do extra things like enable compatibility modes.)

In general, because mingw does exist out there in the world, it's better to use _WIN32 when the thing you are testing for is actually Windows related, and _MSC_VER if you are testing because of a MSVC compiler feature or bug. Most of our #ifs are _WIN32.

This is probably low priority unless we have actually tested on mingw or clang/windows and seen that it fails without swapping _WIN32 in this case.

> Also, is there a practical reason for the LOVR_HAS_X_ATOMICS defines?

No. I guess I thought, if some other code somewhere else in the codebase wanted to use an atomic, it would be good if they didn't have to go through the dance ref.h does. But probably ref.h is not the correct place for that-- if these defines are kept probably it would be better to stick the HAS_X_ATOMICS in their own header. I'd also be fine to remove them.

mcclure:
Should I remove the LOVR_HAS_X_ATOMICS defines?

bjornbytes:
I think if there isn't an immediate need for anyone to know about the atomic method in use (I would be scared if this were the case), it shouldn't be included.

Also, I sketched out a different structure for the ifdefs that may make more sense.  It

- Avoids repetition of `LOVR_ENABLE_THREAD`
- Is way more selective about when it tries to use the C11 atomics, to prevent confusing compile errors.  Still has the final `#error` which should cover the C++ conditional.
- Uses the gcc/clang atomics proposed here so that threads are more likely to work on a modern compiler, even when compiling a translation unit as C++.

```
// If the thread module is off, just use ints
#ifndef LOVR_ENABLE_THREAD

...

// Otherwise, if C11 is supported and stdatomic.h is implemented, prefer that
#elif __STDC_VERSION__ >= 201112L && !__STDC_NO_ATOMICS__

...

// Otherwise, try to use GCC/Clang atomics
#elif (defined(__GNUC_MINOR__) && (__GNUC__ > 4 || (__GNUC__ == 4 && __GNUC_MINOR__ >= 7))) \
  || (defined(__has_builtin) && __has_builtin(__atomic_add_fetch) && __has_builtin(__atomic_sub_fetch))

...

// Otherwise, windows (I would also be okay just doing _WIN32 even though it's
// less correct, since no thought has been put into supporting mingw until now)
#elif _MSC_VER

...

// Otherwise, failure (includes the current C++ edge case I think)
#else
#error "sorry you need atomics to use threads and we tried really hard to use them"
#endif
```

What do you think?  I guess it's just moving the existing `#else` up top, and adding a conditional to more explicitly detect C11.

mcclure:
So the problem with that approach is `__STDC_VERSION__` is not defined in C++. So let's consider two files in a hypothetical lovr fork, `filesystem.c` and `extension.cpp`. `filesystem.c` will be compiled with the C compiler and __STDC_VERSION__ will be true. `extension.cpp` will be compiled with the C++ compiler and __STDC_VERSION__ will not be set, so either you will take the clang path or there will just be an error at this point. Assuming we don't just error out, it will now be the case the C files will use C11 atomics and the C++ files will use the GNU atomics. This is the worst case scenario because the C11 atomics and the GNU atomics might not be interchangeable (they might be; who knows? but by the standard it's unsafe) and a pointer might be passed from `extension.cpp` to `filesystem.c` and be `lovrRetain`ed in both and then who knows what will happen.

The thing that makes this entire problem difficult is (1) there is **no** way to detect via #defines (except for __ANDROID__, which means that C11 atomics simply work with C++) that it is safe to interchange C11 atomics with C++ atomics, and (2) because we are in the preprocessor the C files and C++ files "don't know" about each other— that is, there is no way in C to check "are there any c++ files that included ref.h…?". The goal of this particular patch was to make GNU atomics take precedent over C11 because they are reliably cross compatible with C++ and we can(?) assume if you're using a GNU-compatible C compiler you're using a GNU-compatible C++ compiler.

Making Lovr incompatible with C++ would of course be a reasonable choice to make, honestly, as long as it were an intentional choice.

Anyway, I pushed a commit here which uses your "check no threads first" ordering and removes the unnecessary which-atomics defines.

bjornbytes:
Ok, that makes sense, I'm probably going to merge this in a bit.  I'm still curious about other options, but at this point it doesn't seem worth spending more time on further analysis right now.

- Would it make things easier if I just stop using C11 atomics?  That way we would only be using the compiler-specific intrinsics, which will be compatible with the compiler whether it's compiling as C or C++.
- What happens if `#include "core/ref.h"` is wrapped in `extern C` in C++?  Does that in any way help the C/C++ translation units use the same atomic implementation?

mcclure:
> Would it make things easier if I just stop using C11 atomics? 

The effect of this patch is you stop using C11 atomics on all known platforms. I think including the C11 path, for UNTHINKABLE WEIRDO SITUATIONS, is still worth it.

> if #include "core/ref.h" is wrapped in extern C in C++? Does that in any way help the C/C++ translation units use the same atomic implementation?

Generally no. extern C is about calling convention, but the language being compiled in the extern C block is still C++. My understanding is the _Atomic incompatibility is about the actual language feature _Atomic not being supportable.

bjornbytes:
Ok, merging, thanks for spending time to make this header more robust!!

PR #132, opened 2019-07-10
Fix lovr.filesystem.append()

mcclure:
Previously, this program
```
function lovr.update(dt) lovr.filesystem.append("/test123", lovr.timer.getTime()) end
```
would fail in lovr because lovrFileWrite required the file to be in write mode (not append)

The `src/modules/filesystem/filesystem.c` change has no effect. I only changed it because I **assume** this was done by mistake, and it was confusing me.

### DIFF
--- a/src/modules/filesystem/file.c
+++ b/src/modules/filesystem/file.c
@@ -41,7 +41,7 @@ size_t lovrFileRead(File* file, void* data, size_t bytes) {
 }
 
 size_t lovrFileWrite(File* file, const void* data, size_t bytes) {
-  lovrAssert(file->handle && (file->mode == OPEN_READ || file->mode == OPEN_WRITE), "File must be open for writing");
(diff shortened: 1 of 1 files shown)


src/modules/filesystem/filesystem.c
```diff
@@ -307,5 +307,6 @@ size_t lovrFilesystemWrite(const char* path, const char* content, size_t size, b
   }
 
   size_t bytesWritten = lovrFileWrite(&file, (void*) content, size);
-  return lovrFileDestroy(&file), bytesWritten;
+  lovrFileDestroy(&file);
+  return bytesWritten;
```
bjornbytes: This isn't a mistake, it's using the comma operator to cause a side effect and then return a value.  I kinda like using this idiom in certain situations.

mcclure:
Well, I don't like that, but if it's intentional then it's not appropriate to fix it in this patch anyway. I rolled that back and left only the bugfix.

bjornbytes:
Ok, thanks for the fix!

PR #134, opened 2019-07-11
Fix CMakeLists.txt after animator removed

mcclure:
animator.c was removed but only the tupfile was updated


bjornbytes:
Ah, thanks!

PR #135, opened 2019-07-16
Vector rework (again);

bjornbytes:
Main changes:

- Add `vec2` and `vec4`.
- Add properties and swizzling, e.g. `vec3.zx = vec2(x, y)`.
- `lovr.math.vec3` etc. return temporary vectors.  They can only be used in the frame that they're created in, otherwise there will be an error.
- New `lovr.math.newVec3` etc. functions return permanent vectors that are garbage collected and can be used across frame boundaries.
- Add global `vec3` etc. functions, controlled by `t.math.globals`.
- Remove `Pool` objects.  There is now one global Pool per thread.  Pools use integer handles with type/generation/index bits.  In LuaJIT this seems to be a lot faster than allocating structs.  Pools have a fixed maximum size of 65536 floats, so if you run out of space in a frame you gotta manually drain the pool periodically.
- Re-add `lovr.math.drain` function for draining the pool.
- Add `quat:direction` and `quat:conjugate`.
- Add `mat4:fov` and `mat4:lookAt`.
- `mat4:set` takes more variants.
- Remove `lovr.math.orientationToDirection` and `lovr.math.lookAt`.
- Consolidate all the C vector bindings into `l_vectors.c`.

idk, it feels so complicated.  The vec2/vec3/vec4 is just a bunch of repeated code, plus everything is duplicated again for the LuaJIT version!  I'm not sure if the permanent vectors are a good idea (constantly have to do loads/stores.  maybe tables would just be better?  garbage collection is the enemy no matter what).  But I want to try this out for a while and see how it feels to use.


bjornbytes:
Ok let's give it a go........

PR #136, opened 2019-07-16
Don't reset camera position of desktop driver on reset

mcclure:
This doesn't have to be merged, but: This makes it so when you reset, for example using F5 or lovr-lodr, the fake camera in "desktop" does not reset positions. This is IMO more convenient because the fake camera isn't an "in-game" thing, rather it's a simulation of a vr headset. VR headsets don't reset their positions when the game does…


bjornbytes:
This is a cool idea!

PR #142, opened 2019-08-28
Miniaudio fixes

mcclure:
- Fixed compile on Android
- Updated to 0.9.7


mcclure:
Addressed comments, no longer stuck in one channel, the decoder outputs mono (enforced when AudioSource created) and then the callback copies to all channels

PR #143, opened 2019-08-29
Shadow samplers;

bjornbytes:
This was a bit rushed, going to look over it later.


bjornbytes:
I think I'm supposed to reverse the compare mode (currently doing this for stencil test).

PR #144, opened 2019-09-03
FFS

bjornbytes:
Rips out PhysFS in favor of handwritten filesystem stuff.

Changes:

- Up to 8 archives.
- `setIdentity` can only be called once, and can't be called with `nil`.  The idea is that you set your write directory once at startup and then it is immutable.
- Mount/unmount are no longer thread-safe.
- No memory allocations, except for the actual data read from files
- Filenames can't have `..` in them (previously they could, as long as they didn't have a parent directory path component).
- Fix some buffer overruns (probably added a few, for good measure)
- Fix issue where directory listings could contain duplicates
- Use XDG data path instead of config path and respect environment variable.
- Fix potential crash when reading passwd entry.

Todo:

- [ ] Windows
- [ ] Zip
- [ ] getDirectoryItems deduping (thinking of using Lua for this, but need to profile)


bjornbytes:
I have so many new ideas/discoveries on this that I'm going to start over!  FFS.

PR #146, opened 2019-09-04
model:getMaterialName(idx)

mcclure:
This is not efficient (and maybe should be documented as such?) but I found it useful for debug, like

    local materialCount = model:getMaterialCount()
    print("Model materials: ", materialCount)
    for i=1,materialCount do
        print(i, model:getMaterialName(i))
    end

### DIFF
--- a/src/api/l_model.c
+++ b/src/api/l_model.c
@@ -96,6 +96,25 @@ static int l_lovrModelGetMaterial(lua_State* L) {
   return 1;
 }
 
+static int l_lovrModelGetMaterialName(lua_State* L) {
+  Model* model = luax_checktype(L, 1, Model);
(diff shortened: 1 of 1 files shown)


bjornbytes:
Yeah, we really need something like this, for materials/nodes/animations.

One way to make it more efficient would to have it return a table of all the names, which is convenient to loop over or index into.  This is somewhat bad because it generates garbage.

Or names could be attached to the materials in the ModelData.  But this requires a bit more work to implement.

mcclure:
It's possible it does not need to be efficient… I don't know why a method such as this would be called *frequently*. For the same reason generating a table of the names could also be okay (it generated garbage, but that's acceptable because you only need to call it once per model)

PR #147, opened 2019-09-11
Miniaudio: Add optional Oculus Audio spatializer, make the dummy spatializer good

mcclure:
Multiple changes. Some known issues (see below).

- Adds locking to l_source functions such as seek()— was causing noise bursts
- Adds source.getPosition, source.setPosition
- Adds source.rewind
- Support a LOVR_DEBUG_AUDIOTAP #define in audio.c— set this and it will save a float record of everything played to the speakers. Only really useful when debugging the audio callback itself but very useful at those times.
- Add a mixer to the dummy spatializer. You can now play an arbitrary number of sounds at once.
- Add pop correction to the dummy spatializer. If when you stop a sound, the final played sample of the sound is at a DC offset far from zero, the user will experience the removal of the sound from the mix as a popping sound, which can be very loud. The pop correction slowly ramps down the DC-offset “residue” to zero at a rate which should be inaudible. In my testing this greatly improved the sound with my “break it” test (which stops sounds frequently).
- Adds support for the Oculus Audio spatialization library. (There is a corresponding change to lovr-oculus-audio which I will push later.) This is a complicated change which required making some changes to support, first off, Oculus Audio supports only a fixed number of sounds at once so there is a mechanism for overriding sounds if you play too many at a time. Second off, Oculus Audio can only process audio in fixed sized chunks and those chunks can’t change size over time, so there’s a regularization layer ahead of handler() with oculus audio that chops up or buffers as needed to make handler() always handle the same frame size.
- Some lovrAudio utility functions to support the above.
- One improved error message in SoundData
- Removes an unused file named source.c— seems to have been added by accident, possibly by me

**Known issues**

- Oculus Audio support does not work on Windows with Oculus driver, even though it perfectly well could
- Oculus Audio support does not seem to work right when you turn your head to the side. I think there is some kind of coordinate system mismatch. Seeking help on this from Bjorn.
- Oculus Audio and the dummy mixer are both deeply intertwined with the handler code, there are a lot of ifdefs and it isn’t great.
- In the dummy spatializer, the pop correction is not as good as it could be. If several sounds stop at once, the combined residue will fall off faster than the intended falloff rate (and this could produce an audible artifact). In discussion with Bjorn, we came to the conclusion a more reasonable way to handle this for the current system would be to have source:stop() not halt the sound immediately but rather wait for a zero crossing.
- The Oculus Audio spatializer does not use the pop correction, because in my early testing it did not seem to need it. After more testing I find it does need it.
- In Oculus Audio, the override when you try to play too many sounds at once is not “fair”. It would be easy to pick a strategy such as always overriding the oldest sound.
- Oculus Audio requires all buffers to be aligned to 16 bytes, but I don't do anything to guarantee this because in my testing on Android, arr_t coincidentally happened to always produce buffers aligned to 16 bytes. Probably this shouldn't be trusted completely. There is a bottleneck in audio.c that would be an easy place to add an alignment check.
- lovrSpatializerInit and lovrSpatializerRealInit could be merged
- Slightly horrifyingly, Oculus Audio requires you to use a struct that it does not define. It assumes that if you are using Oculus Audio you are also using an Oculus headset API and have included those headers. However because of build complexities I have gone to great lengths to quarantine the oculus_mobile driver completely from the actual Oculus Mobile SDK, so we do not have access to those headers in the actual lovr binary. The result is in order to communicate the current headset pose to Oculus Audio we have to include a “shim” header with a chunk of Oculus’s math headers pasted in, and then oculus_mobile.c has to go to a great deal of work to convert the BridgeLovr types, which were originally converted from Oculus math structs, BACK into Oculus math structs. As frustrating as this is, it probably cannot be fixed unless either I get a pure-CMake build of the Oculus Mobile SDK ( see: https://developer.oculus.com/bugs/bug/500588460734833/ ) or Bjorn’s pure-Tup Oculus Mobile build gets working.


bjornbytes:
I don't think I'm ready to review/merge this right now, but I think it will be very valuable to have around once more thought is put into finalizing the audio rework and doing effects/spatialization, especially understanding all of the quirks surrounding the Oculus SDK.

mcclure:
Sure. If you want me to pull out the mixer part only (without the oculus audio) and make it a PR by itself let me know

PR #148, opened 2019-09-11
Make lovrViewID signed on Oculus Mobile (fix "mask" example)

mcclure:
The "mask" example was failing on Oculus Mobile on the line of glsl:
if (lovrViewID == 1) {
because lovrViewID was unsigned and 1 was signed. One way to fix this would be to replace 1 with 1u as that is unsigned, but this would be the wrong fix because lovrViewID is in fact signed on all platforms other than Oculus Mobile. lovrViewID was depending on platform defined to either gl_ViewportIndex (signed), a signed uniform, or gl_ViewID_OVR (unsigned). The solution is to place an explicit cast in the multiview definition of lovrViewID so that it is signed on all platforms.


bjornbytes:
Only thing more fun than putting dots and `u`s after all the numbers is explicit constructors...

PR #151, opened 2019-09-24
Throw lua errors in the error handler

sophiabaldonado:
I discussed this with @bjornbytes offline, it should now show the lovr error screen and message on oculus mobile instead of crashing in the cases I was seeing.


bjornbytes:
Going to run a few test cases to make sure that error behavior is the same on desktop and android, across all the different ways errors can happen (load vs. update vs. draw, Lua errors vs. C errors).

bjornbytes:
Ok I did find one problem but it's pretty minor, going to take care of it after merging.

PR #153, opened 2019-09-26
Vibrate function for Oculus Quest

mcclure:
This commit is a fist pass. It only does "simple" vibrate (hard on/off). It supports duration but not frequency. Supporting frequency will probably require Oculus to document their "haptic buffer" function, which currently is IMO not clear enough to work with. I have filed a bug about this: https://developer.oculus.com/bugs/bug/533528907451560/

However even this incomplete version of vibrate probably gives us feature parity with some other platforms.

This commit requires a corresponding commit in the lovr-oculus-mobile repo (see mobile-1.patch, attached:
[mobile-1.txt](https://github.com/bjornbytes/lovr/files/3660168/mobile-1.txt) )

See also attached my test program (as written requires lovr-ent) 
[vibrate.txt](https://github.com/bjornbytes/lovr/files/3660170/vibrate.txt)

### DIFF
--- a/src/modules/headset/oculus_mobile.c
+++ b/src/modules/headset/oculus_mobile.c
@@ -18,6 +18,7 @@
 static struct {
   BridgeLovrDimensions displayDimensions;
   BridgeLovrDevice deviceType;
+  BridgeLovrVibrateFunctionPtr vibrateFunction;
   BridgeLovrUpdateData updateData;
 } bridgeLovrMobileData;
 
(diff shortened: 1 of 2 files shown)


mcclure:
OH, I just remembered, an oddity of the oculus mobile vibrate API is you're not supposed to call the vibrate function more than once per frame (per controller).

Right now if you call more than once per frame it really does call more than once per frame. So that's undefined behavior.

If you think it's important to support calling more than once per frame, I could just set it so it defers actually invoking the vibrate function until end of frame and if you call more than once it just overwrites the first call.

PR #156, opened 2019-10-08
Fix World:Collide to use new luax_checkshape

programmer02:
This appears to have been missed in commit 22fe333, and makes it impossible to use a custom collision resolver (the code for a default collision resolver in the docs results in a "bad argument #1 to 'collide' (Shape expected, got userdata)")


bjornbytes:
Thanks!

bjornbytes:
Found a similar issue with `Collider:addShape` and `Collider:removeShape`, pushed a fix.  Thanks again.

PR #160, opened 2019-10-15
Cast objc_msgSend before calling it

nevyn:
Calling objc_msgSend with its vararg signature will only work most
of the time, and fail very weirdly when it doesn't. It's weird,
but you're supposed to always cast it to the signature of the
selector you're calling.

I think this is even required when building on Catalina, or for arm64?


bjornbytes:
Thanks!

PR #161, opened 2019-10-16
Potential for drawing "visibly" while populating the stencil buffer

mcclure:
Add one more argument to lovr.graphics.stencil(); set true and the colors in the stencil draw will be allowed to draw (instead of being masked out and only the stencil buffer being drawn to)

Will need to be documented at some point

### DIFF
--- a/src/modules/graphics/opengl.c
+++ b/src/modules/graphics/opengl.c
@@ -820,16 +820,7 @@ static void lovrGpuBindPipeline(Pipeline* pipeline) {
         glEnable(GL_STENCIL_TEST);
       }
 
-      GLenum glMode = GL_ALWAYS;
-      switch (state.stencilMode) {
-        case COMPARE_EQUAL: glMode = GL_EQUAL; break;
-        case COMPARE_NEQUAL: glMode = GL_NOTEQUAL; break;
-        case COMPARE_LESS: glMode = GL_GREATER; break;
-        case COMPARE_LEQUAL: glMode = GL_GEQUAL; break;
-        case COMPARE_GREATER: glMode = GL_LESS; break;
-        case COMPARE_GEQUAL: glMode = GL_LEQUAL; break;
-        default: break;
-      }
+      GLenum glMode = convertCompareMode(state.stencilMode);
 
       glStencilFunc(glMode, state.stencilValue, 0xff);
       glStencilOp(GL_KEEP, GL_KEEP, GL_KEEP);
@@ -1225,7 +1216,7 @@ void lovrGpuPresent() {
   memset(&state.stats, 0, sizeof(state.stats));
 }
 
-void lovrGpuStencil(StencilAction action, int replaceValue, bool visibleDraw, StencilCallback callback, void* userdata) {
+void lovrGpuStencil(StencilAction action, int replaceValue, bool visibleDraw, CompareMode stencilMode, StencilCallback callback, void* userdata) {
   lovrGraphicsFlush();
   if (!visibleDraw)
     glColorMask(GL_FALSE, GL_FALSE, GL_FALSE, GL_FALSE);
@@ -1246,7 +1237,9 @@ void lovrGpuStencil(StencilAction action, int replaceValue, bool visibleDraw, St
     default: lovrThrow("Unreachable");
   }
 
-  glStencilFunc(GL_ALWAYS, replaceValue, 0xff);
+  GLenum glMode = convertCompareMode(stencilMode); // Should usually be COMPARE_NONE
+
+  glStencilFunc(glMode, replaceValue, 0xff);
   glStencilOp(GL_KEEP, GL_KEEP, glAction);
 
   state.stencilWriting = true;
--- a/src/modules/graphics/opengl.c
+++ b/src/modules/graphics/opengl.c
@@ -1225,9 +1225,10 @@ void lovrGpuPresent() {
   memset(&state.stats, 0, sizeof(state.stats));
 }
 
-void lovrGpuStencil(StencilAction action, int replaceValue, StencilCallback callback, void* userdata) {
+void lovrGpuStencil(StencilAction action, int replaceValue, bool visibleDraw, StencilCallback callback, void* userdata) {
   lovrGraphicsFlush();
-  glColorMask(GL_FALSE, GL_FALSE, GL_FALSE, GL_FALSE);
+  if (!visibleDraw)
+    glColorMask(GL_FALSE, GL_FALSE, GL_FALSE, GL_FALSE);
 
   if (!state.stencilEnabled) {
     state.stencilEnabled = true;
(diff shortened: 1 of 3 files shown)


mcclure:
Second branch allows you to apply a stencil-test rule to your drawing while you are also drawing into the stencil buffer

src/modules/graphics/opengl.c
```diff
@@ -820,16 +820,7 @@ static void lovrGpuBindPipeline(Pipeline* pipeline) {
...
-      switch (state.stencilMode) {
-        case COMPARE_EQUAL: glMode = GL_EQUAL; break;
-        case COMPARE_NEQUAL: glMode = GL_NOTEQUAL; break;
-        case COMPARE_LESS: glMode = GL_GREATER; break;
-        case COMPARE_LEQUAL: glMode = GL_GEQUAL; break;
-        case COMPARE_GREATER: glMode = GL_LESS; break;
-        case COMPARE_GEQUAL: glMode = GL_LEQUAL; break;
-        default: break;
-      }
+      GLenum glMode = convertCompareMode(state.stencilMode);
```
bjornbytes: Hey so I'm not sure if `convertCompareMode` will work here because it does the sensible thing where e.g. `COMPARE_LESS` maps to `GL_LESS`.

But notice how this switch statement here is doing more of a "reverse compare mode" thing where `COMPARE_LESS` maps to `GL_GREATER`.  This is because it's more intuitive to think about it this way in the LÖVR API:

- In LÖVR `lovr.graphics.setStencilTest('less', 3)` means "only render pixels where stencil buffer is less than 3".
- In OpenGL `glStencilFunc(GL_LESS, 3, 0xff)` means "only render pixels where 3 < stencil buffer".

So we reverse it.

I think I'm going to need to do this reversed-compare-mode thing in another place soon in #143 so feel free to create a `convertCompareModeReversed` if you'd like, otherwise it's fine to just revert this change.

src/modules/graphics/opengl.c
```diff
@@ -1245,15 +1237,18 @@ void lovrGpuStencil(StencilAction action, int replaceValue, StencilCallback call
     default: lovrThrow("Unreachable");
   }
 
-  glStencilFunc(GL_ALWAYS, replaceValue, 0xff);
+  GLenum glMode = convertCompareMode(stencilMode); // Should usually be COMPARE_NONE
```
bjornbytes: I'm wondering if it's possible to just respect people's existing `setStencilMode` state when inside the `lovr.graphics.stencil` callback.  Basically just not touching `glStencilFunc` at all?

It seems like it could be a way to allow the use of stencil-test-while-stenciling without adding an argument, without adding an extra argument.  But I could be missing something.

mcclure: We could do that, or we could make it legal to call setStencilMode inside of a stencil callback. To me, stencils are very confusing and there is something a *little* bit less confusing about treating a stencil callback like a "scope" (IE, the stencil mode you use/set inside this stencil call somehow does not affect the outside world)

Making stencil() calls inherit, or able to set, their own stencil mode would at least make the stencil() parameters less confusing. (Would we remove the stencilaction/value arguments in that case tho?)

bjornbytes: Removing the stencilFunc would allow people to set the stencil test inside the callback, but has the disadvantage of making it possible to forget to set it and inherit unwanted state.  Allowing .setStencilTest inside the callback but resetting it afterwards works too, to make it more scope-y (would it also get reset to "nil" when entering the callback?).

Would still need action/value I think, since `setStencilTest` only determines the test used for pass/fail, not how the stencil buffer is updated.

Also I think .stencil can be better about documenting its behavior on how it ignores/allows/resets the stencil test.

Agh stencils are VERY confusing.

mcclure: I think that the "value to test" and the "value to set" are the same, it's just the second argument to glStencilFunc, right?

bjornbytes: You're right, they're the same.  I was confused because in the lovr.graphics API there are 2 different concepts using a stencil value, but under the hood they both use glStencilFunc.

mcclure: Okay.
Anyway: Do you have a specific ask for how this patch should be different to get it accepted? We seem to have tossed around several ideas at this point.

bjornbytes: Ok.

- So for the `drawVisible`, I think I'd like to just start exposing `lovr.graphics.setColorMask` directly.  This way, people always have control over when color channels are written, which should cover this situation as well as other ones.
- As for the second bit, I'm still just confounded.  I think seeing an example of how you're using it would help a lot.
  - My attempt to explain my confusion: If OpenGL puts us in a situation where "the value to write to stencil buffer" has to be the same as "the value to use for stencil test", then I am unable to express something like "set stencil test to greater than 17 and when that passes replace the stencil value with 3".  Stencil actions like increment/decrement can still work okay, but testing against 17 means I must replace with 17 as well.
    - So we're unable to let you enable a stencil test while also writing to the stencil buffer, unless you're okay with the two values being the same.  And that's kind of awkward.

mcclure: "then I am unable to express something like "set stencil test to greater than 17 and when that passes replace the stencil value with 3… testing against 17 means I must replace with 17 as well". As far as I know, this is literally how the OpenGL stencil feature works, and I agree that's not very useful. Unfortunately useful or no it really does just appear to be how OpenGL works. (You might be able to do some fancy things with `glStencilMask` and `GL_INVERT`, but that sounds confusing and I would not want to use such a feature…) GL_INCR is, as you note, still useful for various reasons.

What I, specifically, want to do (am already doing in my branch) is this:
- Stencil is initially cleared to 0
- Reference value when drawing is 1
- Test type is **GL_NOTEQUAL** 
- If test fails, GL_KEEP (this is probably irrelevant)
- If test succeeds, GL_REPLACE
Once I have set these rules, I draw a series of 80% transparent polygons.  
Reason: I am drawing shadows cast by objects as polygons. The polygons sometimes overlap onto a single pixel. I *do not want* overdraw to result in drawing multiple times, as since the polygons are semitransparent, this will be visible. So my rule is "Draw this shadow polygon, but only onto pixels where I HAVE NOT already drawn a polygon".

glColorMask exposed would be useful for a number of reasons. I note that since glStencil implicitly sets glColorMask, it would be *nice* to have a way of saying "here is the glColorMask to use while running this stencil". But I'm not sure this is really different from just calling `lovr.graphics.colorMask`…

bjornbytes: Ok.  I think that because of this mismatch between what LÖVR tries to expose and what OpenGL provides, I'd rather not expose the "set stencil test while inside stencil callback" feature; it can't be done in a satisfying way.

For the shadow thing specifically, it might be possible to achieve the same effect with depthtest/blendmode/canvas or something?  Meh, who knows.

Anyway, I definitely want to add in color mask.  I'll make a note to add it for this next version, but it'll probably be a little bit before I start on it.  If you want to get to it first then just reopen the PR!

Thanks for the discussions.

mcclure: "I'd rather not expose the "set stencil test while inside stencil callback" feature; it can't be done in a satisfying way."

Sorry, what I mean is setting the **color** mask inside the stencil callback. Right now you change the color mask inside the stencil mask, automatically. So once you add the color mask, the conventional way to get visible pixels inside the stencil callback will be to call the stencil (which turns the color mask off) and then setColorMask() them back on in the first line of the callback. It's not a BIG deal but that's a *little* weird.

bjornbytes: It would be possible to not touch color mask at all and require people to disable the color mask themselves, however this would break existing stencil code and break LÖVE compatibility.  It does seem the common case is to want color mask disabled, but I agree there's something weird about LÖVR forcing that on you.  I could really go either way on it.

PR #164, opened 2019-11-03
Prioritize loading game from command line over embedded

nevyn:
so that it’s possible to override embedded game. This is handy when making a custom build of lovr, but still want to iterate quickly with lodr on the lua stuff.


bjornbytes:
Hey, we talked about this in Slack but I forgot to resolve the PR.

PR #168, opened 2019-11-19
Add lovrMain macro to shaders (no breaking change);

bjornbytes:
A quick proposal to offer an alternative way to write shaders.

Instead of:

```lua
shader = lovr.graphics.newShader([[
  vec4 position(mat4 projection, mat4 transform, vec4 vertex) {
    return projection * transform * vertex;
  }
]], [[
  vec4 color(vec4 graphicsColor, sampler2D image, vec2 uv) {
    return graphicsColor * texture(image, uv);
  }
]])
```

This patch adds the ability to write the following:

```lua
shader = lovr.graphics.newShader([[
  vec4 lovrMain() {
    return lovrProjection * lovrTransform * lovrVertex;
  }
]], [[
  vec4 lovrMain() {
    return lovrGraphicsColor * texture(lovrDiffuseTexture, lovrTexCoord);
  }
]])
```

I find that this leads to a productivity boost because I don't have to go to the docs to copy-paste the function signatures or type them over and over again.


mcclure:
This seems less confusing!

I worry a LITTLE about lovrVertex having an exe cost but of course I can work around that by saying vec4 v = lovrVertex; so whatever.

I didn't know you could use () in glsl defines?!

PR #171, opened 2019-12-18
Remove physfs from .gitmodules

mcclure:
The Physfs dep has been removed, but it is still listed in .gitmodules. This is confusing my third party git client very badly.


bjornbytes:
Oops, thanks!

PR #174, opened 2019-12-18
Android fixes. Major change is lovrMain() {} is now lovrMain {}

mcclure:
The oculus drivers choke on `#define lovrMain()`

I have partially but not fully tested this patch, might make sense to give me until tomorrow so I can give it a more careful once-over. Or we could merge it now and I could send you more fixes as I find them


bjornbytes:
Thanks!

Issue #158, opened 2019-10-10
Microphone:getData Memory Behavior


bjornbytes:
Currently when you get samples from a `Microphone` using `:getData`, it always creates a new `SoundData` object.

Seemingly this is "bad" because it uses more memory and does more allocation work to create and destroy/garbage-collect all of these objects, even in situations where you only need to use the samples temporarily (performing one-time analysis on chunks, stream it to another API/system, etc.).

Can we add an interface that lets you pass in your own Blob or SoundData object that acts as a destination for the microphone samples?  That way one object could be reused as the destination over and over again.  It would be optional but gives you the ability to manage the memory better if you're willing to do so.

bjornbytes:
If the Blob/SoundData is too small or incompatible with the microphone's capture settings, I think that would be an error.

nevyn:
Another alternative design is one of my favorite APIs in iOS, CMSampleBufferPool. Sample buffers are reference counted, and when spawned from a pool, this is overridden so that a count of 0 will return the buffer to the pool. This means the producer of audio sample buffers will create as many in-flight buffers as needed, and the consumer can hold on to them as long or short as they want. by setting a limit on the pool, that also provides a really nice backpressure mechanic.

anyways just my 2c :)

bjornbytes:
`Microphone:getData` now takes `([sampleCount], [soundData], [offset])` !

PR #180, opened 2020-01-02
Real-time raw AudioStreams

nevyn:
Gives LÖVR the ability to play runtime-generated audio (in addition to playing audio from ogg files on disk). This will be used in Alloverse to play voip audio between multiplayer users, but could also be used to generate audio or music programmatically.

This PR adds two APIs:
* `lovr.audio.newAudioStream(bufferSize, channelCount, sampleRate)` creates a "raw" AudioStream which can only get audio from `AudioStream:append()` and not from decoding a file on disk
* `AudioStream:append(blob)` appends raw PCM sound data for playback ASAP.

Tasks remaining before PR is mergeable:

- [x] Implement Lua API
- [ ] Create a queue of blobs to reduce malloc churn (or use a ringbuffer)
- [x] add upper limit for queued buffers

### DIFF
--- a/src/modules/audio/source.c
+++ b/src/modules/audio/source.c
@@ -165,11 +165,6 @@ void lovrSourcePlay(Source* source) {
     return;
   }
 
-  // in case we're replaying an already-used stream, make sure to rewind it if applicable
-  if (!lovrAudioStreamIsRaw(source->stream)) {
-    lovrAudioStreamRewind(source->stream);
-  }
-
   // in case we have some queued buffers, make sure to unqueue them before streaming more data into them.
   ALint processed;
   alGetSourcei(lovrSourceGetId(source), AL_BUFFERS_PROCESSED, &processed);
--- a/src/modules/audio/source.c
+++ b/src/modules/audio/source.c
@@ -165,6 +165,11 @@ void lovrSourcePlay(Source* source) {
     return;
   }
 
+  // in case we're replaying an already-used stream, make sure to rewind it if applicable
+  if (!lovrAudioStreamIsRaw(source->stream)) {
+    lovrAudioStreamRewind(source->stream);
+  }
+
   lovrSourceStream(source, source->buffers, SOURCE_BUFFERS);
   alSourcePlay(source->id);
 }
@@ -285,7 +290,9 @@ void lovrSourceStop(Source* source) {
       alSourcei(source->id, AL_BUFFER, AL_NONE);
 
       // Rewind the decoder
-      lovrAudioStreamRewind(source->stream);
+      if (!lovrAudioStreamIsRaw(source->stream)) {
+        lovrAudioStreamRewind(source->stream);
+      }
       break;
     }
   }
--- a/src/modules/audio/source.c
+++ b/src/modules/audio/source.c
@@ -281,14 +281,14 @@ void lovrSourceStop(Source* source) {
 
     case SOURCE_STREAM: {
 
+      // Stop the source
+      alSourceStop(source->id);
+      alSourcei(source->id, AL_BUFFER, AL_NONE);
+
       // Empty the buffers
       int count = 0;
       alGetSourcei(source->id, AL_BUFFERS_QUEUED, &count);
       alSourceUnqueueBuffers(source->id, count, NULL);
-
-      // Stop the source
-      alSourceStop(source->id);
-      alSourcei(source->id, AL_BUFFER, AL_NONE);
 
       // Rewind the decoder
       if (!lovrAudioStreamIsRaw(source->stream)) {
--- a/src/modules/audio/source.c
+++ b/src/modules/audio/source.c
@@ -291,9 +291,7 @@ void lovrSourceStop(Source* source) {
       alSourceUnqueueBuffers(source->id, count, NULL);
 
       // Rewind the decoder
-      if (!lovrAudioStreamIsRaw(source->stream)) {
-        lovrAudioStreamRewind(source->stream);
-      }
+      lovrAudioStreamRewind(source->stream);
       break;
     }
   }
--- a/src/modules/audio/source.c
+++ b/src/modules/audio/source.c
@@ -230,6 +230,7 @@ void lovrSourceSetFalloff(Source* source, float reference, float max, float roll
 }
 
 void lovrSourceSetLooping(Source* source, bool isLooping) {
+  lovrAssert(!source->stream || lovrAudioStreamIsRaw(source->stream), "Can't loop a raw stream");
   source->isLooping = isLooping;
   if (source->type == SOURCE_STATIC) {
     alSourcei(source->id, AL_LOOPING, isLooping ? AL_TRUE : AL_FALSE);
--- a/src/modules/audio/source.c
+++ b/src/modules/audio/source.c
@@ -169,8 +169,9 @@ void lovrSourcePlay(Source* source) {
   // BEFORE user code calls source:play(). This means that some buffers may still be queued (but processed
   // and completely finished playing). These must be unqueued before we can start using the source again.
   ALint processed;
+  ALuint _unused[SOURCE_BUFFERS];
   alGetSourcei(lovrSourceGetId(source), AL_BUFFERS_PROCESSED, &processed);
-  alSourceUnqueueBuffers(source->id, processed, NULL);
+  alSourceUnqueueBuffers(source->id, processed, &_unused);
 
   lovrSourceStream(source, source->buffers, SOURCE_BUFFERS);
   alSourcePlay(source->id);
--- a/src/modules/audio/source.c
+++ b/src/modules/audio/source.c
@@ -165,11 +165,12 @@ void lovrSourcePlay(Source* source) {
     return;
   }
 
-  // in case we have some queued buffers, make sure to unqueue them before streaming more data into them.
+  // There is no guarantee that lovrAudioUpdate is called AFTER the state of source becomes STOPPED but
+  // BEFORE user code calls source:play(). This means that some buffers may still be queued (but processed
+  // and completely finished playing). These must be unqueued before we can start using the source again.
   ALint processed;
   alGetSourcei(lovrSourceGetId(source), AL_BUFFERS_PROCESSED, &processed);
-  ALuint buffers[SOURCE_BUFFERS];
-  alSourceUnqueueBuffers(source->id, processed, buffers);
+  alSourceUnqueueBuffers(source->id, processed, NULL);
 
   lovrSourceStream(source, source->buffers, SOURCE_BUFFERS);
   alSourcePlay(source->id);
--- a/src/modules/audio/source.c
+++ b/src/modules/audio/source.c
@@ -170,6 +170,12 @@ void lovrSourcePlay(Source* source) {
     lovrAudioStreamRewind(source->stream);
   }
 
+  // in case we have some queued buffers, make sure to unqueue them before streaming more data into them.
+  ALint processed;
+  alGetSourcei(lovrSourceGetId(source), AL_BUFFERS_PROCESSED, &processed);
+  ALuint buffers[SOURCE_BUFFERS];
+  alSourceUnqueueBuffers(source->id, processed, buffers);
+
   lovrSourceStream(source, source->buffers, SOURCE_BUFFERS);
   alSourcePlay(source->id);
 }
--- a/src/modules/data/audioStream.c
+++ b/src/modules/data/audioStream.c
@@ -26,7 +26,7 @@ AudioStream* lovrAudioStreamInit(AudioStream* stream, Blob* blob, size_t bufferS
   return stream;
 }
 
-AudioStream* lovrAudioStreamInitRaw(AudioStream* stream, int channelCount, int sampleRate, size_t bufferSize) {
+AudioStream* lovrAudioStreamInitRaw(AudioStream* stream, int channelCount, int sampleRate, size_t bufferSize, size_t queueLimitInSamples) {
   stream->bitDepth = 16;
   stream->channelCount = channelCount;
   stream->sampleRate = sampleRate;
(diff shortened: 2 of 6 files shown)


bjornbytes:
👍This is looking great so far!

bjornbytes:
There is potential to wombo combo with some of the discussion in

https://github.com/bjornbytes/lovr/issues/158

e.g. telling a Microphone to push its buffered samples to an AudioStream so you can avoid all of the Lua + allocation + copying overhead.  Exciting!

nevyn:
Notes for self so I know where to continue...

My test code:

```
local stream = lovr.data.newAudioStream(1, 48000)
local source = lovr.audio.newSource(stream, "stream")

function lovr.load(args)
  source:play()
end

local f = 0.0
function lovr.update(dt)
	local frac = 1.0/dt 
	local c = frac * 48000
	local sd = lovr.data.newSoundData(c, 48000, 16, 1)
	for i=0, c do 
		sd:setSample(i, sin(f*430))
		f = f + frac
	end
	stream.append(sd)
end
```

Problem: samples are not buffer to the openal source when it's play()'d. OpenAL thus sets it to "STOPPED". In lovrAudioUpdate, it's then rewound and removed. But a raw source can't be rewound, so it'll crash.

Next solution to test: if a source contains a raw stream and is stopped, just play() it again and skip removing it. Downside: how are raw streams then ever removed??

Things to consider in next hack session...

nevyn:
My solution ended up being: follow openal convention, and don't play() until buffers are queued. Call play() again if there's a gap and we end up with no buffers. 

Problem now is that 
1. Audio is playing back broken (not sure if it's my sample lua or the audio subsystem yet)
2. Console is spamming ominous warnings:

```
AL lib: (WW) alSetError: Error generated on context 000001F2D4F9B5D0, code 0xa004, "Modifying storage for in-use buffer 1"
...
AL lib: (WW) alSetError: Error generated on context 000001F2D4F9B5D0, code 0xa003, "Negative storage size -1145324614"
```

These seem REALLY bad, will have to investigate.

nevyn:
`lovrSourceStream` does not seem to have any protection against using buffers that are currently in use. `lovrAudioUpdate` unqueues processed buffers, but if `lovrAudioUpdate` is ever called before all buffers are finished processing, I think this would leave buffers queued, AND would overwrite them.

nevyn:
```
Now queueing up to 1 buffers...
Buffer 1 is queued with 4096 samples
These buffers are done processing: 1,
Now queueing up to 1 buffers...
Buffer 1 is queued with 4096 samples
These buffers are done processing: 1,
Now queueing up to 1 buffers...
Buffer 1 is queued with 4096 samples
Starting playback
Now queueing up to 4 buffers...
Buffer 1 is queued with 4096 samples
```

WHY did we end up in STOPPED without reaching processed&&isStopped in lovrAudioUpdate first??

nevyn:
So the final task is to reduce the malloc churn... My proposal from #158 stands: 

* Add a class called `SoundDataQueue`. Its constructor takes the sample rate, bit depth, channel count and buffer length its buffers should have (and an optional max queue length). 
* Instead of calling `newSoundData(...)`, you call `queue:newSoundData()`. It returns a newly constructed instance if needed, or a recycled one if available.
* SoundData gets custom refcounting. If ref reaches 0, instead of calling constructor, it checks for a destructor callback and calls it. 
* The destructor callback instructs SoundDataQueue to recycle the SoundData and put it in a list to be used later.
* When SoundDataQueue is destroyed, it removes the destructor callbacks from all its sounddatas and either releases them or frees them (depending on if refcount is 0, or more)
* It's up to calling code to use this queue. AudioStream doesn't need to know about the queue at all.

Actually... that makes it feel like it should be a separate PR. What do you think about merging this as it stands, and making the queue a separate PR @bjornbytes?

nevyn:
I just realized that as usual, I’ve mostly only tested the happy path of this code. Things I should test that I haven’t yet...

- [ ] stop
- [ ] move around 
- [ ] pause and unpause
- [ ] destructors 
- [ ] one-shot playback (e g on keydown, generate .5s audio and play())

nevyn:
Wish I could stop thinking about this PR and sleep instead... 

I just realized it would be very easy to add an optional pull-based API to this. Add a “generate x samples of audio now please” callback on stream. In decode_raw, in the check for empty queue, call this callback. In the callback the user code calls append. Back in decode_raw, check again for 0.

This could give you super low latency audio.

nevyn:
@bjornbytes okay I've run through and tested everything I can think of, and fixed a bunch of bugs. This seems rock solid to me now! Now I'm _actually_ ready for review and merge :D

mcclure:
This looks nice.

src/modules/data/audioStream.c
```diff
@@ -1,9 +1,11 @@
 #include "data/audioStream.h"
 #include "data/blob.h"
+#include "data/soundData.h"
 #include "core/ref.h"
 #include "core/util.h"
 #include "lib/stb/stb_vorbis.h"
 #include <stdlib.h>
+#include <memory.h>
```
bjornbytes: Whoa, I had never heard of `memory.h`.  Is it any different from `string.h`?  We're using that for memcpy etc. everywhere else.

src/modules/data/audioStream.c
```diff
@@ -24,13 +26,63 @@ AudioStream* lovrAudioStreamInit(AudioStream* stream, Blob* blob, size_t bufferS
   return stream;
 }
 
+AudioStream* lovrAudioStreamInitRaw(AudioStream* stream, int channelCount, int sampleRate, size_t bufferSize, size_t queueLimitInSamples) {
+  stream->bitDepth = 16;
+  stream->channelCount = channelCount;
+  stream->sampleRate = sampleRate;
+  stream->samples = SIZE_MAX;
```
bjornbytes: Is there a reason this is `SIZE_MAX`?  Could it be zero?

src/modules/data/audioStream.c
```diff
@@ -24,13 +26,63 @@ AudioStream* lovrAudioStreamInit(AudioStream* stream, Blob* blob, size_t bufferS
...
+  Blob* blob = stream->queuedRawBuffers.data[0];
+  if (blob->size <= byteCount) {
+    size_t blobSize = blob->size;
+    // blob fits in destination in its entirety. Copy over, free it and remove from start of array.
+    memcpy(destination, blob->data, blobSize);
+    lovrRelease(Blob, blob);
+    arr_splice(&stream->queuedRawBuffers, 0, 1);
+    return blobSize / sizeof(int16_t);
+  } else {
+    // blob is too big. copy all that fits, and put remainder in a new blob in its place.
```
bjornbytes: I'm wondering if it would be possible to store a cursor into the first blob, to avoid additional allocations/copies for this (potentially common) case.

Details: If the blob is too big, we can read as much as we need and advance a global cursor by that amount.  Then, next time, we can start reading at that cursor.  Any time we nom a blob completely the cursor can be reset to zero.

src/modules/data/audioStream.c
```diff
@@ -39,25 +91,71 @@ size_t lovrAudioStreamDecode(AudioStream* stream, int16_t* destination, size_t s
...
+size_t lovrAudioStreamGetQueueLength(AudioStream* stream)
+{
+  lovrAssert(lovrAudioStreamIsRaw(stream), "Queue length is only available on raw streams");
+  return stream->queueLengthInSamples;
+}
+
+bool lovrAudioStreamIsRaw(AudioStream* stream)
+{
+  return stream->decoder == NULL;
+}
```
bjornbytes: Could the new functions here use the same bracing convention as the other functions?  e.g.

```c
bool lovrAudioStreamIsRaw(AudioStream* stream) {
  return stream->decoder == NULL;
}

File: src/modules/data/audioStream.c [hunk 5]
```diff
@@ -39,25 +91,71 @@ size_t lovrAudioStreamDecode(AudioStream* stream, int16_t* destination, size_t s
...
+  arr_push(&stream->queuedRawBuffers, blob);
+  stream->queueLengthInSamples += blob->size / sizeof(int16_t);
+  return true;
+}
+
+bool lovrAudioStreamAppendRawSound(AudioStream* stream, struct SoundData* sound)
+{
+  lovrAssert(sound->channelCount == stream->channelCount && sound->bitDepth == stream->bitDepth && sound->sampleRate == stream->sampleRate, "SoundData and AudioStream formats must match");
+  lovrAudioStreamAppendRawBlob(stream, &sound->blob);
+  return true;
```
bjornbytes: I think this should return the result of `lovrAudioStreamAppendRawBlob`

src/modules/data/audioStream.c
nevyn: uhhh I don't know what vestige of my brain that I dug this out from; I just wanted memcpy! It doesn't seem to be posix or anything?? I'l change to string.h

src/modules/data/audioStream.c
nevyn: Yeah. A realtime stream potentially never ends, so SIZE_MAX tries to say "infinite" here. Using it as a shorthand for "how long would it play NOW if I didn't queue anything more" makes sense, and then we can remove getQueueLength. Yeah, I think I like it? Slightly overloaded meaning but it works :)  I'll make the change.

src/modules/data/audioStream.c
nevyn: I think that makes perfect sense. I'm not happy about this code either, it'll malloc churn like crazy in bad cases. (and it's not prettilly written :P)

I'll add a counter and avoid the memcpy/realloc!

src/modules/data/audioStream.c
nevyn: yes of course! Sorry, new line for bracket is the alloverse coding convention, I got mixed up :) you can definitely nitpick on style no worries, I'll fix!

src/modules/data/audioStream.c
nevyn: yes I think so too! good catch :)

nevyn:
Todo retest:

- [ ] Duration/queue length getter and logic
- [ ] Blob cursor

nevyn:
@bjornbytes Okay, all review comments addressed! Retested, seems to work fine.

@mcclure does this getDuration() API work for you as well as getQueueLength() did? duration on a raw stream now returns amount of material queued measured in seconds, which replaces getQueueLength which was in samples. Multiplying duration by sample rate should get you roughly the same thing.

nevyn:
urrrgh I just realized I think some part of the code is interpreting a mono audiosource as stereo... that would explain some weird buffer sizes, and why the pitch seems wrong....... will investigate tomorrow

src/modules/audio/source.c
```diff
@@ -165,6 +165,12 @@ void lovrSourcePlay(Source* source) {
     return;
   }
 
+  // in case we have some queued buffers, make sure to unqueue them before streaming more data into them.
+  ALint processed;
+  alGetSourcei(lovrSourceGetId(source), AL_BUFFERS_PROCESSED, &processed);
+  ALuint buffers[SOURCE_BUFFERS];
+  alSourceUnqueueBuffers(source->id, processed, buffers);
+
   lovrSourceStream(source, source->buffers, SOURCE_BUFFERS);
```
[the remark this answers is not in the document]
nevyn: We can end up in play() AFTER all buffers are exhaused and processed, but BEFORE AudioUpdate has run. In this case, the OpenAL source is STOPPED (so the guard at start of play() lets us through), but some buffers are still processed and queued and need to unqueued.

This means that after we unqueue the last processed buffers, ALL buffersd are now unqueued and ready for reuse. So we can now use source->buffers and don't have to care about the set of unqueued buffers.

So this code is correct. I'm 99.9% sure :) and it was likely broken even before my changes for code that very aggressively tries to play() when source becomes STOPPED.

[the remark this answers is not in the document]
bjornbytes: Ok, thanks for the explanation :) a small nit would then be to use NULL for the buffers array (like we do in lovrSourceStop) so future people don't mistakenly assume it was a typo.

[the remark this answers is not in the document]
nevyn: so to clarify: the intent of both the original and this modified code is indeed that play() can't be called unless source was stopped, implicitly (by running out of audio to play) or explicitly! but stopped and processed doesn't mean unqueued; this is taken care of both by audio_update and this code, because there is no guarantee that update is called before someone calls play() on it..

also, waiting for audioupdate to do the queueing is bad for latency :)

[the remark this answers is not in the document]
bjornbytes: Ok, thanks for the explanation :) a small nit would then be to use NULL for the buffers array (like we do in lovrSourceStop) so future people don't mistakenly assume it was a typo.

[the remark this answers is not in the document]
nevyn: fixed and documented!

nevyn:
the problem isn't stereo vs mono -- somehow my audio is playing at 1/6th of the correct pitch?? butt still the correct speed????

nevyn:
OH MY GOD I'M STUPID!! The problem isn't in lovr -- it's that my PCM generation test code is wrong after all!! Here's the riddle. Why would this code generate audio that is at 318,47 hz instead of 2000 hz?

```
sd:setSample(i, math.sin(f*2000))
```

.... So.... what do you need to divide 2000 by to get 318.... it's related to basic trigonometry.....
...
...

yes that's right. math.pi*2.

OKAY @bjornbytes I think all the bugs are fixed after all!! go ahead and merge if you want :P

src/modules/audio/source.c
```diff
@@ -274,15 +283,15 @@ void lovrSourceStop(Source* source) {
...
     case SOURCE_STREAM: {
 
+      // Stop the source
+      alSourceStop(source->id);
+      alSourcei(source->id, AL_BUFFER, AL_NONE);
+
       // Empty the buffers
       int count = 0;
       alGetSourcei(source->id, AL_BUFFERS_QUEUED, &count);
       alSourceUnqueueBuffers(source->id, count, NULL);
```
bjornbytes: I am confused how this line is working if the other one in `source.c` didn't work.  Oh well, it can be a future investigation, I really want to get this merged!

PR #187, opened 2020-01-06
Bump OpenAL to 1.20

mcclure:
1.20 is the most recent official release. Upgrading means "record movie" now has sound on Oculus Mobile.


bjornbytes:
Given that there were problems on old OpenAL/miniaudio, and multiple Oculus bugreports for audio capture, I'm going to hold off on this upgrade for now and hope that it gets fixed on the Oculus end so we can switch to miniaudio instead of maintaining OpenAL.  But if it goes on too long or starts to cause more issues, this solution can be revisited.

PR #189, opened 2020-01-14
Channel ownership stuff

bjornbytes:
Long discussions/debugging related to channel ownership and thread safety, ultimately with a revert to the original implementation? >_>

### DIFF
--- a/src/api/l_thread.c
+++ b/src/api/l_thread.c
@@ -86,7 +86,6 @@ static int l_lovrThreadGetChannel(lua_State* L) {
   const char* name = luaL_checkstring(L, 1);
   Channel* channel = lovrThreadGetChannel(name);
   luax_pushtype(L, Channel, channel);
-  lovrRelease(Channel, channel);
   return 1;
 }
 
--- a/src/api/l_thread.c
+++ b/src/api/l_thread.c
@@ -86,9 +86,7 @@ static int l_lovrThreadGetChannel(lua_State* L) {
   const char* name = luaL_checkstring(L, 1);
   Channel* channel = lovrThreadGetChannel(name);
   luax_pushtype(L, Channel, channel);
-  // Note: Channels are intentionally not released here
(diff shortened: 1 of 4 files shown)


src/modules/thread/thread.c
```diff
@@ -8,51 +8,50 @@
...
-  if (index == MAP_NIL) {
-    return;
+  if (entry.u64 == MAP_NIL) {
+    entry.channel = lovrChannelCreate(hash);
+    map_set(&state.channels, hash, entry.u64);
   }
 
-  map_remove(&state.channelMap, hash);
-  arr_splice(&state.channels, index, 1);
+  mtx_unlock(&state.channelLock);
```
mcclure: Should the channel be lovrRetained before releasing the lock?

src/api/l_thread.c
```diff
@@ -86,7 +86,9 @@ static int l_lovrThreadGetChannel(lua_State* L) {
   const char* name = luaL_checkstring(L, 1);
   Channel* channel = lovrThreadGetChannel(name);
   luax_pushtype(L, Channel, channel);
-  lovrRelease(Channel, channel);
+  // Note: Channels are intentionally not released here
+  // Because the ownership of Channels is different from most objects, it is simpler to clean them
+  // all up in lovrThreadModuleDestroy.
```
mcclure: If channels have different ownership rules from other objects it would be helpful if there were a comment explaining this somewhere. I have C code where I call `lovrThreadGetChannel` directly and later (when I'm done calling it) currently release it. I can follow whatever rules I just need to know what they are.

mcclure:
As far as I can see this looks good. I like that it is simpler than the previous implementation. I will do additional testing at my end once this is merged.

src/modules/thread/thread.c
bjornbytes: That's an interesting idea -- it would allow people using this function to perform the "usual" lovrRelease behavior when they're done with the Channel, instead of needing to remember NOT to.

PR #190, opened 2020-01-15
Android Canvas

bjornbytes:
This is a change that shifts the responsibility of the creation
of OpenGL framebuffers from vrapi-provided swapchain texture handles.
Previously, the LovrApp component of lovr-oculus-mobile was creating
framebuffers and passing native framebuffer IDs to lovr.  With this
change, lovr-oculus-mobile passes vrapi's swapchain textures to lovr
unmodified.  This allows lovr to create canvases using its conventional
method and also means that the properties of the canvases are no longer
hardcoded, so things like resolution and multisampling can be
customized.

There were also some issues with multiview canvases in LÖVR due to some
misconceptions about how multisampled multiview rendering works.  These
issues have also been fixed in this PR as well.

### DIFF
--- a/src/modules/graphics/opengl.c
+++ b/src/modules/graphics/opengl.c
@@ -666,11 +666,6 @@ static void lovrGpuBindCanvas(Canvas* canvas, bool willDraw) {
   }
 #endif
 
-  // Use the read framebuffer as a binding point to bind resolve textures
-  if (canvas->flags.msaa) {
-    glBindFramebuffer(GL_READ_FRAMEBUFFER, canvas->resolveBuffer);
-  }
-
   GLenum buffers[MAX_CANVAS_ATTACHMENTS] = { GL_NONE };
   for (uint32_t i = 0; i < canvas->attachmentCount; i++) {
     GLenum drawBuffer = buffers[i] = GL_COLOR_ATTACHMENT0 + i;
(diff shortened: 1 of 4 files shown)


src/modules/graphics/opengl.c
```diff
@@ -1484,15 +1480,15 @@ void lovrTextureAllocate(Texture* texture, uint32_t width, uint32_t height, uint
 
   GLenum glFormat = convertTextureFormat(format);
   GLenum internalFormat = convertTextureFormatInternal(format, texture->srgb);
-#ifndef LOVR_WEBGL
+#ifdef LOVR_GL
```
mcclure: Just curious, why this change?

mcclure:
This change is very exciting, I do not think this code ever belonged in NativeActivity really.

A small downside is if Oculus makes changes to the CubeWorld sample we will no longer automatically pick up those changes. But that is OK, I will just keep an eye out for that.

src/modules/graphics/opengl.c
bjornbytes: `glTexStorage` is core in GLES3 and WebGL, but not in desktop GL.  We only need to see if texStorage exists on desktop GL, this ifdef must have been written when only GL/WebGL were supported.

mcclure: OK. So _GL is explicitly distinct from GLES

Issue #177, opened 2020-01-01
Proposal: lovr.log


bjornbytes:
This is a proposal to add a logging system to LÖVR.

TLDR;

- Add `lovr.log` callback, it's like `print`.
- C code will call `lovr.log` to log things.
- Lua code can also call `lovr.log` to log things.
- You can override `lovr.log` to render logs in VR or add other logic.
- There is a convention for different log levels.

Goals:

- Expose important information from internal systems, like OpenGL/Vulkan/OpenXR debug callbacks, performance warnings, or deprecation notices.
- Add the ability to intercept logs to render them in VR, filter them, or redirect them.
- Establish a convention for libraries to log things, at different log levels.
- Create a point of indirection that allows Android to work properly without "hacks" (normal printing does not work on Android and we monkeypatch `print` to work right now).

Plan:

- Add a `lovr.log` callback.  The default implementation is in C and prints to stdout on desktop and does android logging on Android.  It accepts multiple arguments, like `print`.
- You can override the callback to do whatever you want, can also save the original implementation into a variable and use it.  `print = lovr.log` might be common.
- C code can call a `lovrLog` function, which will redirect to `lovr.log`.
- Log levels are by convention: Anything can add an allcaps log level as its first argument in square brackets, like this: `lovr.log('[WARN]', 'message')`.  There is nothing special about this, but it could be used by the callback to filter or color the log messages.
  - I think this is better than requiring people to always specify the log level or having separate functions for each level.  But I'm still not sure if it's good.
  - Projects can come up with their own additional conventions (timestamps, log tags)?

Questions:

- Should we just use and override `print` for all this?
- New versions of Lua have a `warn` function in addition to `print`.  We still good?
- Is the log level convention a recipe for disaster, or is it good?
- Should there be a `t.log.level` conf.lua flag that adds basic filtering to the default impl?

bjornbytes:
Had another idea for log levels: Instead of `[INFO]`, do `@info` or `@level=info`.  This is the convention used by Lua's new `warn` function, and it kinda decouples the log metadata from how it is presented.

bjornbytes:
This has mostly been implemented on the `log` branch.

One annoying part is that if anything fires a log message before you're able to install your logging handler, it won't get to your callback.  Overriding `lovr.log` in `conf.lua` helps a lot though.

crertel:
Any chance there could be an option for line-delimited JSON output?

bjornbytes:
[...]

PR #191, opened 2020-01-16
lovr.log

bjornbytes:
#177

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -320,6 +320,7 @@ set(LOVR_SRC
   src/main.c
   src/core/arr.c
   src/core/fs.c
+  src/core/log.c
   src/core/maf.c
   src/core/map.c
   src/core/ref.c
--- a/Tupfile
+++ b/Tupfile
@@ -4,6 +4,7 @@ include_rules
 SRC += src/main.c
 SRC += src/core/arr.c
 SRC += src/core/fs.c
+SRC += src/core/log.c
 SRC += src/core/maf.c
 SRC += src/core/map.c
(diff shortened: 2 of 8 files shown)


mcclure:
I'm definitely happy to have log back in the C code. I have been manually adding lovrLog to basically all of my branches so I can printf debug.

Questions

1. Can `print` be a forward to one of the log levels?
2. In the long run should it be possible to specify which log levels to use? Like, maybe it would be cool if you could say "display logs at LOG_INFO level or above" with a command line option or in `lovr.conf`.

bjornbytes:
1. I was planning on not overriding the default `print` function anymore.  There are a few other places where we kinda augment built-in Lua behaviors (math.random, require) and we don't do overrides there.  Projects would then need to do `print = lovr.log` or just not use `print`.  Although this approach feels like the "right" thing to do, I am still thinking about it because the convenience of automatically overriding print is pretty big.
2. It's a possibility.  It seems easy enough to override `lovr.log` to add whatever kind of filtering you want, but yeah the log level filtering is probably the most common one that would make sense to have a built in flag for.

mcclure:
`Projects would then need to do print = lovr.log or just not use print` I cannot express how much I oppose this idea. "print" is Lua. People will just haul off and use "print", besides which libraries probably use "print". If it's there people will use it and then they'll be sitting there trying to figure out why it works on PC but doesn't work on the oculus target

bjornbytes:
I gotta think more about this before continuing, I actually really like the simplicity of Lua's new `warn` function compared to multiple log levels, but it would need to be polyfilled for older Lua versions and it doesn't provide an immediate solution for non-warn prints.

PR #192, opened 2020-01-18
Oculus Mobile hand tracking [prototype]

mcclure:
Values are passed in via some new lists in bridge.h and are returned via some very simplistic functions in a new table `lovr.headset.hand`.

This is the api used with [andi-hand](https://github.com/mcclure/lovr/tree/andi-hand) branch git:67f8a55063ae and is meant to be compiled paired with [lovr-oculus-mobile](https://github.com/mcclure/lovr-oculus-mobile/tree/lovr-oculus-mobile-hand/) git:d10d1e5199ff . The andi-hand branch is separate because that one contains the "finger guns"/"shoot the can" demo.

I think the bridge changes can possibly go in right now, but maybe we shouldn't merge this until we have a better API than the `lovr.headset.hand` table. More comments on this below.

### DIFF
--- a/src/api/l_headset.c
+++ b/src/api/l_headset.c
@@ -534,9 +534,18 @@ static const luaL_Reg lovrHeadset[] = {
   { NULL, NULL }
 };
 
+static void tableRegister(lua_State* L, const char *name, const luaL_Reg *fnTable) {
+  lua_pushstring(L, name);
+  lua_newtable(L);
+  luaL_register(L, NULL, fnTable);
+  lua_settable(L, -3);
+}
+static const luaL_Reg lovrHeadsetHands[];
+
 int luaopen_lovr_headset(lua_State* L) {
   lua_newtable(L);
   luaL_register(L, NULL, lovrHeadset);
+  tableRegister(L, "hand", lovrHeadsetHands);
 
   luax_pushconf(L);
   lua_getfield(L, -1, "headset");
@@ -578,3 +587,93 @@ int luaopen_lovr_headset(lua_State* L) {
   headsetRenderData.ref = LUA_NOREF;
   return 1;
 }
+
+// TEMPORARY HANDS LIBRARY
+
+#include "headset/oculus_mobile_bridge.h"
+#include "headset/oculus_mobile.h"
+
+static void pushTableOfNumbers(lua_State* L, int idx, int count, ...) { // Takes floats
+  va_list args;
+  va_start(args, count);
+
+  lua_newtable(L);
+  for(int c = 0; c < count; c++) {
+    // Note we pass floats but they become doubles because of some garbage about how va_list works.
+    // This is probably okay becuase lua_number is probably a double anyway.
+    lua_pushnumber(L, va_arg(args, double));
+    lua_rawseti (L, -2, c+1);
+  }
+  lua_rawseti (L, -2, idx);
+
+  va_end(args);
+}
+
+static int l_lovrHeadsetHandsIsHand(lua_State* L) {
+  Device device = luax_optdevice(L, 1);
+  lovrAssert(device == DEVICE_HAND_LEFT || device == DEVICE_HAND_RIGHT, "Only works with hands");
+  LovrOculusMobileHands *hand = &lovrOculusMobileHands[device == DEVICE_HAND_RIGHT];
+
+  int points = 0;
+  lua_pushboolean(L, hand->live);
+
+  return 1;
+}
+
+static int l_lovrHeadsetHandsGetPoints(lua_State* L) {
+  Device device = luax_optdevice(L, 1);
+  lovrAssert(device == DEVICE_HAND_LEFT || device == DEVICE_HAND_RIGHT, "Only works with hands");
+  LovrOculusMobileHands *hand = &lovrOculusMobileHands[device == DEVICE_HAND_RIGHT];
+
+  int points = 0;
+  lua_newtable(L);
+  for(int c = 0; c < hand->handPoses.members; c++) {
+    BridgeLovrPose *pose = &hand->handPoses.poses[c];
+    lua_newtable(L);
+    pushTableOfNumbers(L, 1, 3, pose->x, pose->y, pose->z);
+    pushTableOfNumbers(L, 2, 4, pose->q[0], pose->q[1], pose->q[2], pose->q[3]);
+    lua_rawseti (L, -2, ++points);
+  }
+
+  return 1;
+}
+
+static int l_lovrHeadsetHandsGetBoneNames(lua_State* L) {
+  Device device = luax_optdevice(L, 1);
+  lovrAssert(device == DEVICE_HAND_LEFT || device == DEVICE_HAND_RIGHT, "Only works with hands");
+  LovrOculusMobileHands *hand = &lovrOculusMobileHands[device == DEVICE_HAND_RIGHT];
+
+  lua_newtable(L);
+
+  for(int c = 0; c < hand->bones.members; c++) {
+    lua_pushstring(L, hand->bones.strings[c]);
+    lua_rawseti (L, -2, c+1);
+  }
+
+  return 1;
+}
+
+static int l_lovrHeadsetHandsGetConfidence(lua_State* L) {
+  Device device = luax_optdevice(L, 1);
+  lovrAssert(device == DEVICE_HAND_LEFT || device == DEVICE_HAND_RIGHT, "Only works with hands");
+  LovrOculusMobileHands *hand = &lovrOculusMobileHands[device == DEVICE_HAND_RIGHT];
+
+  lua_newtable(L);
+
+  lua_pushnumber(L, hand->confidence);
+  lua_rawseti (L, -2, 0);
+  for(int c = 0; c < hand->fingerConfidence.members; c++) {
+    lua_pushnumber(L, hand->fingerConfidence.numbers[c]);
+    lua_rawseti (L, -2, c+1);
+  }
+
+  return 1;
+}
+
+static const luaL_Reg lovrHeadsetHands[] = {
+  { "isHand", l_lovrHeadsetHandsIsHand },
+  { "getPoints", l_lovrHeadsetHandsGetPoints },
+  { "getBoneNames", l_lovrHeadsetHandsGetBoneNames },
+  { "getConfidence", l_lovrHeadsetHandsGetConfidence },
+  { NULL, NULL }
+};
--- a/src/modules/headset/oculus_mobile.c
+++ b/src/modules/headset/oculus_mobile.c
@@ -186,7 +186,11 @@ static bool vrapi_isDown(Device device, DeviceButton button, bool* down) {
   if (idx < 0)
     return false;
 
-  return buttonDown(bridgeLovrMobileData.updateData.controllers[idx].buttonDown, button, down);
+  BridgeLovrController *data = &bridgeLovrMobileData.updateData.controllers[idx];
+  if (data->hand & BRIDGE_LOVR_HAND_TRACKING)
+    return false;
+
+  return buttonDown(data->handset.buttonDown, button, down);
 }
 
 static bool vrapi_isTouched(Device device, DeviceButton button, bool* touched) {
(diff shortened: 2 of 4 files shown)


src/api/l_headset.c
```diff
@@ -578,3 +587,93 @@ int luaopen_lovr_headset(lua_State* L) {
...
+    // This is probably okay becuase lua_number is probably a double anyway.
+    lua_pushnumber(L, va_arg(args, double));
+    lua_rawseti (L, -2, c+1);
+  }
+  lua_rawseti (L, -2, idx);
+
+  va_end(args);
+}
+
+static int l_lovrHeadsetHandsIsHand(lua_State* L) {
```
mcclure: True if the hand has hand tracking

src/api/l_headset.c
```diff
@@ -578,3 +587,93 @@ int luaopen_lovr_headset(lua_State* L) {
...
+  lovrAssert(device == DEVICE_HAND_LEFT || device == DEVICE_HAND_RIGHT, "Only works with hands");
+  LovrOculusMobileHands *hand = &lovrOculusMobileHands[device == DEVICE_HAND_RIGHT];
+
+  int points = 0;
+  lua_pushboolean(L, hand->live);
+
+  return 1;
+}
+
+static int l_lovrHeadsetHandsGetPoints(lua_State* L) {
```
mcclure: This function returns ALL joint poses. Each joint is a table containing {{x,y,z}, {a,b,c,d}} where abcd are the quaternion components (so, it's not angle axis like Lovr poses normally are). The order of the joint poses is hardcoded and is the order VrApi lists them in. In my personal code, I make list of this ordering with these two slightly awkward tables.

```
HandCubes.skeletonSkeleton = {
  [1]=2, -- Forearm
  [3]=1, [4]=3, [5]=4, [6]=5, [20]=6, -- Thumb
  [7]=1, [8]=7, [9]=8, [21]=9,        -- Index
  [10]=1, [11]=10, [12]=11, [22]=12,  -- Middle
  [13]=1, [14]=13, [15]=14, [23]=15,  -- Ring
  [16]=1, [17]=16, [18]=17, [19]=18, [24]=19 -- Pinky
}
HandCubes.skeletonTipStart = 20
HandCubes.expectedLength = 24
HandCubes.skeletonFingers = {
	{4, 5, 6, 20}, -- Thumb
	{7, 8, 9, 21},    -- Index
	{10, 11, 12, 22}, -- Middle
	{13, 14, 15, 23}, -- Ring
	{17, 18, 19, 24}, -- Pinky
}
```

Points are returned in "hand space". To get global space you need to compose the normal `lovr.headset.getPose` pose. In my personal code I get the list of global-space points with:

```
		local points = tablex.map(
			function(point)
				return Loc( vec3(unpack(point[1])), quat(unpack(point[2])) ):compose(handAt)
			end,
			lovr.headset.hand.getPoints(controllerName)
		)
```

There was some kind of problem I hit early on involving the left hand being rotated 180 degrees all the time but I don't remember where I resolved it.

src/api/l_headset.c
```diff
@@ -578,3 +587,93 @@ int luaopen_lovr_headset(lua_State* L) {
...
+    lua_newtable(L);
+    pushTableOfNumbers(L, 1, 3, pose->x, pose->y, pose->z);
+    pushTableOfNumbers(L, 2, 4, pose->q[0], pose->q[1], pose->q[2], pose->q[3]);
+    lua_rawseti (L, -2, ++points);
+  }
+
+  return 1;
+}
+
+static int l_lovrHeadsetHandsGetBoneNames(lua_State* L) {
```
mcclure: This returns a list of strings, each one being the skeleton node name for the joint at that index in the GetPoints list.

src/api/l_headset.c
```diff
@@ -578,3 +587,93 @@ int luaopen_lovr_headset(lua_State* L) {
...
+
+  for(int c = 0; c < hand->bones.members; c++) {
+    lua_pushstring(L, hand->bones.strings[c]);
+    lua_rawseti (L, -2, c+1);
+  }
+
+  return 1;
+}
+
+static int l_lovrHeadsetHandsGetConfidence(lua_State* L) {
```
mcclure: This returns a table where indices 1-5 contain the confidence values for the 5 fingers, and index 0 contains the confidence value for the entire hand. A confidence value is a float between 0 and 1. In the current implementation it is always EXACTLY either 0 or 1 because the current vrapi only gives us an enum containing "low value" or "high value". I have suspicions that might change but who knows.

src/modules/headset/oculus_mobile.c
```diff
@@ -456,6 +466,17 @@ void bridgeLovrInit(BridgeLovrInitData *initData) {
...
+  { \
+      size_t arraySize = IN->members * sizeof(TYPE); \
+      if (IN->members > OUT.members) { \
+        free(OUT.DATA); \
+        OUT.DATA = malloc(arraySize); \
+      } \
+      OUT.members = IN->members; \
+      memcpy(OUT.DATA, IN->DATA, arraySize); \
+  }
+
```
mcclure: This is ugly and I hate it. I'm not sure I have a better way of doing this in C. The problem is the bridge has to convey 3 lists, one a list of strings (which never changes for a particular frame), one a list of poses (which changes every frame), one a list of confidence values (which changes every frame). I'm not sure what the best way to shuttle that data around in C is so I wound up creating 3 "list" structs, another thing I don't like but couldn't think of an obvious better option. Maybe I should be using arr_t here, I don't know. 

The "bridge" mechanism is definitely turning out to be a bigger inconvenience than I thought when I originally came up with it.

src/modules/headset/oculus_mobile_bridge.h
```diff
@@ -75,10 +75,16 @@ typedef enum
   BRIDGE_LOVR_TOUCH_TRIGGER_ANTI  = 0x00000200, // "The finger is sufficiently far away from the trigger to not be considered in proximity to it."
 } BridgeLovrTouch;
 
-// Bit identical with VrApi_Input.h ovrControllerCapabilties
```
mcclure: This enum used to be a copy of ovrControllerCapabilties, now it is not.

mcclure:
@bjornbytes ' thoughts on api design, from Slack the other night:

> I guess just because I'm thinking about it now, here are my notes on skeleton APIs so far
> - lovr.headset.newModel(device, [modelType]) new modelType parameter can be "controller" or "hand".  It probably defaults to "controller" to preserve existing behavior.
> - lovr.headset.getSkeleton(device, modelType, [dest]) returns an array of current bone transforms for the device, each in parent-local space.  You can get bones for either the "hand" or the "controller" (for controller models that move based on input state).  If you pass in dest, it can be a table to write to, or a Model in which case we just write the bone transforms directly to that Model's pose.
> - Add 10 new devices, one for each finger of each hand.  They can implement getPose, getVelocity, and getAxis (1D curl and 2D splay)
> Optional:
> - getSkeleton accepts different coordinate spaces like "relative to hand" and "relative to play area"
> - Add separate pinch axis instead of mapping it to trigger, maybe similar thing for grab. (edited)

mcclure:
My thoughts on api design:
- I think in the end the "here's a list of joint poses" approach is probably the best one. The user could make sense of the list of poses by using the "pose names" (maybe we could standardize these) and maybe there could be a "getFingers" or "getSkeletonConnections" function that returns some equivalent of my `HandCubes.skeletonSkeleton` and `HandCubes.skeletonFingers` arrays above.
- I don't think it makes sense to address fingers as devices. For one thing fingers don't have poses, only joints do. For another, there are multiple joints in the hands that logically aren't part of any finger :(
- I think we want to ultimately use the same API for oculus mobile and the Valve Index (we could even come up with a fake hand for the Oculus Touch, like Superhot VR does?). It is not clear to me every hand api will have the same number of joints exactly.
- I like bjorn's idea of getSkeleton or getPoses or whatever taking an enum to say if it's in local or global space, and I like the idea of offering a "pinch" axis (although one problem: pinching what? on oculus mobile "pinch" means pinching the thumb and index finger; should we allow pinching with the middle finger?)
- The WebXR people are trying to solve the exact same problem we are (write a single api that is a cover of both what oculus mobile and what the Valve Index offers) so maybe we could just copy their approach, whatever that turns out to be
- I would like it if the getSkeleton or getPoses or whatever and getBoneNames somehow works for *handsets also*, because handset models also can have bones (ie for buttons being pressed or triggers being pushed). In my personal project I have a bunch of code that sets the skeleton nodes on the handset model depending on what the buttons are doing. I think some variant of this code should be taken up into lovr. One way this could work is like

        local model = lovr.headset.newModel("hand/left", "hand")
        model:setNodePoses(lovr.headset.getSkeletonNames("hand/left"), 
        lovr.headset.getSkeleton("hand/left"))

    If we select standard node names for the models and make them the same as the ones getSkeleton/getSkeletonNames return, then this same approach could work with both the models we return automatically from headset.newModel and also custom models the user provides.

* One last super abstract thought: I think our approach should *leave room* for the idea of supporting special hand models / notation for users who have a hand with fewer than five fingers. I don't know what this would look like and the API doesn't give us this data currently so we can't do this now anyway, but I am pushing both Oculus and WebXR to consider this.

bjornbytes:
> fingers don't have poses, only joints do

I am confused about this.  fingertip poses are something that are usually very important for gameplay code, and all the hand tracking SDKs I've seen (leap, openvr, vrapi) go through the trouble of adding extra fake joints at the ends of fingers just so the poses of the finger tips are queryable.

mcclure:
> fingertip poses

Oh! Yeah. So I guess per-finger things we **can** get are:

- Tip pose
- Curl amount
- Pinch amount (I don't know if we want pinch amount to be per-finger or per-hand)

It might be if you have those things you don't need to look at the individual joints.

src/api/l_headset.c
bjornbytes: I think this should be integrated into `isTracked` and/or the return value for `getSkeleton`.

mcclure: Merging this with lovr.headset.isTracked could make sense, like maybe it could return an enum. Should it still be called isTracked at that point?

Should whatever enum we go with be able to distinguish knuckle / oculus touch-level approximate tracking from fully articulated hand tracking?

bjornbytes: I was just thinking `lovr.headset.isTracked(device, 'skeleton')` since there was talk about doing `lovr.headset.isTracked(device, 'position')` in the past.  It's true that it doesn't expose the "estimated vs. accurate" tracking level.

mcclure: oh, `lovr.headset.isTracked(device, 'skeleton')` is good. I like how that could support several levels of complexity but still degrades to the existing API…

bjornbytes:
I started working on the OpenVR / Leap Motion side of things, mostly copy/paste from old branches https://github.com/bjornbytes/lovr/pull/197

src/api/l_headset.c
bjornbytes: Parent-space joint poses might make more sense for the bridge interface.  Composing the parent space transforms into higher spaces is more straightforward than going backwards from hand/global transforms (e.g. matrix multiplications vs. matrix inversions).  Also `Model:pose` takes poses in parent space, so using parent space would mean that the poses could be directly copied to a model without conversions.

It might be less overall work too, since the [OVR_Skeleton code](https://github.com/mcclure/lovr-oculus-mobile/blob/lovr-oculus-mobile-hand/VrSamples/VrHands/Src/OVR_Skeleton.cpp#L62) used to generate this is already seemingly taking the parent space poses from vrapi and converting them to hand space for the bridge.  It seems wasteful to take parent-space poses, unconditionally convert them to hand space, then invert them back to parent space.  Instead they could remain in parent space and only get converted to hand/global space if that's what the Lua code wants to do.

bjornbytes:
One interesting thing to think about is the basis for the hands, like whether the natural z axis for a hand skeleton in LÖVR goes out of the palm, along the fingers, etc.  As far as I can tell there is no agreement on what this should be across the various APIs, so some conversions will need to take place.

src/modules/headset/oculus_mobile.c
bjornbytes: Could it be a fixed-size array?  The number of bones is unlikely to change very often (I haven't seen a hand tracking SDK with more than 32, so far), and maybe the tradeoff of "may need to increase size limit later" is worth it compared to "need to deal with dynamic sizes/allocations".

src/api/l_headset.c
mcclure: This makes a lot of sense to me.

src/modules/headset/oculus_mobile.c
mcclure: That would simplify things, but the problem with a fixed-size array is I think it would mean (because of the union) that we'd be paying to copy space for the fixed-size array even when using controllers (the typical case).

bjornbytes:
Some new developments:

- There's a new native `vrapi` headset backend, hand tracking should be implemented there now.  Hopefully it's a bit easier with no need to send the dynamic arrays over the bridge.
- OpenXR has published an EXT-level hand tracking extension that exposes joint poses, radii, and velocities, all under a standard skeleton model.  Mapping vrapi's skeleton data onto OpenXR's skeleton would be great because it would make it possible to write cross-device hand tracking stuff.  The difficulty of this is still unknown (just a few rotations, what could go wrong).

I'm going to close this since there would be significant changes required.

PR #193, opened 2020-01-19
Sound adjustments: SoundData has-a blob, and mic sample size getter

nevyn:
Just some small fixes that ease the handling of audio:

* `SoundData` gets a `Blob*` ivar instead of `Blob`. This means the blob can be accessed as a proper child object.
* In that vein, `SoundData` gets `SoundData:getBlob()`. This way, you can `getString` etc on the blob in the `SoundData`, and use the data with other APIs more easily
* Slightly related, `Microphone:getData()` gets a buffer size parameter. I was originally going to implement a ring buffer class, but I only need it to buffer microphone data before it's ready to be sent off to the network. Turns out OpenAL already has a ringbuffer internally, and by adding a parameter to say how much data you want to chop of NOW, we can utilize that ringbuffer downstream. This makes my code much easier.

The last one there is going to make it much easier to use `SoundDataPool` from the microphone class, once I implement it. That's my next step, to reduce the `SoundData` malloc churn.

### DIFF
--- a/src/api/l_audio_microphone.c
+++ b/src/api/l_audio_microphone.c
@@ -18,7 +18,8 @@ static int l_lovrMicrophoneGetChannelCount(lua_State* L) {
 
 static int l_lovrMicrophoneGetData(lua_State* L) {
   Microphone* microphone = luax_checktype(L, 1, Microphone);
-  SoundData* soundData = lovrMicrophoneGetData(microphone);
+  size_t samples = luax_optfloat(L, 2, 0);
+  SoundData* soundData = lovrMicrophoneGetData(microphone, samples);
   luax_pushtype(L, SoundData, soundData);
   lovrRelease(SoundData, soundData);
   return 1;
--- a/src/api/l_audio_microphone.c
+++ b/src/api/l_audio_microphone.c
@@ -18,7 +18,7 @@ static int l_lovrMicrophoneGetChannelCount(lua_State* L) {
 
 static int l_lovrMicrophoneGetData(lua_State* L) {
   Microphone* microphone = luax_checktype(L, 1, Microphone);
-  size_t samples = luax_optfloat(L, 2, 0);
+  size_t samples = luaL_optinteger(L, 2, 0);
   SoundData* soundData = lovrMicrophoneGetData(microphone, samples);
   luax_pushtype(L, SoundData, soundData);
   lovrRelease(SoundData, soundData);
--- a/src/api/l_data_soundData.c
+++ b/src/api/l_data_soundData.c
@@ -51,7 +51,6 @@ static int l_lovrSoundDataGetBlob(lua_State* L) {
   SoundData* soundData = luax_checktype(L, 1, SoundData);
   Blob* blob = soundData->blob;
   luax_pushtype(L, Blob, blob);
-  lovrRelease(Blob, blob);
   return 1;
 }
 
--- a/src/api/l_data_soundData.c
+++ b/src/api/l_data_soundData.c
@@ -47,12 +47,6 @@ static int l_lovrSoundDataSetSample(lua_State* L) {
   return 0;
 }
 
-static int l_lovrSoundDataGetPointer(lua_State* L) {
-  SoundData* soundData = luax_checktype(L, 1, SoundData);
-  lua_pushlightuserdata(L, soundData->blob->data);
-  return 1;
-}
-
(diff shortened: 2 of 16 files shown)


bjornbytes:
Initial thoughts:

- I'd like to apply this treatment to both `TextureData` and `SoundData` for consistency between the two objects, since they are both "blobs + metadata" objects.
- The `:getPointer` function on both of these objects can be removed.

nevyn:
@bjornbytes agree on both points! I can go ahead and do that now.

nevyn:
@bjornbytes this is now done. I'm not at all as familiar with texture data, so I don't know if I messed something vital up.

Removing *Data:getPointer() will also break backwards compatibility, but for a 0.x api, it makes sense to optimize for a pretty API surface imo, so this is nicer.

src/api/l_audio_microphone.c
```diff
@@ -18,7 +18,8 @@ static int l_lovrMicrophoneGetChannelCount(lua_State* L) {
 
 static int l_lovrMicrophoneGetData(lua_State* L) {
   Microphone* microphone = luax_checktype(L, 1, Microphone);
-  SoundData* soundData = lovrMicrophoneGetData(microphone);
+  size_t samples = luax_optfloat(L, 2, 0);
```
bjornbytes: It seems a little bit odd to use `luax_optfloat` to read a `size_t`.

src/api/l_data_soundData.c
```diff
@@ -46,9 +47,11 @@ static int l_lovrSoundDataSetSample(lua_State* L) {
   return 0;
 }
 
-static int l_lovrSoundDataGetPointer(lua_State* L) {
+static int l_lovrSoundDataGetBlob(lua_State* L) {
   SoundData* soundData = luax_checktype(L, 1, SoundData);
-  lua_pushlightuserdata(L, soundData->blob.data);
+  Blob* blob = soundData->blob;
+  luax_pushtype(L, Blob, blob);
+  lovrRelease(Blob, blob);
```
bjornbytes: I don't think releasing the Blob is wanted here.  If `getBlob` was called multiple times, eventually it would get freed.  Usually `lovrRelease` is used in the API if we've just created a new object and have no need for C to keep its reference around.

src/modules/audio/microphone.c
```diff
@@ -41,18 +41,22 @@ uint32_t lovrMicrophoneGetChannelCount(Microphone* microphone) {
...
     return NULL;
   }
 
-  size_t samples = lovrMicrophoneGetSampleCount(microphone);
-  if (samples == 0) {
+  size_t maxSamples = lovrMicrophoneGetSampleCount(microphone);
+  if (maxSamples == 0) {
     return NULL;
   }
+  lovrAssert(samples <= maxSamples, "Requested more audio data than is buffered by the microphone");
```
bjornbytes: Thoughts on using `MIN` here to clamp the input sample count to the max and make this function less throwy?  I don't have a strong opinion right now, but it's interesting to think about.

src/api/l_data_soundData.c
nevyn: Ah I just thought I saw it used like that somewhere else; figured maybe luax_pushtype needed to be balanced (that's one of the pieces of magic I haven't dug too deep into yet). Okay, I'll just remove it from sounddata and texturedata!

src/api/l_audio_microphone.c
nevyn: I don't know what I was thinking... fixed

src/modules/audio/microphone.c
nevyn: It is, isn't it? I can't make up my mind. It's nice that the code is less error prone, but it's un-nice that it ends up doing something else than what was requested. If you lean slightly towards the latter being the lesser evil, I'll go with it.

bjornbytes: Yeah I think that's a good read.  I think it's the kind of thing that you just feel out while using the API for a while, so let's keep it in mind :)

nevyn:
It seems I had forgotten to stage some code last night :S Now pushed that code + the requested fixes.

bjornbytes:
I am just going to do a few style things after merge, don't want to bother you with them though:

- Maybe the core/ref includes might not be needed anymore?
- Spaces after casts
- Put type and star together in pointer types

Thank you for this change!  I think it helps detangle some of the blob type weirdness

PR #194, opened 2020-01-19
WIP SoundDataPool

nevyn:
Pushing this as a PR just to get feedback on the approach. It doesn't work yet, audio is all corrupted, not sure yet what's wrong.

This is based on top of https://github.com/bjornbytes/lovr/pull/193 .

### DIFF
--- a/src/core/ref.h
+++ b/src/core/ref.h
@@ -1,5 +1,6 @@
 #include <stdint.h>
 #include <stddef.h>
+#include <stdbool.h>
 
 #pragma once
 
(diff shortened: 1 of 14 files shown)


src/core/ref.h
```diff
@@ -49,9 +50,21 @@ static inline uint32_t ref_dec(Ref* ref) { return atomic_fetch_sub(ref, 1) - 1;
 
 #endif
 
+typedef bool (*RefDestructor)(void* o, void *extra);
+typedef struct RefHeader {
+  Ref ref;
+  RefDestructor destructor;
+  void* destructorExtra;
+} RefHeader;
```
bjornbytes: This is a pretty big change to the object system, so I just want to make sure that there is a good reason to go from 4 -> 24 bytes.

More background: Refs used to store their destructor/name until I realized that it was a lot of redundant information and the only place we truly needed "polymorphic" information was for Variants.  In 99% of places we already know what destructor to use when releasing the object.

nevyn: Yeah. I saw _lovrRelease but special casing release everywhere that a SoundData might be used felt very error prone to me. Can you think of another approach? I couldn't :S 

Also, it's 8 bytes; size_t was used for malloc even though the payload was only 4 bytes.

mcclure:
Confused… is this just object pooling for SoundData blobs?

Why do this for SoundData specifically instead of making it generalized for blobs?

nevyn:
@mcclure Well, that's a good question... After #193 SoundData isn't a blob, it has a blob. But you're right, the nicer design might be to make only Blob poolable, and then you can easily wrap the vended Blobs in a SoundData/TextureData if you want.

bjornbytes:
In my text editor project I was able to do fixed-size SoundData pooling in Lua by adding an out-arg to `Microphone:getData`:

```lua
local buffers = {}
local chunkSize = 512

function lovr.update(dt)
  while microphone:getSampleCount() > chunkSize do
    local buffer = microphone:getData(chunkSize, table.remove(buffers))
    stream:push(buffer)
  end

  -- stream processes buffers asynchronously,
  -- then returns them back to the pool later...

  while stream:peek() do
    table.insert(buffers, stream:pop())
  end
end
```

It only ever creates a new `SoundData` if there are no more buffers available (implicitly, `table.remove` returns `nil` when there are no more buffers which gets passed as the last arg to `getData` causing it to create a new buffer).

If the buffers don't need to be fixed size or your processing is synchronous it could probably be simplified further.

nevyn:
@bjornbytes nice. soo... merge #193, drop #194 and add your out-arg patch instead? seems a lot simpler than this callback monstrosity I built.

my patch really didn't behave well either... gc is so far in between that waiting for destructor means thousands of buffers can build up in the pool..

bjornbytes:
As long as it works for your use case too, sounds good to me!

nevyn:
I think it will :) Let's close this for now; I can revive the code and give it another iteration if it seems I really need it.

PR #197, opened 2020-01-23
More Hand Stuff

bjornbytes:
WIP finger tracking stuff

**Stuff**

- [x] Add 10 devices, one for each finger of each hand
- [x] Add curl/splay axes
- [x] OpenVR driver implements getPose for finger devices
- [x] OpenVR driver implements curl/splay
- [x] Leap driver implements getPose
- [x] Leap driver implements curl/splay (emulated)
- [x] Add pinch axis
- [x] Leap driver uses pinch axis instead of trigger
- [x] getSkeleton (may do this later, it's way harder at least for leap)

**Thoughts**

- Kinda want shorthand devices for fingers like hands have (`lovr.headset.getPose('left/pinky')`)
- Hardcoding the `WithController` range of motion for OpenVR
- Maybe if we wait long enough we can see what OpenXR does!
- Coordination with #192
- OpenVR is returning weird rotations for fingers and I don't know why


mcclure:
This sounds like a good starting place for work, long term thoughts: 
- as discussed before i don't think the idea of a "device" for a finger makes total sense. I want to suggest the device be "hand/left/finger/thumb/tip" instead of "hand/left/finger/thumb" or something (because it makes sense to request a pose of a fingertip, but not to view the fingertip as a device independent of the hand and it's odd to me to conflate a finger with the fingertip). also in some cases a user might want the pose of EVERY joint and i don't know if we have a method for describing what that would be. maybe another argument should be added to getPose so you could ask for getPose("left/pinky", "tip").
- If you're doing shorthands: I think previously I've talked about the idea of numeric aliases for certain kinds of devices. I don't know how doable this is but it sure would be nice if I could request left/5 as a synonym for left/pinky so I don't have to teach my code the names for each of the 5 fingers if the fingers are all identical from the code's perspective..

PR #202, opened 2020-01-27
Compile shaders using source lengths

### LINKED ISSUE #201: Master branch doesn't null-terminate shader code
The current master branch doesn't null terminate shader code in `luax_checkshadersource` and throws away the length, so OpenGL will non-deterministically run into garbage after the user's shader code.

Discussed on the Slack: this is likely fallout from working on the filesystem API, and it would be good to use pointer + length instead of null-terminating in order to avoid copying. This would also help with compatiblity with binary shader formats like SPIRV bytecode.

porglezomp:
The previous implementation relied on `glShaderSource` inferring source lengths when the lengths weren't specified. This relies on the sources being properly null-terminated, however, which isn't the case due to file loading changes which now use pointer + length. This could cause intermittent crashes.

Changing this on the shader side meant adding some extra arguments for passing around shader source lengths. For most of the other cases, where we're using string literals as the sources, we can just specify -1 as the length, since OpenGL will calculate the string length for you any time the length is negative.

Fixes #201

### DIFF
--- a/src/api/l_graphics.c
+++ b/src/api/l_graphics.c
@@ -1400,23 +1400,26 @@ static int l_lovrGraphicsNewModel(lua_State* L) {
   return 1;
 }
 
-static const char* luax_checkshadersource(lua_State* L, int index) {
+static const char* luax_checkshadersource(lua_State* L, int index, int *outLength) {
   if (lua_isnoneornil(L, index)) {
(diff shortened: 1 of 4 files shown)


bjornbytes:
Thanks for fixing this!

Issue #198, opened 2020-01-26
Expose Headset Views


bjornbytes:
Goals

- Expose the view/projection matrices for headsets.
- Support an arbitrary number of views (e.g. 6-view CAVE, 4-view foveated, 1-view AR).

Details

- A view is a regular pose (3-float position, 4-float orientation) and a field of view (4-float angles).  That's 11 numbers instead of 32 for the view/projection matrices.  Normally there are 2 views.
- Multiview rendering pipeline and Canvas/Camera struct should support N views (at least 6), don't hardcode a "stereo" property.
- Likely just expose view count and per-view pose/fov in the headset module.  It's more of a "display" thing than a "device" thing (it's not like views correspond to eye devices or anything like that).
  - `lovr.headset.getViewCount() --> count`
  - `lovr.headset.getViewPose(index) --> x, y, z, angle, ax, ay, az`
  - `lovr.headset.getViewProjection(index) --> left, right, top, bottom`
  - Naming needs work
- OpenXR does all this really well, it's called "view configurations".
- This is all information that we already have anyway because we need it to render, so it should be relevant for all headset drivers.

Questions

- How do compositor layers/overlays interact with this, if at all?
- Any interactions with foveated rendering?
- Probably want to have some way to pick a view configuration in `conf.lua`.  Can view configurations change over time or is it okay to set it once at startup?  What if the one you want isn't supported?  Should it be a preference list like t.headset.drivers?
- Would anyone actually use any of this

porglezomp:
> * Would anyone actually use any of this

I just ran into this issue because I need to access the FoVs of the camera so I can make a modified projection for rendering effects that extend beyond the screen boundaries, so here's a vote for yes.

bjornbytes:
#206

bjornbytes:
OpenVR and OpenXR and LibOVR expose 4 fov angles, but WebXR exposes a projection matrix and explicitly says that it can include shear so it can't necessarily be decomposed into 4 angles.

https://github.com/immersive-web/webxr/issues/461

Maybe on WebXR `getViewAngles` would return `nil` to indicate that the information can not be reliably exposed.  Or we could switch to exposing a projection matrix.

PR #206, opened 2020-01-28
Headset views;

bjornbytes:
See #198 

- `lovr.headset.getViewCount`
- `lovr.headset.getViewPose`
- `lovr.headset.getViewAngles`

Todo

- [x] Finalize API:
  - Internal API: might be better off as `getViews(views[2], &count)`, for many reasons
  - External API: nil vs. errors, naming

Maybe

- [ ] Write `mat4_getFov` so we can extract fovs from projection matrices (oculus_mobile)
- [ ] Consider exposing `mat4_getFov` as `left, right, up, down = Mat4:fov(nil)`
- [ ] Consider making `mat4:fov` accept a `vec4` for the angles, so you can do `mat4:fov(vec4(lovr.headset.getViewAngles(i)), lovr.headset.getClipDistance())`.
- [ ] Consider exposing `getDisplayTime` and adding a timestamp as the second argument to the accessors, so you can get view/projection predictions for any timestamp.  Probably won't do this, at least right now.
- [ ] Change canvas/shader stereo flag to be a viewCount (but still support stereo and map to viewCount = 2, in a deprecated fashion).  This lets you create a Canvas for a headset render target based on the actual view count!
- [ ] Maybe WebVR but probably later
- [ ] Something else I'm forgetting

### DIFF
--- a/src/api/l_headset.c
+++ b/src/api/l_headset.c
@@ -181,6 +181,44 @@ static int l_lovrHeadsetGetDisplayMask(lua_State* L) {
   return 1;
 }
 
+static int l_lovrHeadsetGetViewCount(lua_State* L) {
+  lua_pushinteger(L, lovrHeadsetDriver->getViewCount());
+  return 1;
+}
+
+static int l_lovrHeadsetGetViewPose(lua_State* L) {
+  float position[4], orientation[4];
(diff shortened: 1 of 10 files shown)


bjornbytes:
I'm going to merge this because it provides some useful functionality in its current state and gets the API in place.  There are likely a few improvements that can be made in the future.

PR #207, opened 2020-01-29
Fix missing install targets in cmake; Add nix derivation.

aiverson:
This PR adds some missing target properties to permit the build to properly package, removing a hidden dependency on binaries in the source directory from the rpath of the generated lovr executable, and populating install destinations for targets that were missing it. This makes it amenable to portable building and packaging.

Additionally, this PR adds the `default.nix` packaging recipe, allowing this repo to be referenced by any nix tool.


bjornbytes:
I don't know much about nix or NixOS, so I'm just curious how adding the nix recipe here is useful.  I saw that nixpkgs contains `default.nix`s for lots of software, is there a difference between adding the recipe there vs. here?  Is it mostly to allow for local package builds?  I've been trying to minimize the amount of clutter in the root of the repository, so I guess there is a bit of a higher bar for adding `default.nix`.

Also thank you for the CMake and rpath fixes!

aiverson:
by putting the default.nix here rather than in nixpkgs it both enables local package builds and makes pinning specific versions of lovr in other projects much easier. It enables getting packages built from specific git revisions by just pointing at the github repo. It also means that doing development on this repo requires only a single nix-build invocation.

bjornbytes:
I'm going to hold off on merging this right now.  Thanks for taking the time to do this, I'm definitely going to keep this patch in mind for the future.

PR #221, opened 2020-02-13
lovr.headset.getDevices (straw proposal, don't merge)

mcclure:
This is something I did in my private branch; I'm PRing it as an example to propose something for the still-incoming isTracked() changes.

Whereas getHands() only returns hand/left and hand/right, getDevices() returns all devices Lovr knows about which could respond to any of getPose, isTracked, isDown or getAxis. This includes the head. It is useful for hypothetical controller support.

I would like to propose a version of this function go into the isTracked() change. I don't know the new isTracked() signature, but this function would have a similar signature. For example if `isTracked("responds","isDown","b")` is yes for an Oculus Touch controller, then `getDevices("responds","isDown","b")` would return all devices that are capable of responding to isDown("b"). This would make it easier to write simple programs that are agnostic to what kinds of devices they communicate with, as long as those devices fit how they will be using the device.

### DIFF
--- a/src/api/l_headset.c
+++ b/src/api/l_headset.c
@@ -502,6 +502,28 @@ static int l_lovrHeadsetGetHands(lua_State* L) {
   return 1;
 }
 
+static int l_lovrHeadsetGetDevices(lua_State* L) {
+  if (lua_istable(L, 1)) {
+    lua_settop(L, 1);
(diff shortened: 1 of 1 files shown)


bjornbytes:
Going to close this since there is discussion going on in #223

PR #223, opened 2020-02-13
lovr.headset.getStatus; lovr.headset.glob;

bjornbytes:
- rm the Device variant of `lovr.headset.getDriver`.  Now it only takes zero arguments and always returns the display driver.
- Rename `lovr.headset.isTracked` to `lovr.headset.getStatus`.
- `lovr.headset.getStatus` takes a Device and an optional TrackingType (`pose`, `velocity`, `button`, `touch`, `axis`, `skeleton`, defaults to pose).
- It returns a TrackingStatus (`unavailable`, `estimated`, `tracked`), and a driver.
- Add `lovr.headset.glob` as a more general version of `lovr.headset.getHands`.  It is equivalent to the following Lua code:

```lua
function lovr.headset.glob(filter, ...)
  local devices = { 'head', 'hand/left', 'hand/right' } -- (i.e. all the devices)
  local result = {}
  for i, device in ipairs(devices) do
    local matched = not filter or device:match(filter)
    local available = ... == nil or lovr.headset.getStatus(device, ...) ~= 'unavailable'
    if matched and available then
      result[#result + 1] = device
    end
  end
  return result
end
-- lovr.headset.glob() --> all devices
-- lovr.headset.glob('hand') --> all hands
-- lovr.headset.glob('hand/', 'pose') --> all hands with a pose
-- lovr.headset.glob(nil, 'button', trigger') --> all devices with a trigger
```

- `lovr.headset.getHands` is sorta unofficially deprecated now, since if `lovr.headset.glob` turns out to be good you can just do `lovr.headset.glob('hand', 'pose')` to get the same behavior.
  - EDIT: It might still be nice to have getHands because you can't do `lovr.headset.glob('hand')` once there are more hand sub-devices like pointer, palm, fingers, etc., it would need to be `lovr.headset.glob('hand/%w+$')`

This enables a lot of cool device querying but it still sucks right now:

- We don't separate pose and velocity into linear and angular components, so we can't detect 3DOF vs. 6DOF.  This sort of thing is important for eye tracking.  I figure we can add the separated components later (it complicates the internal getPose interface) and keep the fused "pose" / "velocity" enums to be a boolean and/or of the other two.
- We still can't detect connectedness, which creates awkward situations and is kinda a bummer.  Connectedness is complicated, and I'm not sure it's even possible in OpenVR and OpenXR anymore.  It needs more thought, we don't want those drivers to always report connected/disconnected, maybe pose could be used as a proxy but that breaks gamepads too so, idk.
- There is a weird conflation between "unknown/disconnected" and "not tracked".  Ideally these are separate concepts, but separating them right now would require a lot of work.  I think it would be a good change to make so we can do better status queries.
- Related: It would be good to have a concept of "device capabilities" instead of always making decisions based on instantaneous tracking status.

### DIFF
--- a/src/api/l_headset.c
+++ b/src/api/l_headset.c
@@ -60,6 +60,23 @@ const char* DeviceAxes[] = {
   NULL
 };
 
+const char* TrackingTypes[] = {
+  [TRACKING_POSE] = "pose",
+  [TRACKING_VELOCITY] = "velocity",
+  [TRACKING_BUTTON] = "button",
+  [TRACKING_TOUCH] = "touch",
+  [TRACKING_AXIS] = "axis",
+  [TRACKING_SKELETON] = "skeleton",
+  NULL
+};
+
+const char* TrackingStatuses[] = {
+  [STATUS_UNAVAILABLE] = "unavailable",
+  [STATUS_ESTIMATED] = "estimated",
+  [STATUS_TRACKED] = "tracked",
+  NULL
+};
+
 typedef struct {
   lua_State* L;
   int ref;
@@ -97,19 +114,8 @@ static Device luax_optdevice(lua_State* L, int index) {
 }
 
 static int l_lovrHeadsetGetDriver(lua_State* L) {
-  if (lua_gettop(L) == 0) {
-    lua_pushstring(L, HeadsetDrivers[lovrHeadsetDriver->driverType]);
-    return 1;
-  } else {
-    Device device = luax_optdevice(L, 1);
-    FOREACH_TRACKING_DRIVER(driver) {
-      if (driver->getPose(device, NULL, NULL)) {
-        lua_pushstring(L, HeadsetDrivers[driver->driverType]);
-        return 1;
-      }
-    }
-  }
-  return 0;
+  lua_pushstring(L, HeadsetDrivers[lovrHeadsetDriver->driverType]);
+  return 1;
 }
 
 static int l_lovrHeadsetGetName(lua_State* L) {
@@ -287,16 +293,75 @@ static int l_lovrHeadsetGetBoundsGeometry(lua_State* L) {
   return 1;
 }
 
-static int l_lovrHeadsetIsTracked(lua_State* L) {
+static int l_lovrHeadsetGetStatus(lua_State* L) {
   Device device = luax_optdevice(L, 1);
-  float position[4], orientation[4];
-  FOREACH_TRACKING_DRIVER(driver) {
-    if (driver->getPose(device, position, orientation)) {
-      lua_pushboolean(L, true);
-      return 1;
-    }
+  TrackingType property = luaL_checkoption(L, 2, "pose", TrackingTypes);
+  DeviceButton button;
+  DeviceAxis axis;
+  float value[4], linear[4], angular[4];
+  bool down, changed;
+
+  switch (property) {
+    case TRACKING_POSE:
+      FOREACH_TRACKING_DRIVER(driver) {
+        if (driver->getPose(device, linear, angular)) {
+          lua_pushstring(L, TrackingStatuses[STATUS_TRACKED]);
+          lua_pushstring(L, HeadsetDrivers[driver->driverType]);
+          return 2;
+        }
+      }
+      break;
+
+    case TRACKING_VELOCITY:
+      FOREACH_TRACKING_DRIVER(driver) {
+        if (driver->getVelocity(device, linear, angular)) {
+          lua_pushstring(L, TrackingStatuses[STATUS_TRACKED]);
+          lua_pushstring(L, HeadsetDrivers[driver->driverType]);
+          return 2;
+        }
+      }
+      break;
+
+    case TRACKING_BUTTON:
+      button = luaL_checkoption(L, 3, NULL, DeviceButtons);
+      FOREACH_TRACKING_DRIVER(driver) {
+        if (driver->isDown(device, button, &down, &changed)) {
+          lua_pushstring(L, TrackingStatuses[STATUS_TRACKED]);
+          lua_pushstring(L, HeadsetDrivers[driver->driverType]);
+          return 2;
+        }
+      }
+      break;
+
+    case TRACKING_TOUCH:
+      button = luaL_checkoption(L, 3, NULL, DeviceButtons);
+      FOREACH_TRACKING_DRIVER(driver) {
+        if (driver->isTouched(device, button, &down)) {
+          lua_pushstring(L, TrackingStatuses[STATUS_TRACKED]);
+          lua_pushstring(L, HeadsetDrivers[driver->driverType]);
+          return 2;
+        }
+      }
+      break;
+
+    case TRACKING_AXIS:
+      axis = luaL_checkoption(L, 3, NULL, DeviceAxes);
+      FOREACH_TRACKING_DRIVER(driver) {
+        if (driver->getAxis(device, axis, value)) {
+          lua_pushstring(L, TrackingStatuses[STATUS_TRACKED]);
+          lua_pushstring(L, HeadsetDrivers[driver->driverType]);
+          return 2;
+        }
+      }
+      break;
+
+    case TRACKING_SKELETON:
+      break;
+
+    default: lovrThrow("Unreachable");
   }
-  lua_pushboolean(L, false);
+
+  lua_pushstring(L, TrackingStatuses[STATUS_UNAVAILABLE]);
   return 1;
 }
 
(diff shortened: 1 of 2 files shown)


bjornbytes:
It increasingly seems like these higher level functions like getHands/glob/getDevices belong in Lua.  Putting them in C has minimal benefit because it's just logic that does not query the hardware, isn't performance sensitive, and is trying to satisfy all the ways people could iterate/query devices.  The main factor is the convenience of having this logic readily available.

mcclure:
Thoughts

1. I *really* don't like the name glob(), personally, it communicates nothing about what the function does. I am not sure someone who is not a linux-head would even understand the etymology of "glob". I would suggest "getDevices" or if that sounds too long to you just "get"?

2. Why is "devices" in glob hardcoded? Why not just iterate over the HeadsetDriver enum as in https://github.com/bjornbytes/lovr/pull/221 ? That would mean you could get the eyes, if you wanted them.

3. "Estimated" tracking status corresponds to what? Oculus Mobile hand tracking with low confidence?

4. I think we should add one more TrackingType which is just something like "exists". I was describing a test app of mine in Slack that iterates over all devices and prints the information those devices expose. If we accept both this and the gamepad, there is no way to iterate over all devices (since hand-tracking hands don't respond to buttons and gamepads don't respond to pose). Adding "exists" would also provide one way of distinguishing between "the handset is connected, but pose tracking is lost" and "the handset is not connected at all" (especially useful with oculus mobile hand tracking hands, because they lose pose tracking frequently). (I understand OpenXR makes this information difficult to get but I think OpenXR made a bad decision there and I don't think we have to replicate their mistakes! I complained about this on their forum too 😛)

5. With TrackingType `buttons` or `axis`, could `getStatus` take one more argument to indicate *which* button/axis is being asked about? As I mentioned in Slack last night, currently there is no way to distinguish if a particular handset uses `thumbstick` or `touchpad` (because if the axis doesn't exist it still returns 0,0)

6. Instead of removing isTracked, I recommend having both getStatus and isTracked present, but with different semantics. isTracked could indicate that "at least" that status is supported. So if you say `lovr.headset.isTracked("hand/left", "pose", "estimated")`, this would return `true` if pose is estimated *or* tracked. 

7. Moving some of these functions into Lua sounds reasonable, where would we define them? boot.lua?

bjornbytes:
1. The name can be anything
2. Devices isn't hardcoded, it iterates over every member of the `Device` enum.
3. Estimated corresponds to a few different situations.  Low-confidence hand tracking, "estimated" hand tracking like knuckles/vive controllers have, eye tracking where the pose of an eye is estimated from the head position instead of being derived from an actual sensor value, etc.
4. I really want to support the "exists" one which is what I mentioned as "connectedness" above but I don't think it's easy right now.  You can iterate over all the devices by doing `lovr.headset.glob(nil)`, which returns everything.
5. `getStatus` accepts the button/axis in question already.
6. I considered this, but I didn't like having multiple ways of doing things.
7. I think they live best in people's Lua projects given how versatile the necessary behavior is.

mcclure:
2. Ah, sorry, got confused by the lua sample code above

4. "You can iterate over all the devices by doing lovr.headset.glob(nil), which returns everything" Surely that will return both connected and non-connected devices though?

    "but I don't think it's easy right now" I'd say it's pretty easy: Just add the callback to the driver interface? *Some* drivers can definitively and easily answer the question of whether a device is connected (and ones that can't say maybe could return "exists":"estimated" for everything)? I don't think it's a good idea to limit our feature set to only those features that *all* drivers support. If you need a feature some drivers don't support, you can choose to ship your app without support for those drivers. But if we drop features because OpenXR does them poorly we've dropped them even on drivers where they work…

5. Awesome. Sorry misread the code \*\_\*

6/7: I think keeping the lovr API small is good, but moving isTracked/getHands specifically into a third party library sounds a bit alarming. First off it is starting to get a bit troublesome that this is looking like the third release in a row where `getControllers`, the one function I call in every one of my lovr scripts, gets renamed \*\_\* But also we still don't have a fantastic story for distributing scripts/libraries. I'm distributing a sizable third-party lovr lib myself that is just picking up more and more stuff, and I don't have a good way to tell people to use it other than telling them to fork my repo.

Not to say we can't rely on third party libraries, but for the basic stuff I think it's friendlier if Lovr can keep the simple stuff simple. If nothing else, think about how it would look in sample code. Every current use of `getHands()` would have to become `getAll("hand")`. Every current use of isTracked would have to become `lovr.headset.getStatus("hand/left") == "estimated" or lovr.headset.getStatus("hand/left") == "tracked"`. The getHands change isn't actually so different here but the change to isTracked() instances seems like a real quality-of-life issue, and is harder to describe to new users…

bjornbytes:
Currently, as designed, the new getStatus function proposed in this PR does not return information but returns the level of validity of a piece of information.  So `lovr.headset.getStatus('device', 'axis', 'myaxis')` indicates whether you can call `lovr.headset.getAxis('device', 'myaxis')` in the current frame and expect to receive valid information.  The second glob function was added to return collections of devices which contain valid information for a certain property.  Note that this is a separate concept from both "the information itself" and from "the capabilities of the device".  Currently there is no way for input drivers to expose the capabilities of devices, you have to glean it implicitly by the stuff they return.

It is possible to expose connectivity information in the headset module in the same way.  The way to do this would be to call `isConnected(device)` to figure whether a device is connected, and `getStatus(device, 'connection')` to figure out if that information is valid, estimated, etc.

It doesn't seem like this would be a satisfactory solution to the problem of retrieving every connected device because `getDevices(nil, 'connection')` would only return the devices that are known to be connected or not.  There would need to be a second pass over the devices that calls `isConnected` to read the actual data, just like how isDown or getAxis is used.

Given that exposing connectivity as a tracked property would still be inconvenient for one of the use cases expressed here, it's possible that there would need to be another helper function like `getConnectedDevices` (iterate over all devices and filter by the actual data rather than the validity of the data), or the consideration that connectivity should not be treated as a sensor value but given some kind of special treatment.  If it was treated as special then it probably does not belong in `getStatus` anyway, and would need to have some other system for querying the data and filtering devices.  It's also possible that the inconsistency here is an indicator that the general system proposed in this PR is not universally useful and should be redesigned.

mcclure:
> Currently there is no way for input drivers to expose the capabilities of devices, you have to glean it implicitly by the stuff they return.

I'm proposing adding such a function to the input drivers. At minimum the Oculus Mobile, Oculus Quest, Gamepad and Leap Motion drivers know exactly which capabilities every connected device they expose have.

> It doesn't seem like this would be a satisfactory solution to the problem of retrieving every connected device because getDevices(nil, 'connection') would only return the devices that are known to be connected or not. There would need to be a second pass over the devices that calls isConnected to read the actual data

(In this case you're hypothesizing adding a 'connected' tracking type that isn't present currently, correct?)

Anyway, that (two passes, first get the list of connected devices and then query getStatus for each one's TrackingTypes) would be the approach I'd take, yes. The thing is, I can only do that second pass if I have the "all connected devices" list to iterate over in the first place.

Consider what an app that wants to discover attached devices and their capabilities would look like _without_ the "all connected devices" iterator. I'd have to one-by-one query glob for each of "pose",  "buttons" and "axis", and build a table using the device name as a key. That is more complicated and more awkward (and might even miss things; if there is a hand-tracking hand "connected" but not currently tracking, that hand will respond "unavailable" to _all_ of getStatus "axis", "button" and "pose"). So in that case I'd definitely prefer a "connected" TrackingType.

bjornbytes:
I'm not 100% sure now but I think there is a misunderstanding of what a 'connected' TrackingType would mean if it was added.

`getStatus` could also be named "getConfidence" because it returns how "valid" a piece of information is when it's queried.  `getDevices`, the way it exists in this PR, filters based on this confidence value.  So `lovr.headset.getDevices('connected')` would not return a list of connected devices.  It would return a list of devices that have a high confidence about their connectivity.  You would then need to iterate over these and call `isConnected` on each one:

```lua
local connectedDevices = {}
for _, device in ipairs(lovr.headset.getDevices(nil, 'connection')) do
  if lovr.headset.isConnected(device) then
    table.insert(connectedDevices, device)
  end
end
return connectedDevices
```

It's analogous to how `getStatus('hand/left', 'button', 'trigger')` gives you the confidence level of the trigger button but you still need to call `isDown` to get the current state of the button.  This is why I don't think a TrackingType of 'connected' has very much utility without a separate helper function, and what you are really looking for is a special-cased `getConnectedDevices` function that actually reads the `isConnected` state.  And this raises questions for me about the usefulness of querying devices by confidence levels.

I've been thinking about these queries as belonging to 3 different categories:

- **Capabilities** -- "will valid information ever be returned?" -- These are usually static properties of the SDK/hardware.  Things like "do the hand devices have analog grip axes" or "does the headset have eye tracking".  Currently LÖVR doesn't expose this, but I think it's a really useful way to ask questions about devices that we are missing right now.
- **Confidence** -- "is valid information currently being returned?" -- You can use this to figure out if tracking is lost, or if information is being indirectly inferred from some other source.  This is what was originally being communicated by returning `nil` from the various accessors, and is now being brought back by the `getStatus` function.
- **Actual Data** -- "what is the information?" -- This is the stuff actually returned from getPose, isDown, getSkeleton, isConnected, etc.

bjornbytes:
I probably won't be merging this in its current form

PR #224, opened 2020-02-17
Fix build with Oculus SDK 1.43.0

mcclure:
The path to libOVR.lib has changed in more recent versions of the Oculus SDK. They no longer ship separate Debug and Release binaries. CMakeLists must change to reflect this.


bjornbytes:
Sweet

bjornbytes:
There is this comment but I think it's still correct

https://github.com/bjornbytes/lovr/blob/master/CMakeLists.txt#L269

mcclure:
We can consider changing the comment later

PR #225, opened 2020-02-17
Fix ref.h in recent MSVC when compiled as C++

mcclure:
Okay. This… one's… weird.

We pass ref directly to _InterlockedIncrement on Windows. However the argument to _InterlockedIncrement is declared volatile and Ref is not declared volatile. Apparently this is fine in C, and it is fine in C++ in both Clang and MSVC 2017. However if you compile Lovr in MSVC 2019, and you have added a .cpp file which #includes ref.h (which is something I have done in my private fork), this will fail. This patch adds a cast which only activates when ref.h is included into a C++ file.

I do not know if this is the best way to do this. It would be better to add a compile flag that tells MSVC 2019 it is okay to pass a `long *` to a `volatile long *` without a cast. But I don't know how to do that.

### DIFF
--- a/src/core/ref.h
+++ b/src/core/ref.h
@@ -17,12 +17,18 @@ static inline uint32_t ref_dec(Ref* ref) { return --*ref; }
 
 #elif defined(_MSC_VER)
 
+#ifdef __cplusplus
+#define _LOVR_REF_H_CPP_CAST (volatile long *)
+#else
+#define _LOVR_REF_H_CPP_CAST
+#endif
+
 // MSVC atomics
 
 #include <intrin.h>
 typedef uint32_t Ref;
-static inline uint32_t ref_inc(Ref* ref) { return _InterlockedIncrement(ref); }
-static inline uint32_t ref_dec(Ref* ref) { return _InterlockedDecrement(ref); }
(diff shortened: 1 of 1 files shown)


bjornbytes:
Is there a reason Ref can't be declared as volatile on win32?

bjornbytes:
At some point I would like to try something like this

```c
#ifdef __cplusplus // maybe also #ifdef _WIN32 now ?
#include <atomic>
using std::atomic_uint;
using std::atomic_fetch_add;
#else
#include <stdatomic.h>
#endif
```

There might be a theoretical binary incompatibility here, but maybe that can be sacrificed.

Also if it's ever the case that you no longer need ref.h to be C++ compatible, it would be helpful to know that.

mcclure:
> "Is there a reason Ref can't always be declared as volatile on win32?"

I tried that, but I got errors because you can't pass volatile long * to free. Apparently we will have to cast either when we call Interlocked or when we call free().

> There might be a theoretical binary incompatibility here, but maybe that can be sacrificed.

If you are including ref.h from a C++ file, it must be specifically for the purposes of interop with Lovr. Atomic and stdatomic are not binary cross-compatible. So allowing C++ files to use std::atomic is probably worse than just erroring when a C++ file tries to load ref.h.

PR #227, opened 2020-02-17
Instead of showing post-distortion in Oculus mirror, show the left eye (no stereo)

mcclure:
Combine this with an upgrade to Oculus SDK 1.43 (which fixes the frame 3 glitch) and https://github.com/bjornbytes/lovr/pull/128 , and the Oculus mirror is now very nice. Now if only we could do something about rendering when the headset comes off…

I'm not sure if we should set this to LeftEye (which I did) or to 0 (would show the eyes "flat" but side-by-side, like OpenVR). Maybe it would make sense to have a conf.lua value at some point indicating whether people would rather have the mirror be stereo or mono.

### DIFF
--- a/src/modules/headset/oculus.c
+++ b/src/modules/headset/oculus.c
@@ -333,7 +333,7 @@ static void oculus_renderTo(void (*callback)(void*), void* userdata) {
       .Width = lovrGraphicsGetWidth(),
       .Height = lovrGraphicsGetHeight(),
       .Format = OVR_FORMAT_R8G8B8A8_UNORM_SRGB,
-      .MirrorOptions = ovrMirrorOption_PostDistortion
+      .MirrorOptions = ovrMirrorOption_LeftEyeOnly
     };
     lovrAssert(OVR_SUCCESS(ovr_CreateMirrorTextureWithOptionsGL(state.session, &mdesc, &state.mirror)), "Unable to create mirror texture");
 


bjornbytes:
There used to be mirroring options; I think the intent now is to use the `lovr.mirror` callback to render the mirror texture (maybe using uv arguments of lovr.graphics.fill) if an app wants something other than the default.

(For this reason, using a stereo mirror texture for Oculus might be slightly better because then you're at least able to render the eye you want instead of being stuck with one of them, but I don't think it's a huge deal).

It might be that the fixed-function mirroring options were removed because they were tricky to implement in C (for every driver).  Now that `lovr.mirror` is in Lua, fixed-function mirroring options might be more successful.

For Oculus, I wonder if using `state.canvas` as the mirror texture A) is valid (might not be due to swapchain) B) would render stuff properly when the headset is off.

mcclure:
> I think the intent now is to use the lovr.mirror callback to render the mirror texture (maybe using uv arguments of lovr.graphics.fill) if an app wants something other than the default.

One problem with this approach is getting the mirror texture and "stretching" it might give you a different result. I think the mirror textures Oculus emits when you specifically request left eye are a different aspect ratio than the ones when you request stereo (I can't tell if it's including additional information or just cropping at the top and bottom)

mcclure:
> , I wonder if using state.canvas as the mirror texture A) is valid

That's interesting

PR #228, opened 2020-02-17
Fix lovr.graphics.skybox()

mcclure:
Skybox was failing with error "Regular non-array variable 'uv' may not be redeclared ". Did something go wrong when lovrMain {} got backed out?


bjornbytes:
Yeah, here's where it got changed

https://github.com/bjornbytes/lovr/commit/6ad2d8ccefa45cf8e774ea53374ca25d4fa43c22#diff-50aefc58abab4409ebecb33b746d5001L311-R317

PR #238, opened 2020-04-12
implement haptics on oculus desktop

shakesoda:
fixes the todo. implemented with a timer so that durations should work, tested working with the following:

```lua
function lovr.draw()
	for i, hand in ipairs(lovr.headset.getHands()) do
		lovr.headset.vibrate(hand, lovr.headset.getAxis(hand, 'trigger'), 0, 0)

		local x, y, z = lovr.headset.getPosition(hand)
		lovr.graphics.sphere(x, y, z, .1)
	end
end
```

it updates the controller haptics state every frame, but I have no reason to believe this is a problem from the docs or test.

### DIFF
--- a/src/modules/headset/oculus.c
+++ b/src/modules/headset/oculus.c
@@ -27,6 +27,10 @@ static struct {
   Canvas* canvas;
   ovrTextureSwapChain chain;
   ovrMirrorTexture mirror;
+  float hapticFrequency[2];
+  float hapticStrength[2];
+  float hapticDuration[2];
+  double hapticLastTime;
   arr_t(Texture*) textures;
   map_t textureLookup;
 } state;
@@ -44,6 +48,10 @@ static Texture* lookupTexture(uint32_t handle) {
   return state.textures.data[index];
 }
 
+static double oculus_getDisplayTime(void) {
+  return ovr_GetPredictedDisplayTime(state.session, state.frameIndex);
+}
+
 static ovrTrackingState *refreshTracking(void) {
   static ovrTrackingState ts;
   if (!state.needRefreshTracking) {
@@ -59,7 +67,7 @@ static ovrTrackingState *refreshTracking(void) {
 
   // get the state head and controllers are predicted to be in at display time,
   // per the manual (frame timing section).
-  double predicted = ovr_GetPredictedDisplayTime(state.session, state.frameIndex);
+  double predicted = oculus_getDisplayTime();
   ts = ovr_GetTrackingState(state.session, predicted, true);
   state.needRefreshTracking = false;
   return &ts;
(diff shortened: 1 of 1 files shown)


src/modules/headset/oculus.c
```diff
@@ -308,7 +312,18 @@ static bool oculus_getAxis(Device device, DeviceAxis axis, vec3 value) {
 }
 
 static bool oculus_vibrate(Device device, float strength, float duration, float frequency) {
-  return false; // TODO
+  int idx = device == DEVICE_HAND_LEFT ? 0 : 1;
```
bjornbytes, line 318: Note that this will vibrate the right hand if you do things like `lovr.headset.vibrate('head')`.

src/modules/headset/oculus.c
```diff
@@ -308,7 +312,18 @@ static bool oculus_getAxis(Device device, DeviceAxis axis, vec3 value) {
...
+  state.hapticStrength[idx] = strength > 0.0f ? (strength <= 1.0f ? strength : 1.0f) : 0.0f;
+  state.hapticDuration[idx] = duration > 0.0f ? duration : 0.0f;
+  float freq = frequency / 320.0f; // 1.0 = 320hz, limit on Rift CV1 touch controllers.
+  if (freq < 0.0f) {
+    freq = 0.0f;
+  }
+  if (freq > 1.0f) {
+    freq = 1.0f;
+  }
```
bjornbytes, line 327: If you want, there are MIN, MAX, and CLAMP macros in util, should already be available in this file.

src/modules/headset/oculus.c
```diff
@@ -349,6 +367,16 @@ static void oculus_renderTo(void (*callback)(void*), void* userdata) {
...
+  float delta = (float)(state.hapticLastTime - sensorSampleTime);
+  state.hapticLastTime = sensorSampleTime;
+  for (int i = 0; i < 2; ++i) {
+    ovr_SetControllerVibration(state.session, ovrControllerType_LTouch + i, state.hapticFrequency[i], state.hapticStrength[i]);
+    state.hapticDuration[i] -= delta;
+    if (state.hapticDuration[i] <= 0.0f) {
+      state.hapticStrength[i] = 0.0f;
+    }
+  }
+
```
bjornbytes, line 379: Maybe this belongs in `oculus_update`?  Feels more update-y than draw-y to me, but maybe here in renderTo makes more sense since you can access the predicted display time a bit more easily.  Not sure.

src/modules/headset/oculus.c
shakesoda: done

src/modules/headset/oculus.c
shakesoda: oculus seems to want the controller updates to happen with frames and the timing information is already available, so it seemed most appropriate to put it here (despite being a bit non-render-y). I can move it if you want, but this is a bit simpler.

PR #239, opened 2020-04-15
Create resources in CMake build

mcclure:
The CMakeLists in my private fork has this wad of junk at the end approximating an xxd invocation. I did not write this and I have no idea where it came from. If I had to guess I'd say I got it from Bjorn in slack? It works well.


bjornbytes:
This wad of junk is already floating in the sea of CMake junk [here](https://github.com/bjornbytes/lovr/blob/eb0379522f88052ef998f88005815383eaf24ee4/CMakeLists.txt#L299)!

mcclure:
Well, that explains it.

PR #241, opened 2020-04-16
Oculus mobile "restart cookie" support; lovr.event.restart() changes

mcclure:
After the changes to add a "cookie" to `lovr.event.quit("restart")`, restarting was broken on Oculus Mobile. This patches oculus mobile so restarting and the cookie work.

In addition, after a slack conversation, these changes were made to quit("restart") and the cookie:

- lovr.event.quit("restart") no longer supported
- lovr.event.quit no longer accepts restart "cookie"
- When lovr.event.restart() called, lovr.quit is not called, instead lovr.restart() is called
- Value returned from lovr.restart() becomes the cookie
- lovr.event.quit receives the lovr.event.quit() return code as an argument

(The return values from lovr.run() are not changed, but boot.lua picks the return values differently and the event code is simplified to remove the cookie plumbing.)

These changes mean that a cookie can be obtained if a restart is triggered by F5 or lodr (as long as lodr is patched so makeWrapper calls lovr.event.restart() instead of returning "restart").

My test program:

```
local count = arg.restart or 0

function lovr.load()
	print(string.format("%d boots", count))
end

local time = 0
function lovr.update(dt)
	time = time + dt
	if time > 20 then
		print("TIME OUT!")
		lovr.event.restart()
	end
end
   
function lovr.restart()
	print("restarting")
	return count + 1
end
```

### DIFF
--- a/src/api/l_event.c
+++ b/src/api/l_event.c
@@ -170,12 +170,7 @@ static int l_lovrEventPush(lua_State* L) {
 static int l_lovrEventQuit(lua_State* L) {
   EventData data;
 
-  int argType = lua_type(L, 1);
-  if (argType == LUA_TNUMBER || lua_isnoneornil(L, 1)) {
-    data.quit.exitCode = luaL_optint(L, 1, 0);
-  } else {
-    return luaL_argerror (L, 1, "number or nil expected");
-  }
+  data.quit.exitCode = luaL_optint(L, 1, 0);
 
   EventType type = EVENT_QUIT;
   Event event = { .type = type, .data = data };
--- a/src/api/l_event.c
+++ b/src/api/l_event.c
@@ -9,6 +9,7 @@
 
 StringEntry EventTypes[] = {
   [EVENT_QUIT] = ENTRY("quit"),
+  [EVENT_RESTART] = ENTRY("restart"),
   [EVENT_FOCUS] = ENTRY("focus"),
 #ifdef LOVR_ENABLE_THREAD
   [EVENT_THREAD_ERROR] = ENTRY("threaderror"),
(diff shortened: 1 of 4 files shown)


src/api/l_event.c
```diff
@@ -176,15 +171,10 @@ static int l_lovrEventQuit(lua_State* L) {
...
-    data.quit.exitCode = 0;
-    luax_checkvariant(L, 2, &data.quit.cookie);
-  } else if (argType == LUA_TNUMBER || lua_isnoneornil(L, 1)) {
-    data.quit.restart = false;
+  if (argType == LUA_TNUMBER || lua_isnoneornil(L, 1)) {
     data.quit.exitCode = luaL_optint(L, 1, 0);
   } else {
-    return luaL_argerror (L, 1, "number, nil or the exact string 'restart' expected");
+    return luaL_argerror (L, 1, "number or nil expected");
   }
```
bjornbytes, line 178: This ifelse could be reduced to just `luaL_optint(L, 1, 0)` since opt* will error if the thing isn't nil or the appropriate type.  Error message will probably be more canonical too.

src/resources/boot.lua
```diff
@@ -149,8 +149,12 @@ function lovr.run()
   return function()
     lovr.event.pump()
     for name, a, b, c, d in lovr.event.poll() do
-      if name == 'quit' and (not lovr.quit or not lovr.quit()) then
-        return a or 0, b
+      if name == 'restart' then
+        local cookie = lovr.restart and lovr.restart()
+        return 'restart', cookie
+      end
+      if name == 'quit' and (not lovr.quit or not lovr.quit(a)) then
```
bjornbytes, line 156: Should this be `elseif`?  idk, not important

mcclure: They are logically equivalent but the elseif reads better.

PR #243, opened 2020-04-21
Fix when depth test is disabled but depth write is enabled

pendingchaos:
`GL_DEPTH_TEST` controls both whether depth testing and depth writes are enabled. So if depth testing is disabled and depth writes are enabled, `GL_DEPTH_TEST` has to be enabled and the compare mode should be `GL_ALWAYS`.

`gpu_gl.c` and `gpu_vk.c` seem to have a similar problem, but that code seems unused, so I can't test it.

### DIFF
--- a/src/modules/graphics/opengl.c
+++ b/src/modules/graphics/opengl.c
@@ -974,25 +974,28 @@ static void lovrGpuBindPipeline(Pipeline* pipeline) {
     }
   }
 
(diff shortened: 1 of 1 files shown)


bjornbytes:
Thanks!

Yeah, core/gpu_* stuff is still WIP, I left a note to apply this fix next time I resume work there.

Issue #219, opened 2020-02-11
Physics API should support vector objects


bjornbytes:

Ulydev:
For methods with multiple vector parameters, should it support any combinations of x,y,z and vectors?
As an example, would https://lovr.org/docs/v0.13.0/Collider:applyForce accept only (x, y, z, px, py, pz) and (v, pv) or also (x, y, z, pv) and (v, px, py, pz)?

bjornbytes:
I think the setters will accept any combination of vectors and numbers.  This is similar to how other functions work, and there is a `luax_readvec3` helper that makes it pretty easy.

For getters, I'm not sure.  I think those would probably stick to just returning numbers.  You can always wrap it in a `vec3` or pass it to `vec3:set`, which isn't too bad.

bjornbytes:
Arg `World:raycast` is sooo painful without vectors!

jmiskovic:
Even when accepting either vectors or numbers, raycast is still painful because of returned values are 6 numbers. Maybe optional argument could trigger using vec3 in raycast callback? Is this too complicated?

bjornbytes:
Yeah, I think a parameter to control the return format is kind of awkward.  An option that seems a little cleaner to me would be to accept "out args", where you can optionally pass 2 vectors as the last arguments, and they get filled in as the result.

I'm still trying to figure out whether it's okay to have functions in the API return vectors in general.  There are things to think about like whether they should be permanent or temporary vectors, and if this could be annoying for people that aren't interested in using vector objects or something.

jmiskovic:
My motivation is to drop xyz coordinates altogether from Lua code. Vector math is easier to write and read, less verbose and more high level. Several places in API make this hard and require manual conversion, for example syncing physics and graphics. Ideally, something like this should work: `lovr.graphics.box('fill', boxCollider:getPose())`. But that would mean returning mat4 object and not `x,y,z,angle,ax,ay,az`.

Regarding returning permanent vs temp vectors. Always permanent is not an option, raycasting is common operation. I would be ok with API always returning temp vectors. Better than having to remember which function returns which. LOVR quite helpfully points out where exactly in code the expired temp vector was used.

PR #247, opened 2020-04-26
Fix collider setOrientation() and setPose()

jmiskovic:
The conversion from `angle,ax,ay,az` to quaternion is already done in physics module.

### DIFF
--- a/src/modules/physics/physics.c
+++ b/src/modules/physics/physics.c
@@ -517,9 +517,7 @@ void lovrColliderGetOrientation(Collider* collider, float* angle, float* x, floa
   quat_getAngleAxis(quaternion, angle, x, y, z);
 }
 
-void lovrColliderSetOrientation(Collider* collider, float angle, float x, float y, float z) {
-  float quaternion[4];
-  quat_fromAngleAxis(quaternion, angle, x, y, z);
+void lovrColliderSetOrientation(Collider* collider, float* quaternion) {
   float q[4] = { quaternion[3], quaternion[0], quaternion[1], quaternion[2] };
   dBodySetQuaternion(collider->body, q);
 }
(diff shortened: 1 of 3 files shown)


bjornbytes:
Whoops, good catch.  Since this removes the ability to use quaternions with setOrientation/setPose, could we instead remove the angle/axis -> quaternion conversion in the physics module?

jmiskovic:
Makes sense. It would be great to revise the rest of physics API and make everything compatible with vec3, quat and mat4.

src/modules/physics/physics.c
```diff
@@ -517,10 +517,7 @@ void lovrColliderGetOrientation(Collider* collider, float* angle, float* x, floa
   quat_getAngleAxis(quaternion, angle, x, y, z);
 }
 
-void lovrColliderSetOrientation(Collider* collider, float angle, float x, float y, float z) {
-  float quaternion[4];
-  quat_fromAngleAxis(quaternion, angle, x, y, z);
-  float q[4] = { quaternion[3], quaternion[0], quaternion[1], quaternion[2] };
```
bjornbytes, line 523: I think we still need to swizzle the quaternion components here (unless it's working as is)

src/modules/physics/physics.h
```diff
@@ -139,7 +139,7 @@ void lovrColliderSetMassData(Collider* collider, float cx, float cy, float cz, f
 void lovrColliderGetPosition(Collider* collider, float* x, float* y, float* z);
 void lovrColliderSetPosition(Collider* collider, float x, float y, float z);
 void lovrColliderGetOrientation(Collider* collider, float* angle, float* x, float* y, float* z);
-void lovrColliderSetOrientation(Collider* collider, float angle, float x, float y, float z);
+void lovrColliderSetOrientation(Collider* collider, dQuaternion q);
```
bjornbytes, line 142: Maybe a minor thing, but I try to avoid library-specific types in the function signatures.  So I would prefer using a `float*` / `quat` type here.

bjornbytes:
Also #219 is tracking the work to fully convert the physics module over to accept vectors.

src/modules/physics/physics.c
jmiskovic: Ouch. Yeah, that reordering is still needed.

bjornbytes:
This looks great, thanks!

PR #249, opened 2020-04-29
Initialize inertia information for new colliders

jmiskovic:
Inertia matrix has to be calculated with algorithm specific for each
shape. Without calculated matrix the physics behaves very floaty and
slowed down.

This commit will fix physics for most users that only use world:new___Collider(). 

It will still not work correctly if shapes are manually added to collider. For that case the inertia matrix would have to be re-calculated for each new shape, which is ...complicated. It needs more discussion because API would have to be adapted.

I also increased MAX_CONTACTS to 10 because 4 is too low. The number represents maximum handled simultaneous collisions as well as connected joints.


bjornbytes:
This is really exciting, thanks!

PR #251, opened 2020-05-01
Physics world accepts both vec3 and number coords

jmiskovic:
When creating colliders, setting gravity or casting rays on world,
arguments accept either coordinate numbers as before, or vec3 objects.
For functions that receive more than one set of coordinates, any
combination of coordinates and vectors is accepted.

Work towards solving #219, only World API is included here.

### DIFF
--- a/src/api/l_physics_world.c
+++ b/src/api/l_physics_world.c
@@ -40,10 +40,9 @@ static void raycastCallback(Shape* shape, float x, float y, float z, float nx, f
 
 static int l_lovrWorldNewCollider(lua_State* L) {
   World* world = luax_checktype(L, 1, World);
-  float x = luax_optfloat(L, 2, 0.f);
-  float y = luax_optfloat(L, 3, 0.f);
-  float z = luax_optfloat(L, 4, 0.f);
-  Collider* collider = lovrColliderCreate(world, x, y, z);
+  float position[4];
+  luax_readvec3(L, 2, position, NULL);
+  Collider* collider = lovrColliderCreate(world, position[0], position[1], position[2]);
   luax_pushtype(L, Collider, collider);
   lovrRelease(Collider, collider);
   return 1;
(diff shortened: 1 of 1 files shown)


src/api/l_physics_world.c
```diff
@@ -40,25 +40,25 @@ static void raycastCallback(Shape* shape, float x, float y, float z, float nx, f
...
-  float sz = luax_optfloat(L, 7, sx);
-  Collider* collider = lovrColliderCreate(world, x, y, z);
-  BoxShape* shape = lovrBoxShapeCreate(sx, sy, sz);
+  float position[4], size[4];
+  int index = luax_readvec3(L, 2, position, NULL);
+  luax_readvec3(L, index, size, NULL);
+  for (int i = 0; i < 3; i++) { // defaults
+    if (size[i] == 0.f)
+      size[i] = i == 0 ? 1.f : size[0];
+  }
```
bjornbytes, line 59: There is a function `luax_readscale` that could be used here, it fills in defaults like this.  The behavior is _slightly_ different since it applies the default if the argument is missing instead of zero, but this should be okay.

I don't mind applying this change myself after merging, just let me know.

src/api/l_physics_world.c
```diff
@@ -223,15 +219,15 @@ static int l_lovrWorldSetSleepingAllowed(lua_State* L) {
...
-  luaL_checktype(L, 8, LUA_TFUNCTION);
-  lua_settop(L, 8);
-  lovrWorldRaycast(world, x1, y1, z1, x2, y2, z2, raycastCallback, L);
+  float start[4], end[4];
+  int index;
+  index = luax_readvec3(L, 2, start, NULL);
+  index = luax_readvec3(L, index, end, NULL);
+  luaL_checktype(L, index, LUA_TFUNCTION);
+  lua_settop(L, index);
+  lovrWorldRaycast(world,
```
bjornbytes, line 228: Minor style: I'd prefer to keep this unwrapped since it's still around 100 characters.

src/api/l_physics_world.c
jmiskovic: Fixed.

PR #263, opened 2020-05-29
Physics simulation access to global ERP and CFM

jmiskovic:
ERP (Error Reduction Parameter) controls the amount of corrective force
applied to colliders that belong to joints. With value of 0, no
corrective force is applied and joined colliders will drift off in time.
With value close to 1, strong force will ensure precision of joints.

CFM (Constraint Force Mixing) controls the rigidness/springiness of
simulation. Lower values make constraints "hard" and result in more
unstable simulation, while values closer to 1 give spongy and soft
simulation.

### DIFF
--- a/src/api/l_physics_joints.c
+++ b/src/api/l_physics_joints.c
@@ -122,10 +122,42 @@ static int l_lovrBallJointSetAnchor(lua_State* L) {
   return 0;
 }
 
+static int l_lovrBallJointGetResponseTime(lua_State* L) {
+  Joint* joint = luax_checkjoint(L, 1);
+  float responseTime = lovrBallJointGetResponseTime(joint);
+  lua_pushnumber(L, responseTime);
+  return 1;
+}
+
+static int l_lovrBallJointSetResponseTime(lua_State* L) {
+  Joint* joint = luax_checkjoint(L, 1);
+  float responseTime = luax_checkfloat(L, 2);
+  lovrBallJointSetResponseTime(joint, responseTime);
+  return 0;
+}
+
+static int l_lovrBallJointGetTightness(lua_State* L) {
+  Joint* joint = luax_checkjoint(L, 1);
+  float tightness = lovrBallJointGetTightness(joint);
+  lua_pushnumber(L, tightness);
+  return 1;
+}
+
+static int l_lovrBallJointSetTightness(lua_State* L) {
+  Joint* joint = luax_checkjoint(L, 1);
+  float tightness = luax_checkfloat(L, 2);
+  lovrBallJointSetTightness(joint, tightness);
+  return 0;
+}
+
 const luaL_Reg lovrBallJoint[] = {
   lovrJoint,
   { "getAnchors", l_lovrBallJointGetAnchors },
   { "setAnchor", l_lovrBallJointSetAnchor },
+  { "getResponseTime", l_lovrBallJointGetResponseTime},
+  { "setResponseTime", l_lovrBallJointSetResponseTime},
+  { "getTightness", l_lovrBallJointGetTightness},
+  { "setTightness", l_lovrBallJointSetTightness},
   { NULL, NULL }
 };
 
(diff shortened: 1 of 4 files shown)


bjornbytes:
Questions

- Is it common to change this at runtime?  Could it be a setting that is provided only at init time?
- Are there more intuitive names we could pick instead of the acronyms/descriptions of them?  Example: "correction force" and "sponginess".  That would make them easier to remember and make them a little easier to use.

jmiskovic:
Good questions. Tweaking these parameters is important to customize how physics feel and how stable the simulation is. For me it's ok for them to be passed into world initialization because LOVR starts up so quickly. It might only cause problems if users decide to tweak them in projects with long loading time.

I agree acronyms are not intuitive, but there is no easy and short name. Maybe ERP can be "correction factor" and CFM could become "simulation softness", but this places more burden on LOVR docs to explain things that are covered in much details in ODE manual. Your call.

I would also like to expose joint CFM parameter to be able to control bounciness and damping of individual joint. This one would need to be changeable in runtime. With this change it should be possible to achieve most common physics constructs.

bjornbytes:
Ok, let's keep it the way it is here with getters/setters.

What do you think about `World:setStrictness` (ERP) and `World:setSoftness` (CFM)?  I kinda like how they sound the same and alphabetize together, which might be helpful since it seems like these settings would often be configured together.  For ERP I think 'correction factor' would work fine too, if you want.  The docs should definitely mention ERP and CFM for those familiar with ODE/physics sims.

One other thing: should these be clamped to 0 - 1?  ODE docs mention this range and also say that negative values can cause instability.

It's cool that you can use these to do spring forces!  I struggled in the past with getting good spring joints.

src/api/l_physics_joints.c
```diff
@@ -95,6 +123,8 @@ static int l_lovrJointSetEnabled(lua_State* L) {
 #define lovrJoint \
   { "destroy", l_lovrJointDestroy }, \
   { "getType", l_lovrJointGetType }, \
+  { "getCFM", l_lovrJointGetCFM}, \
+  { "setCFM", l_lovrJointSetCFM}, \
```
bjornbytes, line 127: Note: ERP functions are missing here

jmiskovic:
That PU update was not intentional, the code is not ready. I took some time tweaking parameters in runtime to see how it all behaves.

ERP is amount of force the joint exerts on collider. Low values make joints loose, high values make it tight and can cause collider to overshot the joint target. With ERP set to 0 the joint loses its function. Going above 1 puts even more energy into joint oscillations, which is desirable sometimes.

The CFM affects the response time of physics simulation, both for collisions and for joint inertia. Low CFM values make simulation tight and fast, higher values make it sluggish. For collisions it affects how fast penetration is resolved, with higher values resulting in spongy objects with more surface penetration and slower collision resolving. For joints the CFM is similar to inertia, with higher CFM values resulting in slow oscillations. The oscillation frequency is also affected by collider mass, so CFM can be used to tweak the joint to get desired frequency with specific collider mass. Values higher than 1 are often desirable, especially for very light objects. Unlike ERP, CFM is tweaked in orders of magnitude with useful values (depending on mass) being between 10^-8 and 10^8.

Third parameter that affects oscillation is collider angular/linear damping. These 3 parameters can be tweaked to get perfect tightness, inertia and damping for springs. Based on this, we could rename setERP() into setTightness(), and setCFM() to setResponseTime().

Negative values of ERP or CFM cause instability and simulation crashing after seconds. I'll clamp ERP and CFM to values > 0.

Frustratingly, ERP and CFM can only be set for distance and ball joint, while slider and hinge will always use world settings. That's based on experimenting and reading ODE code. The manual is not clear on this and there's not much of other material to go by. I'll add setters/getters just for distance and ball joints.

jmiskovic:
There, should be complete. 

I used assert instead of clamping because a) there is no upper limit, b) I felt it's better to throw error than silently modify supplied value. Of course, feel free to modify the API.

bjornbytes:
These physics changes have been great!  Thanks for doing the detailed investigation on these parameters too.

PR #264, opened 2020-06-04
lovrMicrophoneGetData: fix potential buffer overrun

nevyn:
With the check for samples==0 being done BELOW the assert for offset+samples<soundData->samples,
setting samples to 0 and then having more samples available in the mic than present in
the created buffer would cause buffer overrun


nevyn:
Credit to @voxar for finding this bug :) It didn't end up being our crasher so I haven't experienced it in the wild but it certainly looks like it could crash to me.

PR #266, opened 2020-06-05
Bump openal to master

nevyn:
Previous version was 2 years old. Just bumping seems to work fine;
both playback and recording works well for me on windows.


nevyn:
well...... that didn't build...... my CI and your CI are failing in completely different ways. I'll try to figure out mine, and if it applies to this branch I'll try to fix it here too, otherwise I'll just abandon it since you're planning to abandon openal pretty soon anyway

bjornbytes:
Yeah, I think if LÖVR was planning to stick with OpenAL then there would be good reasons to upgrade, but since the miniaudio switchover is relatively imminent I think I'll hold off on this.

Issue #212, opened 2020-02-01
lovr.timer.getTime only changes at updates on Oculus Mobile.


porglezomp:
I was trying to implement a rough CPU timer with `lovr.timer.getTime()` as a first pass and was constantly getting 0. It seems like it's set to only change at events. This should at least be documented, and preferably be changed to give a live version of the time if possible (although it's currently doing stuff to handle pausing and things, so I imagine there's a good reason it is the way it is that I'm not aware of).

bjornbytes:
Hm, that seems bad.  It's returning the headset's predicted display time and using that as the implementation of getTime in the platform layer (which is what `lovr.timer.getTime` queries).  The headset interface has a separate `getDisplayTime` function where the predicted display time is meant to be returned.

I don't fully understand the pause stuff going on but maybe this could be tracked as an "epoch" that is separate from the various timers?  cc @mcclure for more info, I think the pause stuff was tricky/buggy and interacts with android lifecycle stuff...

mcclure:
This… was… me misunderstanding what lovr.timer.getTime was supposed to do, I think.

The complicated pausing stuff is MOSTLY to work around a particular nasty OS bug from early in the Oculus Go's lifespan. That bug has since been fixed so I can simplify the pause logic now. 

Here is a question: Say you pause the app, and then you resume it 20 seconds later (ie you go to the main menu, and then you return). Should lovr.timer.getTime() indicate 20 seconds passed?

If you're basing dynamics on lovr.timer.getTime(), a jump of 20 seconds could make you do… bad things.

bjornbytes:
I think it would advance by 20 seconds because `timer.getTime` is supposed to be wallclock time.  But maybe you could hook into `lovr.focus` events to detect when the pauses happen if you need to stop your game's timers?  Hm

bjornbytes:
It's going to be challenging to emit focus events on android because if we were to `lovrEventPush(EVENT_FOCUS, false)` when android tells us we are paused, no Lua code would be around to poll that event until after resuming...

bjornbytes:
The thing I want to do here is:

- lovr.timer.getTime returns wallclock time on android/oculus_mobile
- When the android activity is paused/resume, focus events are emitted.  Due to the structure of the android main loop and the Lua coroutine, both of these events will be consumed after the activity is resumed.  This is just going to be documented as a quirk/known limitation of the design.
[...]

PR #273, opened 2020-06-30
Android Support

bjornbytes:
This adds support for building Android APKs using tup and CMake, and also adds a vrapi headset backend.  There's a new Android section in the [Compiling Guide](https://lovr.org/docs/Compiling) that documents how to build an apk.

`lovr-oculus-mobile` should continue to work after these changes are merged.  However, the intent is to eventually remove lovr-oculus-mobile bridging in the codebase, since this PR effectively just moves all of the lovr-oculus-mobile code into the codebase.


bjornbytes:
The `vrapi` driver currently lacks support for haptics and hand tracking.

It has a few improvements too:

- `getOriginType` correctly returns head or floor depending on tracking setup.
- `getBoundsGeometry` returns guardian points.
- `isTracked` will return `false` when tracking is lost.
- On Oculus Go the back button is now mapped to the `menu` device button instead of always showing the dashboard.
- Adds support for the new "focus awareness" feature that doesn't pause the app when the dashboard is shown.  This also means that focus events work properly, which fixes #212.

bjornbytes:
`lovr.filesystem.getApplicationId` was removed.  It was an android-specific function that returned the Android package name, but now `lovr.filesystem.getIdentity` returns the package name on Android and can't be set via Lua code.

Additionally, the save directory on Android was changed to the external data path (/sdcard) instead of the internal data path, and no longer has the `LOVR/<identity>` suffix.

nevyn:
holy wow you did it!! amazing!!

Issue #275, opened 2020-07-10
LÖVR master branch does not compile


appgurueu:
LÖVR master branch does not compile on Ubuntu 20.04

Errors:
```
lovr/src/core/os_linux.c:57:17: error: dereferencing pointer to incomplete type ‘struct passwd’
   57 |     path = entry->pw_dir;
      |                 ^~
```
and
```
lovr/src/core/os_linux.c:100:3: error: unknown type name ‘ssize_t’; did you mean ‘size_t’?
  100 |   ssize_t length = readlink("/proc/self/exe", buffer, size - 1);
      |   ^~~~~~~
      |   size_t
```

PR #276, opened 2020-07-10
Fixed errors & warnings when compiling on linux

EliasFleckenstein03:
I added header files with #include to prevent the implicit declaration warnings and the dereferencing pointer to incomplete type error. I also fixed a typo in function size_t lovrPlatformGetExecutablePath(char* buffer, size_t size). See Issue #275.

### DIFF
--- a/src/core/os_linux.c
+++ b/src/core/os_linux.c
@@ -101,7 +101,7 @@ size_t lovrPlatformGetWorkingDirectory(char* buffer, size_t size) {
 }
 
 size_t lovrPlatformGetExecutablePath(char* buffer, size_t size) {
-  size_t length = readlink("/proc/self/exe", buffer, size - 1);
+  ssize_t length = readlink("/proc/self/exe", buffer, size - 1);
   if (length >= 0) {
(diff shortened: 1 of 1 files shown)


EliasFleckenstein03:
LOL I fix errors for the linux build and AppVeyor compiles it on Windows and failes? That is NOT my fault.

appgurueu:
Tested, closes my issue.

bjornbytes:
Thanks for fixing the build.  One small thing -- the man pages for `readlink` say that it does indeed return a `ssize_t` and that it will be -1 if an error occurs.  So I think the change to `size_t` should get reverted before merging, especially since it would mean `lovrPlatformGetExecutablePath` will always exceed (and cause a buffer overrun in case of error, since the -1 will implicitly get converted to SIZE_MAX and write the null terminator there).

Don't worry about appveyor -- it always runs on windows, which will have to be fixed separately.

EliasFleckenstein03:
Done (I made a typo in the commit message writing ssize_T instead of ssize_t)

PR #280, opened 2020-08-04
Fix Pico build

nevyn:
* lovrPlatformGetBundlePath was missing the root argument
* ANDROID_SDK can't be assumed to be the parent of the ndk folder, in case it's a side-by-side installation of the NDK. Instead, ANDROID_SDK should be provided with -D. (this then needs to be updated in the docs)
* One more thing we could mention in the docs that I ran into: Installing java with apt gave me an incompatible version. It worked better to just -DJAVA_HOME= to the java that comes with Android studio (/snap/android-studio/91/android-studio/jre on ubuntu).

It still crashes on launch for me, so there might be more commits to this PR soon :)


bjornbytes:
Thanks for the fixes!  Updated the docs as well.

PR #284, opened 2020-09-08
Fix definition of HOME key on Android

jmiskovic:
There are two HOME keys defined in keycodes.h, the previously used
doesn't work and is labeled with the following comment: "This key
is handled by the framework and is never delivered to applications."


bjornbytes:
Oops, thanks!

PR #287, opened 2020-09-24
Support for trimesh shape colliders

jmiskovic:
The trimesh collider support seems to be working properly now.

suzanne

Few still unsolved issues:

- arrays for vertices and indices are manually allocated in l_lovrWorldNewMeshCollider() and they are not freed in dGeomTriMeshDataDestroy(), destroying trimesh colliders will leak memory
- haven't made API to construct a shape and then add it to collider like other shapes have - would lead to substantial code duplication
- invalid index table winding (CW) leads to negative mass and ODE crashes - would be nice to detect and correct

Even with these issues the feature is already useful for most purposes.

### DIFF
--- a/src/api/l_physics_world.c
+++ b/src/api/l_physics_world.c
@@ -110,6 +110,44 @@ static int l_lovrWorldNewSphereCollider(lua_State* L) {
   return 1;
 }
 
+static int l_lovrWorldNewMeshCollider(lua_State* L) {
+  World* world = luax_checktype(L, 1, World);
+  lovrAssert(lua_istable(L, 2), "Vertices must be a table");
+  lovrAssert(lua_istable(L, 3), "Indices must be a table");
+  int vertexCount = luax_len(L, 2);
+  int indexCount  = luax_len(L, 3);
+  // TODO: this never gets deallocated
+  float    * vertices = malloc(sizeof(float) * vertexCount * 3);
+  unsigned * indices  = malloc(sizeof(unsigned) * indexCount);
+
+  for (int i = 0; i < vertexCount; i++) {
+    lua_rawgeti(L, 2, i + 1);
+    lovrAssert(lua_istable(L, -1), "Each verticle must be a table of coordinates");
+    lua_rawgeti(L, -1, 1); // x
+    vertices[i * 3 + 0] = luaL_optnumber(L, -1, 0.);
+    lua_pop(L, 1);
+    lua_rawgeti(L, -1, 2); // y
+    vertices[i * 3 + 1] = luaL_optnumber(L, -1, 0.);
+    lua_pop(L, 1);
+    lua_rawgeti(L, -1, 3); // z
+    vertices[i * 3 + 2] = luaL_optnumber(L, -1, 0.);
+    lua_pop(L, 2);
+  }
+  for (int i = 0; i < indexCount; i++) {
+    lua_rawgeti(L, 3, i + 1);
+    indices[i] = luaL_checkinteger(L, -1) - 1;
+  }
+  lua_pop(L, indexCount);
+  Collider* collider = lovrColliderCreate(world, 0,0,0);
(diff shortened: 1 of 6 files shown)


jmiskovic:
Lua code example.


[trimesh_test.zip](https://github.com/bjornbytes/lovr/files/5278632/trimesh_test.zip)

src/api/l_physics_world.c
```diff
@@ -110,6 +110,44 @@ static int l_lovrWorldNewSphereCollider(lua_State* L) {
...
+    lua_pop(L, 1);
+    lua_rawgeti(L, -1, 3); // z
+    vertices[i * 3 + 2] = luaL_optnumber(L, -1, 0.);
+    lua_pop(L, 2);
+  }
+  for (int i = 0; i < indexCount; i++) {
+    lua_rawgeti(L, 3, i + 1);
+    indices[i] = luaL_checkinteger(L, -1) - 1;
+  }
+  lua_pop(L, indexCount);
```
bjornbytes, line 140: I wonder if there is a risk of overflowing the Lua stack by saving the pops until the end.

src/modules/physics/physics.h
```diff
@@ -212,6 +214,10 @@ void lovrCylinderShapeSetRadius(CylinderShape* cylinder, float radius);
 float lovrCylinderShapeGetLength(CylinderShape* cylinder);
 void lovrCylinderShapeSetLength(CylinderShape* cylinder, float length);
 
+MeshShape* lovrMeshShapeInit(MeshShape* mesh, int vertexCount, float vertices[], int indexCount, dTriIndex indices[]);
```
bjornbytes, line 217: Going to try to use a non-ODE type for the index data, if possible.  Looks like the exact type of `dTriIndex` depends on various build options.

src/api/l_physics_world.c
```diff
@@ -110,6 +110,44 @@ static int l_lovrWorldNewSphereCollider(lua_State* L) {
   return 1;
 }
 
+static int l_lovrWorldNewMeshCollider(lua_State* L) {
+  World* world = luax_checktype(L, 1, World);
+  lovrAssert(lua_istable(L, 2), "Vertices must be a table");
+  lovrAssert(lua_istable(L, 3), "Indices must be a table");
+  int vertexCount = luax_len(L, 2);
+  int indexCount  = luax_len(L, 3);
+  // TODO: this never gets deallocated
```
bjornbytes, line 119: Was there anything challenging about this?  Does ODE need the vertex/index arrays to stay allocated during the lifetime of the trimesh?

bjornbytes:
Thanks, this is a great feature!

src/modules/physics/physics.h
jmiskovic: Is there a reason to avoid ODE types in adapter code that interfaces ODE?

My thinking was if someone were to change dTRIMESH_16BIT_INDICES or dTRIMESH_GIMPACT flags that affect dTriIndex size, it would be quite nasty to track down the issue.

src/api/l_physics_world.c
jmiskovic: ODE crashes if I those arrays are deallocated while rigid body is still alive. ODE also won't free those allocations for us.

Memory should be freed in `dGeomTriMeshDataDestroy()`. To know pointer values at that time there are two options:

1. Make `MeshShape` struct special and include pointers there. Currently all shapes use same struct so this is a bit awkward.
2. Add a separate map structure that would use shape->id as key and pointers to vertices and indices as value. Still need to figure out how to track this per-physics-world, as different worlds can have same id for different collider.
3. You also mentioned the need for `luax_readtriangles`. Not sure what memory ownership model would be best there.

If you have peference for how to solve this, I'd love to fix this loose end.

src/api/l_physics_world.c
jmiskovic: I haven't thought of that. Strange the overflow never happened and I was testing with quite large trimeshes (icosphere with 5 subdivisions). It's an easy fix, I'll make the change.

PR #289, opened 2020-10-01
Allow lovr.headset.init to fail

nevyn:
with a pcall, like the requires above. so if there
is no matching headset driver, the module is just
turned off.


bjornbytes:
Thanks!

PR #294, opened 2020-10-21
lovr.headset.getSkeleton("local")

nevyn:
API to get the hand's skeleton poses in each node's local coordinate system, so that they can be applied/animated onto a hand model with the same node skeleton hierarchy as the hand tracking data.

This is WIP:

- [ ] BUG For some reason all the poses seem to be at 180 degrees rotation when my hand is at neutral pose. am I missing something oculus specific, or do I have a bug, or what's going on...
- [ ] FEATURE Change C API for getSkeleton to take coordinate space param
- [ ] FEATURE Change Lua API for getSkeleton to take coordinate space param
- [ ] Figure out what to do for the extra coordinate space param for all the headsets I can't test myself...


bjornbytes:
IIRC we solved this in a different way or determined it would be challenging to support, right?  Please reopen if I'm mistaken

nevyn:
@bjornbytes Well, I tried another approach but I'm still finding my bearings. I think we can skip this PR yeah; but I'll ping you if I want to resurrect it.

PR #297, opened 2020-10-24
OpenXR: Fixes for implementations that are more strict with validation

Wallbraker:
Monado requires that the .type field is always set, and that poses are completely valid including setting `w` to 1.0f.


bjornbytes:
Awesome, ran into some of these when testing on Android runtimes but it looks like I missed some

PR #298, opened 2020-10-24
CMake: Add LOVR_SYSTEM_OPENXR;

bjornbytes:
Add a CMake option to search for the system-installed OpenXR loader using pkg-config.


Wallbraker:
Thank you for taking care of this so quickly. Works like a charm on Ubuntu 20.04. So this is both `Tested-by` and `Reviewed-by` `jakob@collabora.com`.

PR #299, opened 2020-10-24
Linux OpenXR support

Wallbraker:
This allows me to at least run on the open source OpenXR runtime [Monado](monado.freedesktop.org) on Linux. Marked it as WIP as I don't really know if this is the right approach here. Also have two extra fixes on the side that isn't mine. But this allows anybody looking for OpenXR support on Linux to not have to go looking for said fixes.


bjornbytes:
Super excited to get this merged!  I'd like to make a change to the way things are structured before merging in order to maintain the existing design.

Currently core/os doesn't rely on the OpenXR headers.  Instead, it "privately" implements accessors to platform-specific handles that headset_openxr.c (and other components) then uses to create the graphics plugin.

I think this PR could use the same approach, where `lovrPlatformGetEGL*` and `lovrPlatformGetX11*` functions could get added to os_glfw.h.  Then, the OpenXR driver could define these and use them to make its structs.

I don't really see a problem with the inverted design that this PR uses where core/os is able to construct structs for other components that need internal handles, but for now I'd like to keep things consistent and just expose the handles.

src/core/os_glfw.h
```diff
@@ -204,12 +204,18 @@ bool lovrPlatformCreateWindow(const WindowFlags* flags) {
...
+#ifdef LOVR_LINUX_EGL
+  glfwWindowHint(GLFW_CONTEXT_CREATION_API, GLFW_EGL_CONTEXT_API);
+#endif
   glfwWindowHint(GLFW_CONTEXT_VERSION_MAJOR, 3);
   glfwWindowHint(GLFW_CONTEXT_VERSION_MINOR, 3);
   glfwWindowHint(GLFW_OPENGL_PROFILE, GLFW_OPENGL_CORE_PROFILE);
   glfwWindowHint(GLFW_OPENGL_FORWARD_COMPAT, GLFW_TRUE);
   glfwWindowHint(GLFW_OPENGL_DEBUG_CONTEXT, flags->debug);
+#ifndef LOVR_LINUX_EGL
   glfwWindowHint(GLFW_CONTEXT_NO_ERROR, !flags->debug);
```
bjornbytes, line 217: Curious, was this hint unavailable or otherwise not working with EGL?

Wallbraker: It failed to create the context if this was enabled.

```txt
Warning: Could not load module "graphics": EGL: Failed to create context: An unrecognized attribute or attribute value was passed in the attribute list
Error:
main.lua:8: attempt to index field 'graphics' (a nil value)
```

Wallbraker:
> Super excited to get this merged! I'd like to make a change to the way things are structured before merging in order to maintain the existing design.
> 
> Currently core/os doesn't rely on the OpenXR headers. Instead, it "privately" implements accessors to platform-specific handles that headset_openxr.c (and other components) then uses to create the graphics plugin.
> 
> I think this PR could use the same approach, where `lovrPlatformGetEGL*` and `lovrPlatformGetX11*` functions could get added to os_glfw.h. Then, the OpenXR driver could define these and use them to make its structs.
> 
> I don't really see a problem with the inverted design that this PR uses where core/os is able to construct structs for other components that need internal handles, but for now I'd like to keep things consistent and just expose the handles.

I did this design because I was not able able to include X11 headers and `os.h`. That is in order to get access to the `glfw` functions I need to `#define GLFW_EXPOSE_NATIVE_X11` before I `#include <GLFW/glfw3native.h>`. But if I do that your struct `KeyCode` conflicts with a X11 `KeyCode` struct.

```txt
In file included from /usr/include/X11/Xlib.h:44,
                 from /home/jakob/XR/lovr/deps/glfw/include/GLFW/glfw3native.h:101,
                 from /home/jakob/XR/lovr/src/core/os_glfw.h:11,
                 from /home/jakob/XR/lovr/src/core/os_linux.c:9:
/usr/include/X11/X.h:108:23: error: conflicting types for ‘KeyCode’
  108 | typedef unsigned char KeyCode;
      |                       ^~~~~~~
In file included from /home/jakob/XR/lovr/src/core/os_linux.c:1:
/home/jakob/XR/lovr/src/core/os.h:125:3: note: previous declaration of ‘KeyCode’ was here
  125 | } KeyCode;
      |   ^~~~~~~
In file included from /home/jakob/XR/lovr/src/core/os_linux.c:9:
/home/jakob/XR/lovr/src/core/os_glfw.h:342:6: error: conflicting types for ‘lovrPlatformIsKeyDown’
  342 | bool lovrPlatformIsKeyDown(KeyCode key) {
      |      ^~~~~~~~~~~~~~~~~~~~~
In file included from /home/jakob/XR/lovr/src/core/os_linux.c:1:
/home/jakob/XR/lovr/src/core/os.h:168:6: note: previous declaration of ‘lovrPlatformIsKeyDown’ was here
  168 | bool lovrPlatformIsKeyDown(KeyCode key);
      |      ^~~~~~~~~~~~~~~~~~~~~
```

bjornbytes:
Ack, sorry about that.  I just pushed a commit to `dev` that renames `KeyCode` to `KeyboardKey` which should unblock that, though proper prefixing should be a better long-term change.

Wallbraker:
Thank you, and I agree.

Wallbraker:
This PR is now updated and ready to land, it adds support for OpenXR on Linux using both EGL and X11.

bjornbytes:
Thanks!  This is very exciting

Issue #307, opened 2020-10-28
Allow password-less keystores


bjornbytes:
It is possible to have keystores that don't have a password.  If the `ANDROID_KEYSTORE_PASS` variable is not set, the `apksigner` invocation breaks because there is no argument passed to `--ks-pass`.  In this case, we should omit the `--ks-pass` option or use an empty string if that's valid.

PR #309, opened 2020-10-28
Add a check so apksigner works even without ANDROID_KEYSTORE_PASS

mcclure:
This is done by omitting the --ks-pass argument completely


mcclure:
Addresses #307. I don't know if it does it right, but it does it.

PR #319, opened 2020-10-30
on lodr restart, make sure lovr.headset is nil'd if none is available

nevyn:
In boot.lua, it assumes that lovr.headset.init will assert if no driver
is available. This was previously only true on the first call to it,
since after it's initialized, it'll just return early and won't assert.
This will later crash since your lua code will now see a lovr.headset
being available, but calling anything in it will crash since
lovrHeadsetDisplayDriver is NULL

### DIFF
--- a/src/api/l_headset.c
+++ b/src/api/l_headset.c
@@ -142,9 +142,8 @@ static int l_lovrHeadsetInit(lua_State* L) {
     lua_pop(L, 1);
   }
 
-  if (lovrHeadsetInit(drivers, driverCount, supersample, offset, msaa)) {
-    luax_atexit(L, lovrHeadsetDestroy);
-  }
+  luax_atexit(L, lovrHeadsetDestroy); // Always make sure the headset module gets cleaned up
+  lovrHeadsetInit(drivers, driverCount, supersample, offset, msaa);
 
   lua_pop(L, 2);
   return 0;


bjornbytes:
This error message seems weird to me.

If my Lua code does

```lua
lovr.headset.init()
lovr.headset.init()
```

Then won't this print an error saying that "no driver is available" even if I have a driver?

One other thing to keep in mind is that other threads will see whatever behavior is here when they require lovr.headset.

nevyn:
@bjornbytes No, it's an assert, not a print, so the error will only print if `initialized` is true AND `lovrHeadsetDisplayDriver` is NULL (i e there really is no headset driver available).

Is the headset module supposed to be thread safe?? it's sharing global state :S though I can't see how this patch affects thread safetiess;this is no more or less threadsafe than it was before, is it?

bjornbytes:
Ok I think I understand what is going on now.

My test project had `t.headset.drivers` set to an empty table and called `lovr.event.restart`, which causes a crash like you describe.

I think the root of the issue is that `lovrHeadsetDestroy` is not getting called if `lovrHeadsetInit` throws an error.

This is because the Lua bindings for `lovr.headset.init` do something like this:

```c
  if (lovrHeadsetInit()) {
    luax_atexit(L, lovrHeadsetDestroy);
  }
```

So `luax_atexit` doesn't get a chance to register its destructor.

A possible (tested) fix is to change it to something like this:

```c
luax_atexit(L, lovrHeadsetDestroy); // Always make sure the headset module gets cleaned up
lovrHeadsetInit();
```

This will ensure that after a restart, the headset module is in a clean state and the regular assert will always get thrown.  Note that this is the same kind of approach that the graphics module takes in `l_lovrGraphicsCreateWindow`.

Thoughts on this?  To me it looks like changing the atexit will fix this issue as well as any other issues that could arise from not destroying the headset module properly, but it's possible that your change here is better for other reasons.

re: thread safety, I honestly don't know how thread safe the headset module is right now, so maybe my earlier comment about threads was off the mark.  I do think that "`lovr.headset.init` can only be called from one thread" is true right now.

nevyn:
Ahh alright. Yep, I didn't know whether headset module was supposed to be initialized once globally or once per restart. If the latter was intended, your solution seems much better. I can try it in my env and see if it helps, and if so, push that as a fix instead!

nevyn:
@bjornbytes I've confirmed your proposal works for me too, and I've thus amended this PR to match. I like this approach a lot better, thank you for taking the time!

bjornbytes:
Sweet!

PR #323, opened 2020-11-06
don't free window icon before it's used

nevyn:
It seems to me like that lovrRelease will delete textureData->blob immediately,
which means the windowing system later can't use it because it's already freed.
There's already a free on line 378 which looks more correct.

Also, icon appears flipped if 'flipped' is set to true here on Linux. Is GLFW
inconsistent between linux and windows, or should it indeed be false?


bjornbytes:
Looks good to me, I can't imagine that flipping would be required because that's an OpenGL thing

nevyn:
DAMN you're fast :sweat_smile: thanks!!

PR #327, opened 2020-11-12
Create empty curve with specified number of points

jmiskovic:
This `lovr.math.newCurve(n)` variant is already described in API
documentation, now it is implemented.

### DIFF
--- a/src/api/l_math.c
+++ b/src/api/l_math.c
@@ -110,6 +110,13 @@ static int l_lovrMathNewCurve(lua_State* L) {
       i += 3 + components;
       lua_pop(L, 3);
     }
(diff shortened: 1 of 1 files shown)


bjornbytes:
It's interesting how the variant was removed [here](https://github.com/bjornbytes/lovr/commit/eb1e2572090c10e29aed8a57cfd45309e00cb90b#diff-d5d4ecefff689b2a37caeef3c5bb2440ddb891487216cc41c0ddd372198c4c08L84-L91)

jmiskovic:
OK that removed code is cleaner. I'll resubmit the PU.

jmiskovic:
The only bit I could reuse is the check that input param is number type. The `lovrCurveCreate()` no longer accepts number of points and it seems better that way.

PR #329, opened 2020-11-12
Change default damping threshold to zero

jmiskovic:
Zero as default makes more sense. Colliders can come to full stop which
allows them to go to sleep for CPU optimization. Also in zero gravity
colliders crawling through air with 0.01 velocity are infuriating.


bjornbytes:
This seems fine, I'm sure the bindings only used .01 because that's what ODE did.  I'm sad it was missing from the docs.

Could this change go into the `dev` branch though?  I can change the target branch here but yours will need to be rebased.  Alternatively I can do the merge if that's easier.

bjornbytes:
Nevermind, I was able to rebase dev so this change can merge in cleanly.

PR #331, opened 2020-11-16
Fix NaN return from mat4 unpacking

jmiskovic:
The caluclated angle is forced to PI if cos(angle) falls outside the
[-1, +1] range due to floating point precision error.


bjornbytes:
This was nasty!

PR #333, opened 2020-11-18
Do not try to create hand tracking devices if handTracking is not sup…

Wallbraker:
…ported


bjornbytes:
I think here I was relying on the failure code of CreateHandTracker for some reason, but it looks like this change would avoid some overhead.  I'm going to test it on Ultraleap's layer before merging just to be sure that still works.  Thanks!

bjornbytes:
Tested on Ultraleap 👍

PR #339, opened 2020-11-25
getting my feet wet with miniaudio branch

nevyn:
Just getting a feel for things, picking some low hanging fruit. Making this PR for visiblity and let you comment if I stray too far; not asking for a merge (though do feel free to cherry/merge/whatev if you want).

- [x] Change setSpatial to be a constructor option
- [x] Get audio capture working
- [x] Explicit start/stop capture
- [x] Don't record automatically
- [x] Audio permissions -- implicit when starting recording
- [x] See if we can get rid of appcompat
- [x] Bump miniaudio
- [ ] Setting capture device
- [ ] Audio streams


bjornbytes:
I was planning on keeping the `spatial` flag as a static option specified upfront in `lovr.audio.newSource`.  I only did it that way because a lot of things are simplified if it's static (e.g. don't need to add a mutex to protect against the spatialness changing while the audio callback is running, don't need to recreate the converters, don't need to worry about the spatializer needing to recreate objects it needs for the source).  I couldn't think of use cases where you'd want to change this property, but it's possible I glossed over something important...

Potentially related: SteamAudio has a thing where it can blend a source between spatialized and unspatialized

> Added support for spatial blend in the binaural effect, which smoothly blends between fully-spatialized point source and unspatialized stereo input audio.

nevyn:
yeah I realized after I had already written the code :S and I have a hard time figuring out a use case for changing the spatialness. I'll change it to constructor time. thanks for the feedback!

nevyn:
Audio capture is working now! :heart_eyes:

nevyn:
@bjornbytes Working on capture on android. A few notes:

* Needed appcompat for the android permissions api, and without gradle I resorted to bundling the jar in the repo. I don't know if there's a prettier solution.
* I'm not super happy about the structure in audio.c. It feels weird to fail initialization of playback because capture failed. I'm also not sure why lovrAudioInit takes a config to start with; it'll always fail for capture on first boot because permissions won't be accepted yet, and then you'll get no playback?
* I tried to adapt lovrAudioStart to work with the existing structure, but that means state.config is now mutable state. I feel like the code might not be conforming to what you imagined for it, so I'm letting you know so you can suggest course changes.
* I added permissions to the Oculus implementation now, but the three Activities should probably have a common base impl instead.
* I'm starting to have changed things a lot, so perhaps this should be merged into miniaudio-new soon?

Weirdness: This code:

```
    local captured = lovr.audio.getCaptureDuration("samples")
    if captured > 0 then
        local data = lovr.audio.capture()
        lovr.filesystem.append("audio.pcm", data:getBlob():getString())
```

works fine on desktop, but on Quest I get `main.lua:37: calling 'getBlob' on bad self (TextureData expected, got userdata)`. super weird. any ideas? otherwise I'll just dig in printf debugging... I'm using [this branch of this sample code](https://github.com/alloverse/lovr-sample-audio/tree/miniaudio).

Gonna attack permissions-on-demand now.

bjornbytes:
I'll review this soon!

Would like to aggressively try to find a way to avoid jars/java, but need to learn more about appcompat.

I think you raise a good point that the audio module needs to be more tolerant of failed devices.

src/api/l_audio.c
```diff
@@ -24,8 +24,9 @@ static int l_lovrAudioReset(lua_State* L) {
 
 static int l_lovrAudioStart(lua_State* L) {
   AudioType type = luax_checkenum(L, 1, AudioType, "playback");
-  lovrAudioStart(type);
-  return 0;
+  bool ret = lovrAudioStart(type);
+  lua_pushboolean(L, ret);
```
bjornbytes, line 28: I'd prefer a name with a little more meaning like `success` / `started` or something (or just to have it inlined into the pushboolean call to avoid naming altogether).

src/api/l_audio.c
```diff
@@ -48,15 +49,16 @@ static int l_lovrAudioSetVolume(lua_State* L) {
 static int l_lovrAudioNewSource(lua_State* L) {
   Source* source = NULL;
   SoundData* soundData = luax_totype(L, 1, SoundData);
+  bool spatial = lua_isboolean(L, 2) ? lua_toboolean(L, 2) : true;
```
bjornbytes, line 52: Maybe this should be in an options table.  It would make it more clear when the spatial property is being set:

```lua
lovr.audio.newSource(sound, false) -- Hmm, what is that boolean for
lovr.audio.newSource(sound, { spatial = false }) -- More clear
```

and make it easier to add more configuration to Source objects without making breaking changes to the API.

I think it's okay to still have the bool in the C API, some sort of SourceInfo struct could be added there later if needed.

src/api/l_audio.c
```diff
@@ -65,21 +67,88 @@ static int l_lovrAudioNewSource(lua_State* L) {
...
+  luax_pushtype(L, SoundData, soundData);
+  lovrRelease(SoundData, soundData);
+  return 1;
+}
+
+static int l_lovrAudioSetCaptureDevice(lua_State *L) {
+  //
+
+  return 0;
+}
```
bjornbytes, line 117: I was planning on not exposing device iteration and always using the default device, unless people started asking for it. 
 For VR games and utilities it seems most common to use the single microphone built into the headset.

If this is added though, maybe it could be an optional table passed to `start` indicating desired device and settings for stuff like format/sample rate/buffer size preferences?

While I was still planning on doing device iteration I had a function like this:

```lua
for id, name, type, minChannels, maxChannels in lovr.audio.devices() do

end
```

src/api/l_audio_source.c
```diff
@@ -51,6 +51,22 @@ static int l_lovrSourceSetVolume(lua_State* L) {
   return 0;
 }
 
+static int l_lovrSourceGetSpatial(lua_State* L) {
```
bjornbytes, line 54: Lua API functions use the `is` prefix instead of `get` when exposing a boolean (see `isLooping` and `isPlaying`).

src/lib/miniaudio/miniaudio.h
```diff
@@ -19845,7 +19845,7 @@ static ma_result ma_device_init__pulse(ma_context* pContext, const ma_device_con
             goto on_error3;
         }
 
-        streamFlags = MA_PA_STREAM_START_CORKED | MA_PA_STREAM_FIX_FORMAT | MA_PA_STREAM_FIX_RATE | MA_PA_STREAM_FIX_CHANNELS;
+        streamFlags = MA_PA_STREAM_START_CORKED | MA_PA_STREAM_ADJUST_LATENCY | MA_PA_STREAM_FIX_FORMAT | MA_PA_STREAM_FIX_RATE | MA_PA_STREAM_FIX_CHANNELS;
```
bjornbytes, line 19848: The version of miniaudio on this branch is pretty out of date, and it looks like newer versions include this change, so maybe we should just upgrade?

src/modules/audio/audio.c
```diff
@@ -30,21 +36,18 @@ struct Source {
   bool playing;
   bool looping;
   bool spatial;
+  float pose[16];
```
bjornbytes, line 39: Usually the `pose` term is used for an 8-float position + quaternion, `transform` would be a more consistent name for a 16-float mat4 transform.

src/modules/audio/audio.c
```diff
@@ -30,21 +36,18 @@ struct Source {
...
-
 static struct {
   bool initialized;
   ma_context context;
-  AudioConfig config[2];
-  ma_device devices[2];
-  ma_mutex locks[2];
+  AudioConfig config[AUDIO_TYPE_COUNT];
+  ma_device devices[AUDIO_TYPE_COUNT];
+  ma_mutex locks[AUDIO_TYPE_COUNT];
```
bjornbytes, line 48: Using a lockfree ringbuffer for capture is really cool, maybe we could just have a single `state.lock` now, unless we anticipate future concurrency issues with audio capture.

src/modules/audio/audio.c
```diff
@@ -30,21 +36,18 @@ struct Source {
   bool playing;
   bool looping;
   bool spatial;
+  float pose[16];
+  int output_channel_count;
```
bjornbytes, line 40: Style: this struct field (and capture_ringbuffer) should be camelCase

bjornbytes: Also, this value depends entirely on `spatial` and can be computed from it efficiently.  Storing it separately doesn't feel necessary and there is a risk that they accidentally drift (e.g. we add the ability to change spatialness and now have to remember to keep the output channel count in sync).  This is pretty minor though.  It could also be the case that in the future we *do* want to separate them for a valid reason.

src/modules/audio/audio.c
```diff
@@ -67,13 +70,19 @@ static bool mix(Source* source, float* output, uint32_t count) {
...
 
     uint64_t framesIn = source->sound->read(source->sound, source->offset, chunk, raw);
-    uint64_t framesOut = sizeof(aux) / (sizeof(float) * (2 >> source->spatial));
+    uint64_t framesOut = sizeof(aux) / (sizeof(float) * source->output_channel_count);
 
     ma_data_converter_process_pcm_frames(source->converter, raw, &framesIn, aux, &framesOut);
 
-    memcpy(mix, aux, framesOut * 2 * sizeof(float));
+    if(source->spatial) {
+      if(state.spatializer) {
```
bjornbytes, line 78: If the source is spatial but there isn't a spatializer, `mix` never gets written to, right?

Hm, it may be that the `spatial` flag needs to get forced to false for Sources if no spatializer successfully loads at initialization time.

bjornbytes: Style: ifs should have spaces after them

src/modules/audio/audio.c
```diff
@@ -108,32 +117,37 @@ static void onPlayback(ma_device* device, void* output, const void* _, uint32_t
...
 static void onCapture(ma_device* device, void* output, const void* input, uint32_t frames) {
-  ma_mutex_lock(&state.locks[1]);
-  ma_mutex_unlock(&state.locks[1]);
+  // note: ma_pcm_rb is lockless
+  void *store;
+  while(frames > 0) {
+    uint32_t available_frames = frames;
+    ma_result acquire_status = ma_pcm_rb_acquire_write(&state.capture_ringbuffer, &available_frames, &store);
+    if(acquire_status != MA_SUCCESS) {
+      fprintf(stderr, "Dropping mic audio, failed to acquire ring buffer: %d\n", acquire_status);
```
bjornbytes, line 130: Would like to try to use `lovrAssert` or `lovrLog` here if possible.  Calling those on threads isn't possible, but it may be that some sort of `captureBroken` flag could be set in state that `lovr.audio.capture` checks?  Not sure.

src/modules/audio/audio.c
```diff
@@ -108,32 +117,37 @@ static void onPlayback(ma_device* device, void* output, const void* _, uint32_t
...
+    }
+    memcpy(store, input, available_frames*sizeof(float)*CAPTURE_CHANNELS);
+    ma_result commit_status = ma_pcm_rb_commit_write(&state.capture_ringbuffer, available_frames, store);
+    if(commit_status != MA_SUCCESS) {
+      fprintf(stderr, "Dropping mic audio, failed to commit ring buffer: %d\n", acquire_status);
+      return;
+    }
+    frames -= available_frames;
+    input += available_frames*sizeof(float)*CAPTURE_CHANNELS;
+  }
```
bjornbytes, line 144: Style notes: spaces before ifs/whiles, camelCase variables, spaces around binary operators

bjornbytes (commented):
Here are my initial comments!  Don't worry about stuff marked "Style" if it feels trivial, it can always get cleaned up later.

src/resources/boot.lua
```diff
@@ -190,6 +190,11 @@ function lovr.run()
     if lovr.headset then
       lovr.headset.update(dt)
     end
+    if lovr.audio then
+      if lovr.headset then
```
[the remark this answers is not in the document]
nevyn: nope, I just copy-pasted from old boot! I'll fix.

src/api/l_audio.c
nevyn: that's a MUCH better API! I also prefer keyword-style APIs and clear parameters! (I'm an old ObjC programmer, after all :P) will fix!

src/api/l_audio.c
nevyn: Alright. I really really need device iteration for Alloverse, so I'll need it in our fork anyway. I'll submit a suggestion and you can take it out in mainline if you want. (though in my experience, automatic device selection can work really poorly in PCVR titles outside of Oculus headsets, and taking off headset to switch recording device can be a hassle).

that's an interesting API! I can adapt something similar :)

src/modules/audio/audio.c
nevyn: I made it a separate computed value to be explicit. I don't usually like duplicate state but this felt better than hard-coding the assumption between spatial and channels into parts of the code. Could make a little static inline helper function to make it explicit in a single place (should have same performance)

src/modules/audio/audio.c
nevyn: Good call. Spatial forced to false if no spatializer is available.

Though, in my implementation there should always be a spatializer. If no third party one is available, the dummy one should work at least. There's a bug so that it's null in some cases but we can fix that; and my math in it is wrong so audio orientation is wrong when you turn; but with those fixes spatial audio will always be available even though it's not oculus/valve quality.

nevyn:
@bjornbytes Haven't addressed your above concerns yet but I agree with all of them and will fix when I get a chance! Wanted to get to permissions-on-demand first. It's now implemented, and

nevyn:
@bjornbytes I agree with all the proposed changes above and I'll get to them asap, but haven't yet because I wanted to get permissions-on-demand down first.

And now it works! It requests mic permissions when you start audio capture and not earlier; and it re-tries audio capture once permissions have been granted.

As per @mcclure's request, I've attempted to make it a bit generic. There is no Lua-facing API for requesting permission, but there is an os.h `lovrPlatformRequestAudioCapture`.

There is also a `lovrPlatformOnPermissionEvent` which hooks into event.c to publish a new "permission" event, with two args: permission enum, and bool granted.

The default implementation of lovr.permission() checks if permission is audio_capture, and if it is and that permission is granted, it tries to re-start audio capture. However, a lovr user can override to provide custom behavior. Perhaps `lovrPlatformRequestAudioCapture` should be renamed to `lovrPlatformRequestPermission()` (but it's not public API so whatev), but other than that I feel like it's pretty nicely generic and in line with other Lovr APIs. What do you two think?

src/modules/audio/audio.c
bjornbytes: Ooh yeah, if we can guarantee that there's a spatializer (even a dummy/null spatializer that does nothing), then we don't even need to worry about this.  Saves on branching too.

src/api/l_audio.c
bjornbytes: Ok, it's helpful to know that you need multiple devices, that makes it a higher priority.

nevyn: I don't need to connect to multiple devices at once, but I need to be able to switch between them (so I can present a mic selector in my app), just to be clear :)

src/modules/audio/audio.c
nevyn: Done my best to address these but they're sorta hard-coded into my hands so more might've snuck in without me noticing :S

src/api/l_audio.c
nevyn: @bjornbytes would `for _, deviceDescription in ipairs(lovr.audio.devices()) ` with `deviceDescription` being a table with name, type, etc be ok, worse, better?

bjornbytes: That's probably better.  If it returns a table it ought to be called `getDevices` though.

src/modules/audio/audio.c
bjornbytes: Also, what's the story with the dummy spatializer?  Do you think it's good enough to be used as the default if one of the fancier spatializers isn't loaded?  Or was it just for testing here

bjornbytes:
The permissions system looks great.

I'm curious how the new macOS permissions stuff works, and if it could be implemented under this API.  Seems pretty likely?

src/api/l_event.c
```diff
@@ -110,6 +111,12 @@ StringEntry lovrKeyboardKey[] = {
   { 0 }
 };
 
+StringEntry lovrPermission[] = {
+  [AUDIO_CAPTURE_PERMISSION] = ENTRY("audio_capture"),
```
bjornbytes, line 115: It is kind of silly, but the "lovely" enum style is to string multiple words together like `audiocapture`

src/resources/boot.lua
```diff
@@ -172,6 +172,12 @@ function lovr.boot()
   return lovr.run()
 end
 
+function lovr.permission(permission, granted)
+  if permission == "audio_capture" and granted then
+    lovr.audio.start("capture")
+  end
```
bjornbytes, line 178: Should this be implemented as default behavior?  Will other permissions have default behavior?  Hmm, it's interesting to think about.

src/api/l_event.c
nevyn: oh okay :sweat_smile:

src/resources/boot.lua
nevyn: I think this is the least surprising implementation for a game engine; you ask for audio, user says yes, you get audio. I think it should either work this way, or have an explicit check-for-permission API so dev can check it up front...

nevyn:
> I'm curious how the new macOS permissions stuff works, and if it could be implemented under this API. Seems pretty likely?

Yes, I've ugly-hacked it in in Alloverse and can confirm that it will work fine under this API :)

nevyn:
@bjornbytes FYI, it seems to work fine without AppCompat :) I've pulled the jars out and removed references to those libs.

bjornbytes:
Oh, nice.  Is it okay if I merge this into miniaudio-new now?

nevyn:
@bjornbytes yes I think so! I also bumped miniaudio to latest, seems to be working fine.

src/lib/miniaudio/miniaudio.h
nevyn: done!

PR #340, opened 2020-11-26
filesystem.append didn't append in Unix

nevyn:
At least on my Linux system it'd overwrite the contents of the file each time.


bjornbytes:
Thanks!

PR #346, opened 2020-12-09
Miniaudio streams

nevyn:
WIP as usual, just showing my work

- [x] Lua API for append
- [x] Find and fix all the bugs
- [x] Remove asserts from `lovrSoundDataReadRing` because it's not on the main thread
- [x] `SoundData:getDuration()` so we can know how much is buffered

Note: I had to add a struct tag in miniaudio.h or otherwise I'd have to pollute the soundData.h header with miniaudio.h :S just FYI...


nevyn:
@bjornbytes I think this is done. I'll branch off of this work on setting sound devices...

src/modules/data/soundData.h
```diff
@@ -17,17 +19,24 @@ typedef struct SoundData {
   SoundDataReader* read;
   void* decoder;
   struct Blob* blob;
-  struct Blob** queue;
-  uint32_t buffers;
-  uint32_t head;
-  uint32_t tail;
+  ma_pcm_rb *ring;
```
bjornbytes, line 22: Could this be a `void*` (like decoder) to avoid patching miniaudio and 'exposing' an internal type?

src/modules/data/soundData.c
```diff
@@ -119,9 +143,66 @@ SoundData* lovrSoundDataCreateFromFile(struct Blob* blob, bool decode) {
...
+
+size_t lovrSoundDataStreamAppendSound(SoundData *dest, SoundData *src) {
+  lovrAssert(dest->channels == src->channels && dest->sampleRate == src->sampleRate && dest->format == src->format, "Source and destination SoundData formats must match");
+  lovrAssert(src->blob && src->read == lovrSoundDataReadRaw, "Source SoundData must have static PCM data and not be a stream");
+  return lovrSoundDataStreamAppendBlob(dest, src->blob);
+}
+
+void lovrSoundDataSetSample(SoundData* soundData, size_t index, float value) {
+  size_t byteIndex = index * bytesPerAudioFrame(soundData->channels, soundData->format);
+  lovrAssert(byteIndex < soundData->blob->size, "Sample index out of range");
```
bjornbytes, line 181: Any reason not to do `index < soundData->frames` here?

src/modules/audio/audio.c
```diff
@@ -353,6 +346,7 @@ bool lovrSourceIsLooping(Source* source) {
 }
 
 void lovrSourceSetLooping(Source* source, bool loop) {
+  lovrAssert(loop == false || lovrSoundDataIsStream(source->sound) == false, "Can't loop streams");
```
bjornbytes, line 349: Would it be useful to have a looping audio stream if that means that it constantly plays whatever is pushed to it instead of needing to be :played all the time?  Maybe a special case isn't needed here

src/api/l_data_soundData.c
```diff
@@ -9,8 +9,32 @@ static int l_lovrSoundDataGetBlob(lua_State* L) {
   return 1;
 }
 
+static int l_lovrSoundDataAppend(lua_State* L) {
+  SoundData* soundData = luax_checktype(L, 1, SoundData);
+  Blob* blob = luax_totype(L, 2, Blob);
+  SoundData* sound = luax_totype(L, 2, SoundData);
+  lovrAssert(blob || sound, "Invalid blob appended");
+  size_t appendedSamples = false;
```
bjornbytes, line 17: Wait is this a number or a boolean?

src/api/l_data_soundData.c
```diff
@@ -9,8 +9,32 @@ static int l_lovrSoundDataGetBlob(lua_State* L) {
   return 1;
 }
 
+static int l_lovrSoundDataAppend(lua_State* L) {
+  SoundData* soundData = luax_checktype(L, 1, SoundData);
+  Blob* blob = luax_totype(L, 2, Blob);
+  SoundData* sound = luax_totype(L, 2, SoundData);
+  lovrAssert(blob || sound, "Invalid blob appended");
```
bjornbytes, line 16: Could use `luaL_argerror` here to get a conventional "Bad argument N ..." error message.  It could even get added as an else down below so an extra if isn't needed here.

src/api/l_data.c
```diff
@@ -70,10 +76,9 @@ static int l_lovrDataNewSoundData(lua_State* L) {
     uint64_t frames = luaL_checkinteger(L, 1);
     uint32_t channels = luaL_optinteger(L, 2, 2);
     uint32_t sampleRate = luaL_optinteger(L, 3, 44100);
-    uint32_t format = luaL_optinteger(L, 4, 16);
+    SampleFormat format = luax_checkenum(L, 4, SampleFormat, "i16");
     Blob* blob = luax_totype(L, 5, Blob);
```
bjornbytes, line 80: This is probably just my tendency to condense everything too much, but if the Blob argument here was thought of as "contents", then then contents for your SoundData could be:

- `Blob`: You have data you want to use as the contents
- `nil`: You want empty contents initialized for you
- `'stream'` or `'append'`: You want to stream the contents to the SoundData using `:append`
- ...potentially others?

That would avoid the need for an extra newSoundDataStream function and just puts the word 'stream' at the end of this one :)

src/modules/audio/audio.c
```diff
@@ -290,13 +281,14 @@ static void _lovrSourceAssignConverter(Source *source) {
...
-    config.formatIn = formats[source->sound->format];
-    config.formatOut = formats[OUTPUT_FORMAT];
+    config.formatIn = miniAudioFormatFromLovr[source->sound->format];
+    config.formatOut = miniAudioFormatFromLovr[OUTPUT_FORMAT];
     config.channelsIn = source->sound->channels;
     config.channelsOut = outputChannelCountForSource(source);
     config.sampleRateIn = source->sound->sampleRate;
     config.sampleRateOut = LOVR_AUDIO_SAMPLE_RATE;
-    arr_expand(&state.converters, 1);
+    // can't use arr_expand because that will destroy the internal pointers inside the converter
```
bjornbytes, line 290: Woops, sorry about that.  Maybe Sources could store the index of their converter?  It would take up less space and avoid having this limit

src/modules/data/soundData.c
```diff
@@ -3,18 +3,16 @@
 #include "core/util.h"
 #include "core/ref.h"
 #include "lib/stb/stb_vorbis.h"
+#include "lib/miniaudio/miniaudio.h"
+#include "audio/audio_internal.h"
```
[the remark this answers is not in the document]
nevyn: Okay, sure I can do that (inverting relationship so audio depends on audio instead of other way around).

Heh I understand the sentinment, but yeah the ringbuffer in miniaudio is amazing and it'd be silly not to use it imho :S it's even threadsafe so one could have a SoundData in a background thread and just feed it data!!

src/modules/data/soundData.h
nevyn: huh. Yes it can, and it doesn't even need casts since void* casts automatically... I think even in VS iirc? :S I'll commit and we'll see if it builds.

src/modules/data/soundData.c
nevyn: No reason. Fixed. Also made sure that soundData has blob :sweat_smile:

src/api/l_data_soundData.c
nevyn: oops :S it used to be bool now it's number

src/modules/audio/audio.c
nevyn: the idea of moving the memory of a struct owned by another library unnerves me :S Something inside of miniaudio could take a pointer to something inside the converter and assume that it wouldn't move between init and deinit, which I think is a sane assumption. If we want to have a dynamic number of converters, I think the choices are either heap allocing each converter, or using a linked list.

src/api/l_data_soundData.c
nevyn: that's much nicer! gret call :)

src/api/l_data.c
nevyn: well... I wouldn't phrase it as a "need" for an extra method, but an opportunity for a clearer API :P but that's just my Objective-C influenced opinion; it's your API, I can implement the above suggestion :)

src/modules/audio/audio.c
nevyn: Hmm. I'm not sure that's great. For one thing, the audio code already has a few places where loops very easily turn into infinite loops. E g, setting looping on a stream right now hangs the mix function. Reusing the looping flag for something else feels like it overcomplicates something which is already complicated.

The second reason for not doing it is that in understanding/experience, you would want to ensure that you have enough data buffered so you have headroom for stutters before you play again. This headroom is very implementation specific (low if you're generating data on another thread, higher if you're receiving data from network), so a generic solution would likely just create a lot of stutter as the source goes from playing to non-playing every frame or so.

Maybe it's just my implementations that are too fragile :P But that's my take on why I think stream sources should only played when asked to. If you don't agree, I can just avoid using the flag myself in my high latency scenarios, so I don't mind implementing your suggestion if you still think it's a good idea!

src/modules/audio/audio.c
nevyn: @bjornbytes I made converters heap allocated instead so the array is an array of pointers and can be safely expanded. I also added converter deiniting, which was missing.

nevyn:
I bumped to a version of miniaudio that fixes a Quest issue, so now it should work on all platforms afaik

PR #351, opened 2020-12-16
New API for mic capture

nevyn:
I realized it's silly to have two separate ring buffer abstractions. Nicer to back mic capture by a proper SoundData. Then I can even move that over a channel to another thread, and read from it without a lock! (E g move mic data _directly_ from audio thread to network thread, without touching main thread).

I think the API is nicer too.

This is based on top of https://github.com/bjornbytes/lovr/pull/347 .

### DIFF
--- a/src/api/l_audio.c
+++ b/src/api/l_audio.c
@@ -24,6 +24,13 @@ static int l_lovrAudioStop(lua_State* L) {
   return 0;
 }
 
+static int l_lovrAudioIsRunning(lua_State* L) {
+  AudioType type = luax_checkenum(L, 1, AudioType, "playback");
+  bool isRunning = lovrAudioIsRunning(type);
+  lua_pushboolean(L, isRunning);
+  return 1;
+}
+
 static int l_lovrAudioGetVolume(lua_State* L) {
   lua_pushnumber(L, lovrAudioGetVolume());
   return 1;
@@ -119,6 +126,7 @@ static int l_lovrAudioSetCaptureFormat(lua_State *L) {
 static const luaL_Reg lovrAudio[] = {
   { "start", l_lovrAudioStart },
   { "stop", l_lovrAudioStop },
+  { "isRunning", l_lovrAudioIsRunning },
   { "getVolume", l_lovrAudioGetVolume },
   { "setVolume", l_lovrAudioSetVolume },
   { "newSource", l_lovrAudioNewSource },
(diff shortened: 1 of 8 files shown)


nevyn:
I have some minor commits that are not related to new API for mic capture that I'm going to add here because they're small and I don't want a _fourth_ PR going for such small stuff :S

nevyn:
This PR now includes all my outstanding miniaudio changes, and has merged my review fixes from `miniaudio-devices`

nevyn:
@bjornbytes This is ready for review if you're happy with https://github.com/bjornbytes/lovr/pull/347

nevyn:
@bjornbytes I've merged latest miniaudio-new into this so this is ready to be reviewed/merged :)

src/api/l_audio.c
```diff
@@ -30,6 +24,13 @@ static int l_lovrAudioStop(lua_State* L) {
   return 0;
 }
 
+static int l_lovrAudioIsRunning(lua_State* L) {
```
bjornbytes, line 27: I may change to `isStarted` to match start/stop

src/modules/audio/audio.c
```diff
@@ -524,7 +472,15 @@ void lovrAudioSetCaptureFormat(SampleFormat format, int sampleRate)
 
 void lovrAudioUseDevice(AudioType type, const char *deviceName) {
   free(state.config[type].deviceName);
-  state.config[type].deviceName = strdup(deviceName);
+
+#ifdef ANDROID
+  // XX<nevyn> miniaudio doesn't seem to be happy to set a specific device an android (fails with
+  // error -2 on device init). Since there is only one playback and one capture device in OpenSL,
+  // we can just set this to NULL and make this call a no-op.
```
bjornbytes, line 479: Strange.  If getDevices returns valid devices but they don't work with useDevice, maybe we should file this with miniaudio.

src/api/l_audio.c
nevyn: ooh that's much better!

src/modules/audio/audio.c
nevyn: yes you're right. There seems to be two different bugs on Quest and Quest 2 too. Quest 1 returns _no_ capture devices even though it clearly has one; and Quest 2 has the above bug. I should file them with miniaudio. Until it's fixed upstream we'll need this workaround though. I can make sure to remove the workaround once upstream has a fix

PR #352, opened 2020-12-19
Oculus spatializer for miniaudio branch

mcclure:
This is a large patch which adds a new Oculus Audio spatializer. Oculus Audio is slightly different from the dummy spatializer in a few ways:

- It *must* receive fixed-size input buffers, every time, always.
- It can only handle a fixed number of spatialized sound sources at a time.
- It has a concept of "tails"; the spatialization of a sound can continue after the sound itself ends (eg echo).

Changes to audio.c were needed to support Oculus Audio's quirks:

- audio.c now supports a "fixedBuffer" mode which invokes the generator/spatializer in fixed size chunks
- Each source now has an intptr_t "memo" field that the spatializer may use to store whatever (Oculus spatializer uses this to handle the sound source limit).
- The spatializer interface got a couple new methods: A "tail" method which returns a sound buffer after all sources are processed; and "create" and "destroy" methods that are called when a sound source is created or destroyed (Oculus spatializer uses this to populate/clear the "memo" field).

Along the way some other miscellaneous changes got made:

- lovr.audio.getSpatializerName() returns the current spatializer
- Spatializer init now takes in "config in" and "config out" structs (Spatializer changes fields in config out to request things, currently fixed buffer mode).
- lovr.conf now takes t.audio.spatializer (string name of desired spatializer) and t.audio.spatializerMaxSourcesHint (Spatializers with max sources limits like Oculus will use this as the limit).
- audio.c went back to tracking position/orientation as vectors rather than a matrix
- A file oculus_spatializer_math_shim.h was added containing a minimal copypaste of OVR_CAPI.h from Oculus SDK to support a ovrPoseStatef the spatializer API needs. This may have license consequences but we are probably OK via a combination of fair use and the fact that a user cannot use this header file without accepting Oculus's license through other means.
- (Standalone commits) Fixed build problems with MSVC2019 and Mercurial
- (Standalone commit) I added an "audio tap" debug feature that can only be enabled by uncommenting a particular line in audio.c. This dumps all audio played to a file. This is invaluable for debugging weird audio bugs.

Some work remains to be done, in particular there is an entire reverb feature I did not touch and **LOVR_USE_OCULUS_AUDIO cannot be activated from tup**. Oculus Spatializer works better when it has velocity and time information but this patch does not supply it.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -25,6 +25,7 @@ option(LOVR_USE_OCULUS "Enable the LibOVR backend for the headset module (be sur
 option(LOVR_USE_VRAPI "Enable the VrApi backend for the headset module" OFF)
 option(LOVR_USE_PICO "Enable the Pico backend for the headset module" OFF)
 option(LOVR_USE_DESKTOP_HEADSET "Enable the keyboard/mouse backend for the headset module" ON)
+option(LOVR_USE_OCULUS_AUDIO "Enable the Oculus Audio spatializer (be sure to also set LOVR_OCULUS_AUDIO_PATH)" OFF)
 option(LOVR_USE_LINUX_EGL "Use the EGL graphics extension on Linux" OFF)
 
 option(LOVR_SYSTEM_ENET "Use the system-provided enet" OFF)
@@ -343,6 +344,36 @@ if(LOVR_ENABLE_AUDIO)
     src/api/l_audio_source.c
     src/lib/miniaudio/miniaudio.c
   )
+
+  if(LOVR_USE_OCULUS_AUDIO)
+    if(NOT LOVR_OCULUS_AUDIO_PATH)
+      message(FATAL_ERROR "LOVR_USE_OCULUS_AUDIO requires the LOVR_OCULUS_AUDIO_PATH to be set to the location of the Oculus Spatializer Native (AudioSDK) folder")
+    endif()
+
+    include_directories("${LOVR_OCULUS_AUDIO_PATH}/Include")
+
+    add_library(OculusAudio SHARED IMPORTED)
+    if (ANDROID)
+      set_target_properties(OculusAudio PROPERTIES IMPORTED_LOCATION "${LOVR_OCULUS_AUDIO_PATH}/Lib/Android/arm64-v8a/libovraudio64.so")
+    elseif(WIN32) # Note: This has *not* been tested.
+      if (CMAKE_GENERATOR_PLATFORM STREQUAL "x64")
+        set_target_properties(OculusAudio PROPERTIES IMPORTED_IMPLIB "${LOVR_OCULUS_AUDIO_PATH}/Lib/x64/ovraudio64.lib")
+        set_target_properties(OculusAudio PROPERTIES IMPORTED_LOCATION "${LOVR_OCULUS_AUDIO_PATH}/Lib/x64/ovraudio64.dll")
+      else () # ARM and ARM64 are also valid values for CMAKE_GENERATOR_PLATFORM but we will ignore this
+        set_target_properties(OculusAudio PROPERTIES IMPORTED_IMPLIB "${LOVR_OCULUS_AUDIO_PATH}/Lib/Win32/ovraudio32.lib")
+        set_target_properties(OculusAudio PROPERTIES IMPORTED_LOCATION "${LOVR_OCULUS_AUDIO_PATH}/Lib/Win32/ovraudio32.dll")
+      endif()
+    elseif (APPLE) # This *might* be possible-- there is a .framework in "macub/" ("universal binary"?)
+      message(FATAL_ERROR "LOVR_USE_OCULUS_AUDIO is not currently supported on Apple platforms")
+    else() # Assume Linux. Note: This has *not* been tested. FIXME: When is the .so copied?
+      set_target_properties(OculusAudio PROPERTIES IMPORTED_LOCATION "${LOVR_OCULUS_AUDIO_PATH}/Lib/Linux64/libovraudio64.so")
+    endif()
+
+    set(LOVR_OCULUS_AUDIO OculusAudio)
+    add_definitions(-DLOVR_ENABLE_OCULUS_AUDIO)
+    target_sources(lovr PRIVATE src/modules/audio/spatializers/oculus_spatializer.c)
+    target_link_libraries(lovr ${LOVR_OCULUS_AUDIO}) # TODO: Reorder?
+  endif()
 endif()
 
 if(LOVR_ENABLE_DATA)
@@ -533,6 +564,7 @@ if(WIN32)
   move_dll(${LOVR_ODE})
   move_dll(${LOVR_OPENVR})
   move_dll(${LOVR_MSDF})
+  move_dll(${LOVR_OCULUS_AUDIO})
   target_compile_definitions(lovr PRIVATE -DLOVR_GL)
 elseif(APPLE)
   target_link_libraries(lovr objc)
--- a/src/api/l_audio.c
+++ b/src/api/l_audio.c
@@ -17,6 +17,8 @@ StringEntry lovrTimeUnit[] = {
   { 0 }
 };
 
+#define AUDIO_SPATIALIZER_MAX_SOURCES_HINT 16
+
 static int l_lovrAudioReset(lua_State* L) {
   lovrAudioReset();
   return 0;
@@ -137,6 +139,11 @@ static int l_lovrAudioListCaptureDevices(lua_State *L) {
   return 0;
 }
 
+static int l_lovrAudioGetSpatializerName(lua_State *L) {
+  lua_pushstring(L, lovrSourceGetSpatializerName());
+  return 1;
+}
+
 static const luaL_Reg lovrAudio[] = {
   { "reset", l_lovrAudioReset },
   { "start", l_lovrAudioStart },
@@ -150,6 +157,7 @@ static const luaL_Reg lovrAudio[] = {
   { "setCaptureDevice", l_lovrAudioSetCaptureDevice },
   { "getCaptureDevice", l_lovrAudioGetCaptureDevice },
   { "listCaptureDevices", l_lovrAudioListCaptureDevices },
+  { "getSpatializerName", l_lovrAudioGetSpatializerName },
   { NULL, NULL }
 };
 
@@ -157,8 +165,35 @@ int luaopen_lovr_audio(lua_State* L) {
   lua_newtable(L);
   luax_register(L, lovrAudio);
   luax_registertype(L, Source);
-  AudioConfig config[2] = { { .enable = true, .start = true }, { .enable = false, .start = false } };
-  if (lovrAudioInit(config)) {
+
+  char *spatializer = NULL;
+  int spatializerMaxSourcesHint = AUDIO_SPATIALIZER_MAX_SOURCES_HINT;
+  luax_pushconf(L);
+  lua_getfield(L, -1, "audio");
+  if (lua_istable(L, -1)) {
+    lua_getfield(L, -1, "spatializerMaxSourcesHint");
+    if (lua_type(L, -1) == LUA_TNUMBER)
+      spatializerMaxSourcesHint = lua_tointeger(L, -1);
+    lua_pop(L, 1);
+
+    lua_getfield(L, -1, "spatializer");
+    if (lua_type(L, -1) == LUA_TSTRING)
+      spatializer = lua_tostring(L, -1);
+    lua_pop(L, 1);
+  }
+  lua_pop(L, 2);
+
+  AudioConfig config = {
+    .spatializer = spatializer,
+    .spatializerMaxSourcesHint=spatializerMaxSourcesHint
+  };
+
+  // First config is for output device, second is for input device
+  AudioDeviceConfig deviceConfig[2] = {
+    { .enable = true, .start = true },
+    { .enable = false, .start = false }
+  };
+  if (lovrAudioInit(config, deviceConfig)) {
     luax_atexit(L, lovrAudioDestroy);
   }
   return 1;
--- a/src/api/l_audio.c
+++ b/src/api/l_audio.c
@@ -166,7 +166,7 @@ int luaopen_lovr_audio(lua_State* L) {
   luax_register(L, lovrAudio);
   luax_registertype(L, Source);
 
-  char *spatializer = NULL;
+  const char *spatializer = NULL;
   int spatializerMaxSourcesHint = AUDIO_SPATIALIZER_MAX_SOURCES_HINT;
   luax_pushconf(L);
   lua_getfield(L, -1, "audio");
@@ -190,8 +190,8 @@ int luaopen_lovr_audio(lua_State* L) {
 
   // First config is for output device, second is for input device
   AudioDeviceConfig deviceConfig[2] = {
-    { .enable = true, .start = true },
-    { .enable = false, .start = false }
+    [AUDIO_PLAYBACK] = { .enable = true, .start = true },
+    [AUDIO_CAPTURE] = { .enable = false, .start = false }
   };
   if (lovrAudioInit(config, deviceConfig)) {
     luax_atexit(L, lovrAudioDestroy);
--- a/src/modules/audio/audio.c
+++ b/src/modules/audio/audio.c
@@ -90,7 +90,7 @@ static int generateSource(Source* source, float* output, uint32_t count) {
   ma_uint64 framesIn = 0;
   while (framesIn < count) { // Read from source until raw buffer filled
     ma_uint64 framesRequested = count - framesIn;
-    // FIXME: Buffer size math will break (crash) if channels > 1
+    // FIXME: Buffer size math will break (crash) if channels > 2
     ma_uint64 framesRead = source->sound->read(source->sound, source->offset, framesRequested,
       raw + framesIn * SampleFormatBytesPerFrame(source->sound->channels, source->sound->format));
     framesIn += framesRead;
--- a/src/modules/audio/audio.c
+++ b/src/modules/audio/audio.c
@@ -19,7 +19,10 @@ static const ma_format miniAudioFormatFromLovr[] = {
 #define OUTPUT_FORMAT SAMPLE_F32
 #define OUTPUT_CHANNELS 2
 #define CAPTURE_CHANNELS 1
-#define CAPTURE_BUFFER_SIZE ((int)(LOVR_AUDIO_SAMPLE_RATE * 1.0))
+
+#define CALLBACK_PERIODS 3
+#define PERIOD_LENGTH 128
+#define CALLBACK_LENGTH (PERIOD_LENGTH*CALLBACK_PERIODS)
 
 //#define LOVR_DEBUG_AUDIOTAP
 #ifdef LOVR_DEBUG_AUDIOTAP
@@ -39,7 +42,9 @@ struct Source {
   bool playing;
   bool looping;
   bool spatial;
-  float transform[16];
+  float position[4];
+  float orientation[4];
+  intptr_t spatializerMemo; // Spatializer can put anything it wants here
 };
 
 static inline int outputChannelCountForSource(Source *source) { return source->spatial ? 1 : OUTPUT_CHANNELS; }
@@ -47,13 +52,19 @@ static inline int outputChannelCountForSource(Source *source) { return source->s
 static struct {
   bool initialized;
   ma_context context;
-  AudioConfig config[AUDIO_TYPE_COUNT];
+  AudioDeviceConfig deviceConfig[AUDIO_TYPE_COUNT];
   ma_device devices[AUDIO_TYPE_COUNT];
   ma_mutex playbackLock;
   Source* sources;
   ma_pcm_rb captureRingbuffer;
   arr_t(ma_data_converter*) converters;
   Spatializer* spatializer;
+  bool fixedBuffer;
+  uint32_t bufferSize;
+  float *scratchBuffer1, *scratchBuffer2; // Used internally by mix(). Contains bufferSize stereo frames.
+  float *persistBuffer;  // If fixedBuffer, preserves excess audio between frames.
+  float *persistBufferContent; // Pointer into persistBuffer
+  uint32_t persistBufferRemaining; // In fixedBuffer mode, how much of the previous frame's mixBuffer was consumed?
 
 #ifdef LOVR_DEBUG_AUDIOTAP
   bool audiotapWriting;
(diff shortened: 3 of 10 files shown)


src/api/l_audio.c
```diff
@@ -150,15 +157,43 @@ static const luaL_Reg lovrAudio[] = {
...
+  int spatializerMaxSourcesHint = AUDIO_SPATIALIZER_MAX_SOURCES_HINT;
+  luax_pushconf(L);
+  lua_getfield(L, -1, "audio");
+  if (lua_istable(L, -1)) {
+    lua_getfield(L, -1, "spatializerMaxSourcesHint");
+    if (lua_type(L, -1) == LUA_TNUMBER)
+      spatializerMaxSourcesHint = lua_tointeger(L, -1);
+    lua_pop(L, 1);
+
+    lua_getfield(L, -1, "spatializer");
```
mcclure, line 179: - Is it weird you set this with "spatializer" but read it back with "spatializerName"?
- Should this optionally accept a table list of names rather than a single string name?

src/modules/audio/audio.h
```diff
@@ -54,9 +59,12 @@ float lovrSourceGetVolume(Source* source);
 void lovrSourceSetVolume(Source* source, float volume);
 bool lovrSourceGetSpatial(Source *source);
 void lovrSourceSetPose(Source *source, float position[4], float orientation[4]);
+void lovrSourceGetPose(Source *source, float position[4], float orientation[4]);
 uint32_t lovrSourceGetTime(Source* source);
 void lovrSourceSetTime(Source* source, uint32_t sample);
 struct SoundData* lovrSourceGetSoundData(Source* source);
+const char *lovrSourceGetSpatializerName();
+intptr_t *lovrSourceGetSpatializerMemoField(Source *source);
```
mcclure, line 67: Couldn't come up with a better way to do this (you need to be able to both get and set the memo field rapidly and frequently).

src/modules/audio/spatializer.h
```diff
@@ -2,14 +2,46 @@
...
-  void (*apply)(Source* source, mat4 transform, const float* input /*mono*/, float* output/*stereo*/, uint32_t frames);
+
+  // input is mono, output is interleaved stereo, framesIn is mono frames, framesOut is stereo frames.
+  // Safe to assume framesIn == framesOut unless spatializer requests needFixedBuffer.
+  // Return value is number of samples written into output.
+  uint32_t (*apply)(Source* source, const float* input, float* output, uint32_t framesIn, uint32_t framesOut);
+  // called at end of frame for any "additional noise", like echo.
+  // output is stereo, frames is stereo frames, scratch is a buffer the length of output (in case that helps)
+  // return value is number of stereo frames written.
+  uint32_t (*tail)(float *scratch, float* output, uint32_t frames);
```
mcclure, line 26: Forgot to mention this in the commit note but-- apply() may now return 0 to silence a source totally

src/modules/audio/spatializers/oculus_spatializer.c
```diff
@@ -0,0 +1,222 @@
...
+  state.sampleRate = configIn.sampleRate;
+  configOut->needFixedBuffer = true;
+  state.sourceMax = configIn.maxSourcesHint;
+  state.sources = calloc(state.sourceMax, sizeof(SourceRecord));
+
+  if (!state.poseLockInited) {
+    int mutexStatus = ma_mutex_init(&state.poseLock);
+    lovrAssert(mutexStatus == MA_SUCCESS, "Failed to create audio mutex");
+    state.poseLockInited = true;
+  }
```
mcclure, line 43: I don't kill the mutex between destroy/init cycles. Is this the correct convention? I'm terrified of deleting a mutex while a thread is waiting on it

src/modules/audio/spatializers/oculus_spatializer.c
```diff
@@ -0,0 +1,222 @@
...
+        ma_mutex_unlock(&state.poseLock);
+
+        ovrAudio_SetListenerPoseStatef(state.context, &pose); // Upload pose
+      }
+      state.poseUpdated = false;
+    }
+  }
+
+  intptr_t *spatializerMemo = lovrSourceGetSpatializerMemoField(source);
+
```
mcclure, line 94: I apologize for the complexity of this next section-- it's just plain doing something complicated. It's actually simpler than the first two versions…!
Things to note: Aside from the tail-killing rule it doesn't try to be "fair", if you play too many sounds at once it mutes the sources that haven't nabbed record fields yet, then they will suddenly (and unfairly) unmute as other sources stop.
Note this is all complicated and therefore a likely site for bugs, but I haven't stress-tested it yet.

src/modules/audio/spatializers/oculus_spatializer.c
```diff
@@ -0,0 +1,222 @@
...
+      if (outStatus & ovrAudioSpatializationStatus_Finished) { // Source done playing, echo tailoff is done
+        state.sources[idx].occupied = false;
+      }
+    }
+    return framesOut;
+  }
+  return 0;
+}
+
+static uint32_t oculus_spatializer_tail(float *scratch, float* output, uint32_t frames) {
```
mcclure, line 152: The "scratch" buffer I added here is admittedly weird but audio.c does have an appropriately-sized buffer but sitting around.

Note that if we changed the semantics here so that instead of returning a buffer that audio.c mixes in, the tail function "transformed" the otherwise finished buffer (ie: mixed in the tails to the output itself), we could save both time (save one mix pass in audio.c) and space (scratchBuffer2 in audio.c could be half its current length because it would no longer need to store stereo). But that is less clean so I didn't want to do it without consulting…

src/modules/audio/spatializers/oculus_spatializer.c
```diff
@@ -0,0 +1,222 @@
...
+        memset(output, 0, frames*sizeof(float)*2);
+      }
+      ovrAudio_SpatializeMonoSourceInterleaved(state.context, idx, &outStatus, scratch, NULL);
+      if (outStatus & ovrAudioSpatializationStatus_Finished) {
+        state.sources[idx].occupied = false;
+      }
+      for(int i = 0; i < frames*2; i++) {
+        output[i] += scratch[i];
+      }
+    }
```
mcclure, line 168: Reverb would go here if we added support for it

src/modules/audio/spatializers/oculus_spatializer.c
```diff
@@ -0,0 +1,222 @@
...
+  ovrPosef pose;
+  oculusUnpackVec(&pose.Position, position);
+  oculusUnpackQuat(&pose.Orientation, orientation);
+  out->ThePose = pose;
+  float zero[4] = {0}; // TODO
+  oculusUnpackVec(&out->AngularVelocity, zero);
+  oculusUnpackVec(&out->LinearVelocity, zero);
+  oculusUnpackVec(&out->AngularAcceleration, zero);
+  oculusUnpackVec(&out->LinearAcceleration, zero);
+  out->TimeInSeconds = 0; //TODO-OS
```
mcclure, line 192: Note velocity and time inputs would go here

src/modules/audio/spatializers/oculus_spatializer.c
```diff
@@ -0,0 +1,225 @@
...
+  ma_mutex_unlock(&state.poseLock);
+}
+static void oculus_spatializer_source_create(Source *source) {
+  intptr_t *spatializerMemo = lovrSourceGetSpatializerMemoField(source);
+  *spatializerMemo = -1;
+}
+static void oculus_spatializer_source_destroy(Source *source) {
+  intptr_t *spatializerMemo = lovrSourceGetSpatializerMemoField(source);
+  if (*spatializerMemo >= 0)
+    state.sources[*spatializerMemo].source = NULL;
```
mcclure, line 214: Note this line means we MUST delete sources THEN call lovrAudioDestroy (in that order) or this line will crash

mcclure:
Very slightly modified version of break showcase I've been using to test this https://gist.github.com/mcclure/44afd6d8303c2b0c43ea3267e991eca6

src/modules/audio/audio.c
```diff
@@ -31,90 +42,192 @@ struct Source {
...
+    aux = output;
+  }
 
-    if (source->spatial) {
-      state.spatializer->apply(source, source->transform, aux, mix, framesOut);
+  bool sourceFinished = false;
+  ma_uint64 framesIn = 0;
+  while (framesIn < count) { // Read from source until raw buffer filled
+    ma_uint64 framesRequested = count - framesIn;
+    // FIXME: Buffer size math will break (crash) if channels > 1
```
mcclure, line 93: Despite the comment this will work for channels <= 2, however this is still a problem. We could fix this by moving the ma_data_converter_process_pcm_frames call inside this loop. Is it possible to load a channels > 2 source or is there a check stopping this earlier?

src/modules/audio/audio.c
```diff
@@ -210,6 +356,8 @@ bool lovrAudioInitDevice(AudioType type) {
   config.playback.channels = OUTPUT_CHANNELS;
   config.capture.channels = CAPTURE_CHANNELS;
   config.dataCallback = callbacks[type];
+  config.periodSizeInFrames = PERIOD_LENGTH;
+  config.periods = 3;
```
mcclure, line 360: In my tests last year on Quest, setting these gave me buffers which were always a multiple of PERIOD_LENGTH. Testing on Windows this does not happen, it ignores what I tell it and gives me buffers of length 441 always, which means we will always be a little delayed.

src/modules/audio/audio.c
bjornbytes: Currently only mono and stereo SoundData/Sources are supported.  However, I would like to support ambisonic sound formats soon, which can have many channels (I think 4 is common).  This may or may not cause problems, because I'm not sure if ambisonics will fall under the same codepaths in the mixer or if they'll need special handling.

src/api/l_audio.c
```diff
@@ -150,15 +157,43 @@ static const luaL_Reg lovrAudio[] = {
...
+
+  char *spatializer = NULL;
+  int spatializerMaxSourcesHint = AUDIO_SPATIALIZER_MAX_SOURCES_HINT;
+  luax_pushconf(L);
+  lua_getfield(L, -1, "audio");
+  if (lua_istable(L, -1)) {
+    lua_getfield(L, -1, "spatializerMaxSourcesHint");
+    if (lua_type(L, -1) == LUA_TNUMBER)
+      spatializerMaxSourcesHint = lua_tointeger(L, -1);
+    lua_pop(L, 1);
```
bjornbytes, line 177: If LÖVR was like "the max number of sources is always 64" would this help simplify some of this stuff surrounding the max source count

src/api/l_audio.c
```diff
@@ -150,15 +157,43 @@ static const luaL_Reg lovrAudio[] = {
...
+  }
+  lua_pop(L, 2);
+
+  AudioConfig config = {
+    .spatializer = spatializer,
+    .spatializerMaxSourcesHint=spatializerMaxSourcesHint
+  };
+
+  // First config is for output device, second is for input device
```
bjornbytes, line 191: If this is confusing, maybe the array could read like

```c
deviceConfig[2] = {
  [AUDIO_PLAYBACK] = { .enable = true, .start = true },
  [AUDIO_CAPTURE] = { .enable = false, .start = false }
};
```

src/modules/audio/audio.c
```diff
@@ -155,8 +271,36 @@ bool lovrAudioInit(AudioConfig config[2]) {
...
+    if (config.spatializer && strcmp(config.spatializer, spatializers[i]->name))
+      continue; // If a name was provided, only match spatializers with that exact name
+    if (spatializers[i]->init(spatializerConfigIn, &spatializerConfigOut)) {
+      state.spatializer = spatializers[i];
+      break;
+    }
+  }
+  lovrAssert(state.spatializer != NULL, "Must have at least one spatializer");
+
+  state.fixedBuffer = spatializerConfigOut.needFixedBuffer;
```
bjornbytes, line 291: Given oculus/phonon both need fixed-size buffers, maybe this should always be true?  Non-spatial sources could skip the buffering, but maybe it's bad to mix buffered and non-buffered sources because then you'd get weird timing.

src/api/l_audio.c
mcclure: If LOVR set that requirement then we could remove almost all this logic from oculus spatializer (effectively because the source-limit logic would then be in audio.c) but there would still be a little bit of ugliness because oculus spatializer *does* have "tails", audio.c doesn't know about the tails, and the tails are bound to ID numbers. I don't think this is worth removing capability from apps not using this spatializer.

src/modules/audio/audio.c
mcclure: In the current implementation, when fixed buffer mode is on, non-spatial sources use the fixed buffer too. It would be potentially tricky to do otherwise.

I think it's worth having an unbuffered mode, if only because some games might not care about audio spatialization at all and maybe some do care about the (minimal, like .005sec) delay that the buffering introduces. (I actually have been slowly making simple music software in Lovr that renders entirely into the mirror and doesn't use the headset or 3d graphics, LOL)

src/modules/audio/spatializer.h
```diff
@@ -2,14 +2,46 @@
 #include "core/maf.h"
 
 typedef struct {
-  bool (*init)(void);
+	int maxSourcesHint;
+	int fixedBuffer;
+	int sampleRate;
```
bjornbytes, line 7: Caution: #347 adds the ability to dynamically change the device sample rate from Lua.

It seems like either the spatializer would need to be able to reinitialize when this occurs, or the dynamic sample rate feature would need to be removed.

Note that SteamAudio also needs the sample rate specified and would run into this same issue.

mcclure: It would not be at all hard to shut down and restart the spatializer as long as it is okay for all sounds to potentially stop. In fact a good way to implement the sample rate change would be to call spatializer.destroy and then spatializer.init again. This would require zero change to the spatializer code (unless the audio callback is running during the sample rate change).

bjornbytes: Interesting, phonon only supports 24000, 44100, 48000.  Not immediately relevant though

src/modules/audio/spatializers/oculus_spatializer.c
```diff
@@ -0,0 +1,225 @@
...
+static uint32_t oculus_spatializer_tail(float *scratch, float* output, uint32_t frames) {
+  bool didAnything = false;
+  for (int idx = 0; idx < state.sourceMax; idx++) {
+    if (state.sources[idx].occupied && !state.sources[idx].usedSourceThisPlayback) {
+      uint32_t outStatus = 0;
+      if (!didAnything) {
+        didAnything = true;
+        memset(output, 0, frames*sizeof(float)*2);
+      }
+      ovrAudio_SpatializeMonoSourceInterleaved(state.context, idx, &outStatus, scratch, NULL);
```
bjornbytes, line 163: What are the semantics of passing in NULL for the input buffer?  I guess -- what is the tail actually doing here?

mcclure: In this scenario, the sound is over, so there is no input to pass in. However the spatialized version of the sound may be longer than the "normal" version of the sound. ovrAudio knows what sound we are talking about because we are passing in the index.

The lifetime of a OVR source looks like:
- Call "reset" on index to set up a new sound source
- One or more times, call SetSourcePos on the index and Spatialize on the index with input
- After done with input, call Spatialize again with NULL input on index until a Spatialize call returns the Finished status

Why are spatialized sounds longer than nonspatialized ones? How much longer? What does the "tail" contain? I... don't know. I thought it was reverb (see section "Finishing Reverb Tails" in header file docs), but actually there is a separate MixInSharedReverb function (which would go in the tail() call, but which we're not calling) and we haven't turned reverb on anyway. I assume if the engine has nothing to add, we'll get Finished essentially immediately.

Anyway I'll add a comment to make this clearer.

src/modules/audio/audio.c
bjornbytes: Given that it is unreliable, should we still set it?  It looks like the current fixed buffer implementation is set up to work with any callback size.  If the only benefit to setting the period config is to sometimes get good callback sizes on certain backends, maybe we should instead just leave it unset and let miniaudio do its thing.  I trust miniaudio to pick better buffer sizes/timing intervals than lovr can (it's possible this is just a reflection of me not understanding this stuff well enough though).

Either way, explicitly setting the periods here overrides the `config.performanceProfile` setting below, because the performance profile is just a higher level way to configure it.  Seems like we should pick one or the other.

src/modules/audio/audio.c
```diff
@@ -198,6 +337,13 @@ void lovrAudioDestroy() {
   }
   arr_free(&state.converters);
   memset(&state, 0, sizeof(state));
+  free(state.scratchBuffer1); state.scratchBuffer1 = NULL;
+  free(state.scratchBuffer2); state.scratchBuffer2 = NULL;
+  free(state.persistBuffer);  state.persistBuffer = NULL;
+
+#ifdef LOVR_DEBUG_AUDIOTAP
+  state.audiotapWriting = false;
+#endif
```
bjornbytes, line 346: The buffers aren't getting freed here.  The memset clears `state`, so free is getting NULL.  Can move the frees above the memset and skip the explicit setting to NULL/false.

src/modules/audio/audio.c
```diff
@@ -19,7 +19,18 @@ static const ma_format miniAudioFormatFromLovr[] = {
...
+#define PERIOD_LENGTH 128
+#define CALLBACK_LENGTH (PERIOD_LENGTH*CALLBACK_PERIODS)
+
+//#define LOVR_DEBUG_AUDIOTAP
+#ifdef LOVR_DEBUG_AUDIOTAP
+// To get a record of what the audio callback is playing, define LOVR_DEBUG_AUDIOTAP,
+// after running look in the lovr save directory for lovrDebugAudio.raw,
+// and open as raw 32-bit stereo floats (Audacity can do this, or Amadeus on Mac)
+#include "filesystem/filesystem.h"
+#endif
```
bjornbytes, line 33: I am a little weary about maintaining this.  It seems prone to rot since it will be infrequently used, and adds a inter-module dependency which I try to avoid.

I would be more comfortable turning this into a real feature like "loopbacks".  miniaudio has real loopback devices but they're windows only.  It could maybe be emulated like this:

- In the playback callback, if loopback is enabled write everything to a ringbuffer/stream-SoundData.  Expose that to Lua so people can do whatever they want with the contents (write to file, send over network, calculate amplitude, etc.).  Loopback could be treated as enabled if Lua has set a loopback destination buffer (`lovr.audio.setLoopback(stream|nil)`, or whatever it's named).

CMakeLists.txt
```diff
@@ -343,6 +344,36 @@ if(LOVR_ENABLE_AUDIO)
...
+      message(FATAL_ERROR "LOVR_USE_OCULUS_AUDIO is not currently supported on Apple platforms")
+    else() # Assume Linux. Note: This has *not* been tested. FIXME: When is the .so copied?
+      set_target_properties(OculusAudio PROPERTIES IMPORTED_LOCATION "${LOVR_OCULUS_AUDIO_PATH}/Lib/Linux64/libovraudio64.so")
+    endif()
+
+    set(LOVR_OCULUS_AUDIO OculusAudio)
+    add_definitions(-DLOVR_ENABLE_OCULUS_AUDIO)
+    target_sources(lovr PRIVATE src/modules/audio/spatializers/oculus_spatializer.c)
+    target_link_libraries(lovr ${LOVR_OCULUS_AUDIO}) # TODO: Reorder?
+  endif()
```
bjornbytes, line 376: This should follow the same pattern for 3rd party libraries where it goes up above the lovr target and sets the `LOVR_OCULUS_AUDIO` var to the target (I don't know *why* things are like this but...should at least be consistent about it).  The variable can be added to the big list of targets on line 320ish.  The `target_sources` and `add_definitions` should stay here though.

For the FIXME about moving the .so on linux, don't worry about it.  Libraries aren't moved on linux and right now the rpath takes care of it.  I am going to do some relative rpath stuff soon so it matches macos though.

How does the library get moved into lib/arm64-v8a on android?

src/modules/audio/audio.c
mcclure: If it is unreliable on PC but reliable on Quest, that still makes it useful on Quest. It also might be it can be made reliable on PC, or possibly reliability depends on some weird system quirk like which Windows sound backend is being used.

src/modules/audio/audio.c
mcclure: All I know is I wind up hacking this in almost every time I need to work on lovr audio.

Would you be more comfortable if there were an explicit comment saying "this is for debug convenience but not maintained"? The loopback feature sounds super cool but I don't know how long it would take us to develop it.

(In one of my portaudio branches I implemented an ad hoc loopback feature like this and used it to display the waveforms in realtime-- it was pretty nice!)

CMakeLists.txt
mcclure: "How does the library get moved into lib/arm64-v8a on android?" Uh... this is a good question. I think I haven't tested this specific patch on Android (the CMake code is copied from the previous version which WAS tested on android, but that was only tested on the old lovr-oculus-mobile system).

src/modules/audio/audio.c
mcclure: (To be clear if you're asking me to remove it, I'll remove it)

src/api/l_audio.c
```diff
@@ -150,15 +157,43 @@ static const luaL_Reg lovrAudio[] = {
...
-  if (lovrAudioInit(config)) {
+
+  const char *spatializer = NULL;
+  int spatializerMaxSourcesHint = AUDIO_SPATIALIZER_MAX_SOURCES_HINT;
+  luax_pushconf(L);
+  lua_getfield(L, -1, "audio");
+  if (lua_istable(L, -1)) {
+    lua_getfield(L, -1, "spatializerMaxSourcesHint");
+    if (lua_type(L, -1) == LUA_TNUMBER)
+      spatializerMaxSourcesHint = lua_tointeger(L, -1);
```
bjornbytes, line 176: Maybe this should use optinteger so there are errors if it's an unexpected type (also please use braces for multiline ifs)

src/api/l_audio.c
bjornbytes: - I think it's okay
- table vs. string thoughts
  - If I set the spatializer to 'oculus' but accidentally run it on a build without the spatializer compiled in, should I get an error or fall back to the null spatializer?
  - Do I ever want to build in multiple spatializers and have it try multiple (maybe for a cross-platform project)?  This seems unlikely because I am the one compiling in the spatializers and it seems rare for a spatializer to be "unsupported" once it is compiled in and tested.  Dynamically changing the spatializer seems like a choice I can make at packaging time, not run time.
  - I guess I lean towards string, or even just a boolean (aka no selection, spatializer depends on build).

src/modules/audio/spatializers/oculus_spatializer.c
bjornbytes: It seems like it should be destroyed somewhere.  I think we can maintain invariants that A) the callbacks aren't firing when the spatializer is destroyed (devices aren't started while spatializer is being created/destroyed) B) Sources aren't doing anything while spatializers are being created/destroyed (this requires spatializers to only be destroyed in lovrAudioDestroy after sources are gone and/or destroying a spatializer any other time to take out the playback lock maybe?)

src/modules/audio/spatializers/oculus_spatializer.c
bjornbytes: I see that docs section now

> If late reverberation and simple box room modeling are enabled then the there
will be more output for the reverberation tail after the input sound has finished.
To ensure that the reverberation tail is not cut off, you can continue to feed the
spatialization functions with silence (e.g. NULL source data) after all of the
input data has been processed to get the rest of the output. When the tail has
finished the OutStatus will contain ovrAudioSpatializationStatus_Finished flag.

so because we didn't call ovrAudio_Enable with the LateReverberation and SimpleRoomModeling flags, the tails (presumably) don't do anything and they're just there for if we add reverb later?

mcclure: That sounds right to me.

src/modules/audio/audio.c
bjornbytes: Yeah, 256 samples at 44100 is almost 6 milliseconds which feels like a long time.

src/modules/audio/spatializers/oculus_spatializer.c
bjornbytes: Maybe we should wait to do the tails until we're actually doing reverb?  There's still some stuff to figure out for that (deciding how to expose the room mesh/size, designing a common API/implementation for all spatializers).  It would also simplify the id assignment for now

src/modules/audio/audio.c
bjornbytes: I can add the loopback feature before merging miniaudio-new

bjornbytes:
I resolved the conflicts and merged this into miniaudio-new

mcclure:
Sorry I never got around to the style issues. Maybe I can fix those when I come back to add reverb.

src/modules/audio/audio.c
```diff
@@ -31,90 +42,192 @@ struct Source {
...
-        return false;
-      }
-    } else {
-      source->offset += framesIn;
+  ma_uint64 framesOut = framesIn;
+
+  if (source->spatial) {
+    // Fixed buffer mode we have to pad buffer with silence if it underran
+    if (state.fixedBuffer && sourceFinished) {
+      memset(aux + framesIn, 0, (count - framesIn)*sizeof(float)); // Note always mono
```
bjornbytes, line 118: I wonder if this ought to use `framesConverted` instead of `framesIn` -- the aux buffer stores converted samples, but this line is scrolling through the aux buffer based on the number of raw samples *read* (framesIn), instead of the number of samples *converted* (framesConverted).  This could lead to accidentally zeroing out valid data or leaving in garbage data if conversion leads to a different sample count.  Do I have that right?  Still squinting at it

src/modules/audio/spatializers/oculus_spatializer.c
```diff
@@ -0,0 +1,226 @@
...
+    if (state.poseUpdated) {
+      { // Tell Oculus Audio where the headset is
+        ovrPoseStatef pose;
+
+        ma_mutex_lock(&state.poseLock); // Do nothing inside lock but make a copy of the pose
+        memcpy(&pose, &state.pose, sizeof(pose));
+        state.poseUpdated = false;
+        ma_mutex_unlock(&state.poseLock);
+
+        ovrAudio_SetListenerPoseStatef(state.context, &pose); // Upload pose
```
bjornbytes, line 87: I'm actually starting to get skeptical that Oculus is actually using the full pose info.  They offer `ovrAudio_SetListenerVectors` that takes raw floats instead of ovrPosef, and the docs for SetListenerPoseStatef say

> there is no GetListenerPoseStatef since the pose data is not cached internally.  Use ovrAudio_GetListenerVectors() instead to get the listener's position and orientation info.

They also explicitly say that they don't model the doppler effect.

If they really do only need the position and orientation, there is an opportunity to remove the pasted OVR_Math header and greatly simplify the setListenerPose function in here.

bjornbytes: Also, using `ovrAudio_setListenerVectors` can be no worse that what we are currently doing, since we aren't filling in any of the velocity/acceleration/timestamp info.

mcclure: Interesting. I'd like to ask on the Oculus dev forums for clarification (note: they will ignore me)

PR #356, opened 2021-01-10
Add support for fixed joint in physics

jmiskovic:
Fixed joint is used to instructing colliders to maintain fixed distance
AND relative rotation. Such functionality required 2 joints before.This
is similar to adding multiple shapes to same collider. The difference is
fixed joint is more dynamic (can be added and removed anytime without
disturbing simulation) and is less rigid.

The joint's force can be changed with `joint:setTightness(f)`, while
speed of response and frequency of spring oscillations is changed with
`joint:setResponseTime(t)`.

When created, joint marks current relative position and orientation
between colliders. At any later time that relative orientation can be
set again by moving any of colliders and calling `joint:reset()`.

In VR the fixed joint can be used for rigging physical systems. Some
examples are gripping and holding onto physical objects, making item
that floats in air, trampoline, inventory item strapped to belt,
breakable objects...

### DIFF
--- a/src/api/api.h
+++ b/src/api/api.h
@@ -33,6 +33,7 @@ extern const luaL_Reg lovrCollider[];
 extern const luaL_Reg lovrCurve[];
 extern const luaL_Reg lovrCylinderShape[];
 extern const luaL_Reg lovrDistanceJoint[];
+extern const luaL_Reg lovrFixedJoint[];
 extern const luaL_Reg lovrFont[];
 extern const luaL_Reg lovrHingeJoint[];
 extern const luaL_Reg lovrMat4[];
--- a/src/api/l_physics.c
+++ b/src/api/l_physics.c
@@ -120,6 +120,16 @@ static int l_lovrPhysicsNewSliderJoint(lua_State* L) {
   return 1;
 }
 
+static int l_lovrPhysicsNewFixedJoint(lua_State* L) {
+  Collider* a = luax_checktype(L, 1, Collider);
+  Collider* b = luax_checktype(L, 2, Collider);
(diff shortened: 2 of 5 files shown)


jmiskovic:
`world = lovr.physics.newWorld(0,-2,0, false)

frame = world:newBoxCollider(0, 0, 0, 0.2, 0.2, 0.2)
frame:setKinematic(true)

box = world:newBoxCollider(0, 1.5, -1, 0.2)
box:setLinearDamping(0)
box:setAngularDamping(0)

joint = lovr.physics.newFixedJoint(frame, box)
joint:setTightness(0.7)
joint:setResponseTime(155)

function lovr.draw()
  lovr.graphics.cube('fill', mat4(box:getPose()):scale(0.2))
end

function lovr.update(dt)
  t = lovr.timer.getTime()
  if t % 2 < dt then
    box:applyForce(2,  1, 0)
    box:setOrientation(0.5, 0, 1, 1)
  end
  world:update(1/72)
end`

jmiskovic:
Having tested it a bit, it would seem that only _positional_ part of joint takes ERP and CFM into consideration. The _orientation_ is always fast and tight, even with low ERP and high CFM. This makes the fixed joint less useful than I thought it would be.

jmiskovic:
Also the app crashes on start on Quest and I assume this commit is responsible (found `dJointDestroy` in backtrace in logcat).

Let's put this one on hold.

bjornbytes:
This investigation is still useful.  Feel free to reopen if the situation changes

Something I learned while reviewing this: ball joints also accept CFM/ERP (only a couple joints have it).

PR #367, opened 2021-01-29
Store the Lua source location that created shaders and canvases

nevyn:
Otherwise it's really hard to know from error messages where
the fault originated in your source code, since the validation
of your shader/canvas occurs much later than the creation.

I implemented this patch to fix the pico build of alloverse. it was really hard to find the location on my own, but with this debug code it was super easy. It's not much of an overhead and could probably be super useful for debugging other shader problems in the future, so I figured I'd suggest it as an upstream feature. Feel free to discard if you think it's too noisy :)

### DIFF
--- a/src/modules/graphics/opengl.c
+++ b/src/modules/graphics/opengl.c
@@ -86,6 +86,7 @@ struct Canvas {
   bool needsAttach;
   bool needsResolve;
   bool immortal;
+  char *from;
 };
 
 struct ShaderBlock {
@@ -104,6 +105,7 @@ struct Shader {
   map_t uniformMap;
   map_t blockMap;
   bool multiview;
+  char *from;
 };
 
 struct Mesh {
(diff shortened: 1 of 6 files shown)


src/modules/graphics/opengl.c
```diff
@@ -1437,7 +1439,7 @@ void lovrGpuDiscard(Canvas* canvas, bool color, bool depth, bool stencil) {
 }
 
 void lovrGpuDraw(DrawCommand* draw) {
-  lovrAssert(state.singlepass != MULTIVIEW || draw->shader->multiview == draw->canvas->flags.stereo, "Shader and Canvas multiview settings must match!");
+  lovrAssert(state.singlepass != MULTIVIEW || draw->shader->multiview == draw->canvas->flags.stereo, "Shader and Canvas multiview settings must match! Canvas was created at %s, shader at %s", draw->canvas->from, draw->shader->from);
```
nevyn, line 1442: here's the line that saved me, and the reason for this whole change

bjornbytes:
In the new graphics rework all objects can have a `label` given to them (in the table passed to new<Object>), this string is used in error/log messages and also shows up in renderdoc.  It allows you to use any string, so you could potentially use `debug.traceback` from Lua and use that as the label.  Would that work?  It would be a forward-compatible change.

nevyn:
Oh that sounds great!

I like that it’s automatic; perhaps the label could automatically be set to the first line of trace back by default? :)

bjornbytes:
I don't think I'd want to use the stack trace as the default behavior:

- Getting the traceback isn't free.  If this behavior is the default when creating objects then some lovrers that don't want this would need to explicitly disable it.
- It could end up cluttering error/log messages, especially if the resource is created in a file nested in several subdirectories.
- If someone has a loadShader function that creates their shaders, the label would always be the same.  Seeing every shader named 'assets/shaders/loader.lua:51' in renderdoc might be less useful than the default behavior of 'Shader #1', #2, etc.

Admittedly these reasons are somewhat esoteric/minor.

These "asynchronous error messages" (like this multiview thing that's only validated last-minute at draw time) are pretty nasty though so I think lovr should try to provide as much context as it can or at least make it easy.

I should also mention that this canvas+multiview issue won't be a thing on vulkan

shakesoda:
if you're going to generate labels from source at all i strongly recommend tagging them with a unique id yourself in addition to the origin (e.g. `shader <id> (origin)`, id stays first to scan lists easily).

PR #369, opened 2021-02-11
Pass -DLOVR_SANITIZE to CMake config for AddressSanitizer

mcclure:
Unsure this is ready to merge.

Problems:
- I send the same flags to all three of the C compiler, C++ compiler, and Linker. Are all three really necessary?
- It would be nice to turn on ubsan too. The problem is in order to turn on ubsan you have to block ubsan from being used in (1) ODE (2) LuaJIT. I'm not sure I understand LuaJIT well enough to selectively set CMAKE_C_FLAGS for some projects but not others.

This COULD go in master but I put it on dev because applying it to master would create a dev merge conflict.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -24,6 +24,7 @@ option(LOVR_USE_PICO "Enable the Pico backend for the headset module" OFF)
 option(LOVR_USE_DESKTOP_HEADSET "Enable the keyboard/mouse backend for the headset module" ON)
 option(LOVR_USE_OCULUS_AUDIO "Enable the Oculus Audio spatializer (be sure to also set LOVR_OCULUS_AUDIO_PATH)" OFF)
 option(LOVR_USE_LINUX_EGL "Use the EGL graphics extension on Linux" OFF)
+option(LOVR_SANITIZE "Enable Address Sanitizer" OFF)
 
 option(LOVR_SYSTEM_GLFW "Use the system-provided glfw" OFF)
 option(LOVR_SYSTEM_LUA "Use the system-provided Lua" OFF)
@@ -334,6 +335,12 @@ target_link_libraries(lovr
   ${LOVR_PTHREADS}
   ${LOVR_EMSCRIPTEN_FLAGS}
 )
+
+if (LOVR_SANITIZE)
+  set(LOVR_SANITIZE_FLAGS "-fsanitize=address,undefined" "-O1" "-fno-omit-frame-pointer")
+  target_compile_options(lovr PRIVATE ${LOVR_SANITIZE_FLAGS})
+  target_link_options(lovr PRIVATE ${LOVR_SANITIZE_FLAGS})
+endif()
 
 if(LOVR_ENABLE_AUDIO OR LOVR_ENABLE_DATA)
   target_sources(lovr PRIVATE


bjornbytes:
- The compiler and linker both need the flags.  However, since LÖVR is written in C, the C++ flags shouldn't be necessary.
- Perhaps the asan flags could be applied only to the `lovr` target to prevent current/future dependencies from causing issues (I now realize that this causes mixing of asan and non-asan code, but this is also how I've always been doing things with tup.  Maybe that's okay or maybe it's just accidentally been working fine this whole time).  The way to do that would be `target_compile_options(lovr PRIVATE "-fsanitize=address")`.
- How come `-O1` is used?  Doesn't this prevent the normal optimization settings from working properly?

mcclure:
* Applying to the lovr target only makes sense. Maybe some of the other targets that aren't known-unsanitary could have it applied as well. That would also justify dropping the C++ flags since we're narrowly capturing the lovr code (C only).
* I used -O1 because that's what the Clang instructions said to do https://clang.llvm.org/docs/AddressSanitizer.html The reason is that ASAN prints stack traces but the stack traces can be wrong without O1.

mcclure:
I have a better patch for this, will push

mcclure:
Okay. I think this version is ready to merge. 

Two potential changes first/after:
- It's called LOVR_USE_ASAN but it's both ASAN and UBSAN. Should it just be LOVR_USE_SAN or LOVR_USE_SANITIZE?
- The asan and ubsan are added to only the lovr target. In fact, they should probably be added to *every* target EXCEPT ODE and LuaJIT. It would probably be nice to know if a library is corrupting the heap.
    - Also ODE could get ASAN-- it just happens to violate ubsan harmlessly in one place.
    - Also LuaJIT could get ASAN (but not ubsan) *if* we either strategically patch it with a couple `__attribute__((no_sanitize("address")))`s or use the asan blacklist feature. (LuaJIT normally never violates ASAN with a single exception, which is that its string functions like to do things in 32-bit chunks even if this means running past the end of a string. This is safe because all modern mallocs round up to the nearest 32 bit chunk, but ASAN pretends it doesn't know that.)

Unless you have opinions about the name I think we could merge this now, and maybe later add ASAN/UBSAN to the libraries with maybe a blacklist for the known ODE ubsan violation and LuaJIT asan violation.

bjornbytes:
The tup option is called SANITIZE, I think `LOVR_SANITIZE` or `LOVR_USE_ASAN` would be fine for CMake.

Does this work on MSVC?  Is it safe to pass these options to that compiler?

mcclure:
Accidentally misclicked and closed this.

I replaced _USE_ASAN with SANITIZE to match tup.

> Does this work on MSVC? Is it safe to pass these options to that compiler?

This is an interesting question. Probably MSVC has a similar feature (looking on Google I also find MSVC appears to support, literally, Asan, though this might require the clang plugin) but it's not going to work with the same options.

I'm probably gonna be back on Windows by the end of the week, I can look into that then. If it turns out unviable, how would you feel about LOVR_SANITIZE just throwing a "not supported" error if you try to use it with the MSVC generator?

bjornbytes:
Ok.  I don't have access to wndows right now so I will wait for you/others to confirm.  If asan doesn't work there then the error sounds good to me.  I do think the check should specifically be for MSVC or 'not gcc/clang', since clang can be used on win32.

bjornbytes:
Just gonna merge this as-is, it will only harm unsuspecting MSVC users that dare to enable it, and the value it provides likely eclipses whatever pain that will cause.

PR #370, opened 2021-02-11
Give shaderBlock:send() the ability to send only part of the block

mcclure:
shaderBlock:send() has two additional arguments with this patch: An offset and a size.
If present, only the range [offset…offset+size) will be sent.

Usually this is not super useful because it is usually efficient-ish to send even a large shader block all at once.
I believe the reason I added this was I had a shaderblock that was being simultaneously written into by send()s from code and written into by a compute shader, so I wanted to avoid stomping the compute-shader-owned part of the shader block.
Anyway it's nice to have.

This is an API change and will require a doc change.

### DIFF
--- a/src/api/l_graphics_shaderBlock.c
+++ b/src/api/l_graphics_shaderBlock.c
@@ -42,10 +42,15 @@ static int l_lovrShaderBlockSend(lua_State* L) {
     Blob* blob = luax_checktype(L, 2, Blob);
     Buffer* buffer = lovrShaderBlockGetBuffer(block);
     void* data = lovrBufferMap(buffer, 0, false);
+    int offset = luaL_optinteger(L, 3, 0);
+    lovrAssert(offset >= 0, "Negative offset");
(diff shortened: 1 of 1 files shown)


bjornbytes:
Please stick to the formatting used by the rest of the codebase (spaces after casts, spaces around binary operators).

mcclure:
Apologies, we've discussed this before and I keep forgetting.

Two small style-adjacent notes:
- I provide more information in these lovrAsserts than lovr asserts usually do. Usually the asserts just say "Bound exceeded" and don't provide numbers.
- The parenthesis on line 52 are unnecessary, though they make it look more readable to me.

bjornbytes:
Thanks.  For errors, basically anything goes right now.  At some point there will be more structured helpers like `lovrCheckRange` (maybe with a config flag to disable the checks) but it constantly tumbles down in priority.  Extra parens are cool, I've been using those for readability for ternaries and stuff.

PR #372, opened 2021-02-13
First attempt at texture-based fonts / custom glyphs

mcclure:
This is a feature for custom glyphs and entirely custom fonts. It consists of two things:

- A lovr.texture.newTextureFont() — this creates a font which is not backed by a TTF file. Instead, a texture is provided at creation time. (To mesh well with other Lovr APIs, the font does have a rasterizer, but it is a "dummy" rasterizer with no TTF.)
- A font:addTextureGlyph() — this takes an existing font (it does not have to be a texture font) and adds a custom glyph with specific metrics and with manually supplied texture coordinates.

Potential uses for this are:
- "Pixel art fonts" which do not use SDF (this is what I developed it for)
- Create a font as a prebaked SDF texture instead of a TTF
- Potential future work: Add additional glyphs to a TTF font— if we:
    - Add the ability to add arbitrary image content to a TTF font's texture (notice that all I've added is the ability to add *glyphs*, with specified texture coordinates— the image content must already be present)
    - Add a SVG->MSDF converter

    Then we could create the ability to add custom glyphs for unicode symbols by selecting SVGs, converting them to MSDF, and adding the glyph to the font. This would allow you to, for example, print icons relevant to your game into the middle of printed text.

Attached is a demo. It uses a PNG containing a dump of the font bitmap from the original Apple // computer and creates a font out of it (with custom per-character advance— the image is packed as 7x8 glyphs but displays as 8x8 glyphs). It does require some fancy footwork on the Lua side (see pixelFont.lua), but I think that may be necessary here to have a simple and general interface that accommodates a range of ways of importing custom fonts.
[textureFontDemo.zip](https://github.com/bjornbytes/lovr/files/5976375/textureFontDemo.zip)
Screen Shot 2021-02-13 at 2 50 44 PM

### DIFF
--- a/src/api/l_graphics.c
+++ b/src/api/l_graphics.c
@@ -1264,6 +1264,24 @@ static int l_lovrGraphicsNewFont(lua_State* L) {
   return 1;
 }
 
+static int l_lovrGraphicsNewTextureFont(lua_State* L) {
+  Texture* texture = luax_checktype(L, 1, Texture);
+  float pixelDensity = luaL_checknumber(L, 2);
+  int height = luaL_checkinteger(L, 3);
+  double spread = luaL_optnumber(L, 4, 4.); // Only needed if we are using SDF font
+  Rasterizer* rasterizer = lovrRasterizerDummyCreate(1, height);
+  Font* font = lovrFontCreate(rasterizer, 0, spread); // FIXME: Might a texture font ever need to set padding?
+
+  lovrRetain(texture);
+  lovrFontSetPixelDensity(font, pixelDensity);
+  lovrFontSetTextureDataExternal(font, texture);
+
+  luax_pushtype(L, Font, font);
+  lovrRelease(rasterizer, lovrRasterizerDestroy);
+  lovrRelease(font, lovrFontDestroy);
+  return 1;
+}
+
 static int l_lovrGraphicsNewMaterial(lua_State* L) {
   Material* material = lovrMaterialCreate();
 
@@ -1821,6 +1839,7 @@ static const luaL_Reg lovrGraphics[] = {
   // Types
   { "newCanvas", l_lovrGraphicsNewCanvas },
   { "newFont", l_lovrGraphicsNewFont },
+  { "newTextureFont", l_lovrGraphicsNewTextureFont },
   { "newMaterial", l_lovrGraphicsNewMaterial },
   { "newMesh", l_lovrGraphicsNewMesh },
   { "newModel", l_lovrGraphicsNewModel },
--- a/src/api/l_graphics_font.c
+++ b/src/api/l_graphics_font.c
@@ -94,15 +94,48 @@ static int l_lovrFontHasGlyphs(lua_State* L) {
   Font* font = luax_checktype(L, 1, Font);
   Rasterizer* rasterizer = lovrFontGetRasterizer(font);
   bool hasGlyphs = true;
-  for (int i = 2; i <= lua_gettop(L); i++) {
-    if (lua_type(L, i) == LUA_TSTRING) {
-      hasGlyphs &= lovrRasterizerHasGlyphs(rasterizer, lua_tostring(L, i));
-    } else {
-      hasGlyphs &= lovrRasterizerHasGlyph(rasterizer, luaL_checkinteger(L, i));
+  if (lovrRasterizerIsDummy(rasterizer)) {
+    for (int i = 2; i <= lua_gettop(L); i++) {
+      if (lua_type(L, i) == LUA_TSTRING) {
+        hasGlyphs &= lovrFontHasGlyphsCached(font, lua_tostring(L, i));
+      } else {
+        hasGlyphs &= lovrFontHasGlyphCached(font, luaL_checkinteger(L, i));
+      }
+    }
+  } else {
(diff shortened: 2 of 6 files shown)


bjornbytes:
I will just post this incomplete thought early but I am still working to understand everything here: it seems like if Rasterizer can preserve or strengthen its role as "object that stores or generates glyphs" then that would somehow be good.  If this image font was simply an official second type of Rasterizer that had an ImageData, maybe there are advantages compared to having a somewhat awkward dummy rasterizer.  Maybe it would reduce conditionals or simplify the stuff Font needs to care about or allow things to be done if you aren't using the GPU (e.g. on threads).

bjornbytes:
It might depend on, when using this feature, whether your atlas is more likely to be stored in RAM (Image, Rasterizer) or more likely to be stored in VRAM (Texture, Font).

mcclure:
Thinking out loud:

> it seems like if Rasterizer can preserve or strengthen its role as "object that stores or generates glyphs" then that would somehow be good

At the moment, in the Lovr code, a Font is a thing that tracks a texture and an atlas of glyphs within that texture, and a Rasterizer is a thing that knows how to render glyphs and on demand amends a texture by rendering a new glyph into it. The Rasterizer, at present, does no management of the texture. Expanding the texture, and even deciding where in the texture a glyph is to be rendered, is the Font's job. The Rasterizer only serves to tell the font how big a glyph is going to be before the glyph is rendered (so it knows how to pack it) and then rendering it. It is a box for managing a TrueType/OTF font and pulling metric and image data out of that font.

In my model I'm proposing, there will sometimes be fonts that have no TrueType/OTF source associated with them. There will still be a rasterizer, because in the Lovr API the rasterizer is responsible for reporting certain kinds of "defaults" and those defaults might still be meaningful for a dummy (no OTF source) font. In the current patch, the dummy rasterizer does no work at all because the rasterizer's only job in the Lovr font model is to be a *fallback* (ie: a glyph is requested, it is not already, here is a source for producing a new one) and with the dummy rasterizer all glyphs that will ever be supported are present from the beginning. 

If we move forward with my proposed improvements, we *will* be adding glyphs to the font after its first initialization, *but not on demand*. The model I'm suggesting is that when the font (whether dummy or normal) is first created, before the font is used the client code will go over a list of symbols it expects to need and add those symbols to the texture. If the symbols are added from SVGs, we will need some sort of logic for rendering MSDFs, and if it makes more sense to keep MSDF rendering in rasterizer.c then sure, maybe that code can live in rasterizer.c. However my initial inclination is that this function (add an image to the texture and register it as a glyph) most naturally lies with the font rather than the rasterizer, because the rasterizer's model is about being a *fallback* which renders glyphs *on demand*. If it makes sense for there to be a "second kind" of rasterizer (as opposed to the do-nothing dummy rasterizer) the only thing I could think of would be some kind of callback you register with a rasterizer— when a rasterizer is asked to rasterize a glyph it looks for it in its font, if the glyph isn't in the font or the rasterizer doesn't have a font then instead it talks to the callback which pulls SVGs from a directory or does something algorithmic (?). I feel like I can't think of a really good use case for this callback approach, but maybe if I was trying to render CJK I'd feel differently.

bjornbytes:
I agree with the distinction between 

> when a rasterizer is asked to rasterize a glyph it looks for it in its font, if the glyph isn't in the font or the rasterizer doesn't have a font then instead it talks to the callback

LÖVE actually supports this "fallback font" system.  You can specify a list of Rasterizer objects that the Font is able to pull glyphs from, instead of just 1.  When it needs to paste a new glyph into its atlas it will try all of its Rasterizers in order until one of them has the glyph.

It seems like this could be useful for one of the features presented in the original PR description

> Add the ability to add arbitrary image content to a TTF font's texture

If I have a TTF font, and I want to augment it with an atlas of emoji or SVG icons that I made (ignoring MSDF for now), I think both approaches could work:

- With the current patch, the process would be:
  - Offline step: Create a combined atlas of the TTF glyphs and the icon atlas (LÖVR can't rasterize the TTF glyphs for me because the rasterizer needs to be a dummy rasterizer, right?)
  - Upload it to the GPU (newTexture)
  - Load it as a texture font
  - Specify glyph codepoints/uvs/metrics
- If there were two types of rasterizers (Image and TTF), and Font could have fallback rasterizers, the process is:
  - Create a TTF Rasterizer from the font file
  - Create an Image Rasterizer from the emoji atlas
    - Specify glyph codepoints/uvs/metrics for this rasterizer's atlas
  - Create a Font from both rasterizers, the current atlas system would function unmodified

Observations:

- When using a custom font texture, I need to make sure that the Texture has all the glyphs I need, and I'll probably need to make combined atlases for that.  To me that offline step and atlas combining feels cumbersome.  Maybe being able to add individual glyphs/Images/Textures to the Font after it's created would help with that, but I'm not sure how that would work since the Texture is supposed to be static.
- The Image Rasterizer isn't convenient if I am generating glyphs/atlases dynamically using the GPU.  I'd need to render them, read back the texture into an Image, and then create a rasterizer from that.  Not sure how common that is.

bjornbytes:
This needs design work resolved before it can move forwards, I'm going to close but discussion should still continue

PR #373, opened 2021-02-17
Steam Audio

bjornbytes:
This adds a spatializer that uses SteamAudio (also known as the much cooler name: phonon).

- I took a approach that is different than some of the other "optional libraries".  I linked with phonon at runtime instead of at build time.  This means that phonon could be enabled with regular lovr builds by putting the phonon library next to the lovr executable.  There is probably some disadvantage to using runtime linking (linking with the `dl` library for example, platform restrictions), but the advantage of not needing to recompile lovr may be worth it.  This approach could possibly be used with other libraries if it's successful.  There could also be a compile flag to optionally use this runtime linking method (it would only be 2-3 ifdefs).
- New Source accessors:
  - [x] `Source:setDirectivity` controls the dipole weight/power for making directional sources.
  - [x] `Source:setAbsorptionEnabled` controls air absorption.
  - [x] `Source:setFalloffEnabled` controls distance attenuation.
  - [x] `Source:setOcclusionEnabled` enables occlusion.
  - [x] `Source:setReverbEnabled` enables reverb.
  - [x] `Source:setTransmissionEnabled` enables transmission (hear through walls)
  - [x] `Source:setInterpolation` is used to control bilinear vs. nearest HRTF interpolation.
  - [x] `Source:setSpatialBlend` smoothly transitions between stereo and binaural.
  - [x] `Source:setRadius` allows for smooth occlusion transitions
- [x] `lovr.audio.setGeometry` sets scene geometry used for occlusion/reverb.
  - [ ] AudioMaterial for setting preset materials for the scene.
- [x] Reverb/Reflections

Features I'm skipping that could be supported at some point:

- Custom SOFA HRTFs
- GPU Raytracing
- Audio probes and baking (no idea how this works)
- Some way of using the simpler panning spatialization instead of HRTFs
- Per-triangle audio materials
- Dynamic audio meshes (meshes that can be instanced into the scene and moved around)


bjornbytes:
The dynamic meshes appear to be supported only with the Embree raytracer, so maybe I won't look into those.

PR #375, opened 2021-02-19
Use find_library to link android liblog.so

mcclure:
This is iffy.

So a couple weeks back I was having a problem with the Lovr Android. I looked on Stack Overflow. Somebody recommended to find android "log" lib using find_library instead of hardcoding -llog. I tried this. They claimed this is required for newer versions of Android / newer Android toolchains. It did not help me.
But!
It does appear the documentation suggests doing this in the case of log.
 https://developer.android.com/studio/projects/configure-cmake

Why?
Is this required?
Will it be required in future?
Is the documentation just trying to demonstrate how find_library works?

I have no idea!
It doesn't seem to make anything *worse*.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -602,7 +602,11 @@ elseif(EMSCRIPTEN)
   target_sources(lovr PRIVATE src/core/os_web.c)
   target_compile_definitions(lovr PRIVATE LOVR_WEBGL)
 elseif(ANDROID)
-  target_link_libraries(lovr log EGL GLESv3 OpenSLES android dl)
+  find_library( # Right now just linking "log" will always work, but this may not be true in future.
+    log-lib
+    log
+  )
+  target_link_libraries(lovr ${log-lib} EGL GLESv3 OpenSLES android dl)
   target_compile_definitions(lovr PRIVATE LOVR_GLES)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -602,7 +602,12 @@ elseif(EMSCRIPTEN)
   target_sources(lovr PRIVATE src/core/os_web.c)
   target_compile_definitions(lovr PRIVATE LOVR_WEBGL)
 elseif(ANDROID)
-  target_link_libraries(lovr log EGL GLESv3 OpenSLES android dl)
+  target_sources(lovr PRIVATE src/core/os_android.c)
```
bjornbytes, line 605: Is this line adding os_android here by mistake?  That file can't be used with pico

bjornbytes:
To maybe make things more weird: tup doesn't link to `log` at all and printf/__android_log_print still works fine

bjornbytes:
It seems like this might link against the absolute path to the library (in the NDK sysroot) instead of `-llog`, which is maybe better.  But it seems like if find_library is used, it should be used consistently for all of those system libraries, unless there's a reason not to?

bjornbytes:
Closing due to staleness, feel free to resurrect

PR #376, opened 2021-02-19
lovrSoundCreateFromCallback

mcclure:
Allows a Sound to be created from a callback. Because Sound already has a "read" callback, this will integrate cleanly with the existing lovr.audio module. The callback will be called to generate the audio.

The callback-based Sound is allowed to be of infinite length. It is allowed to store a "memo" pointer in the Sound. It is also allowed to have a custom "destroy" callback called on destruction (which must destroy the memo pointer if appropriate).

This is of zero immediate use. It is handy to me because I have forked lovr. If this were accepted upstream, it would make a lovr fork (like mine) easier to maintain because custom audio types could be added without modifying audio.c. However, most people have not forked lovr.

In future, I would like to attempt live pcm audio generation in Lua. (I have a prototype API for this in my PortAudio branch already.) This callback feature could naturally be used as the basis for the lua audio generation feature, because it would allow l_audio.c to contain the Lua callback functionality entirely within itself and not make audio.c have to be aware of Lua. However, that hypothetical feature would be substantially more complicated and probably require an architecture discussion.

### DIFF
--- a/src/modules/data/sound.c
+++ b/src/modules/data/sound.c
@@ -16,7 +16,7 @@ static const ma_format miniaudioFormats[] = {
 
 struct Sound {
   uint32_t ref;
-  uint32_t (*read)(Sound* sound, uint32_t offset, uint32_t count, void* data);
+  SoundCallback* read;
   struct Blob* blob;
   void* decoder;
   void* stream;
@@ -25,13 +25,15 @@ struct Sound {
   uint32_t sampleRate;
   uint32_t frames;
   uint32_t cursor;
+  void* callbackMemo; // When using lovrSoundCreateFromCallback, any state the read callback uses should be stored here
+  SoundDestroyCallback* callbackMemoDestroy; // This should be used to free the callbackMemo pointer (if appropriate)
 };
 
 // Readers
 
 static uint32_t lovrSoundReadRaw(Sound* sound, uint32_t offset, uint32_t count, void* data) {
   uint8_t* p = sound->blob->data;
-  uint32_t n = MIN(count, sound->frames - offset);
+  uint32_t n = sound->frames == LOVR_SOUND_ENDLESS ? count : MIN(count, sound->frames - offset);
   size_t stride = lovrSoundGetStride(sound);
   memcpy(data, p + offset * stride, n * stride);
   return n;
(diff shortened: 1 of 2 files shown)


src/modules/data/sound.c
```diff
@@ -25,13 +25,15 @@ struct Sound {
   uint32_t sampleRate;
   uint32_t frames;
   uint32_t cursor;
+  void *callbackMemo;
+  SoundDestroyCallback callbackMemoDestroy;
```
mcclure, line 29: In the spatializer, I named the custom data field "spatializerMemo", so I named it "memo" here as well.

bjornbytes, line 29: For informational purposes can you explain what the memo/destroy callback are being used for?  At some point it may help to know how it's being used.

src/modules/data/sound.c
```diff
@@ -362,3 +381,7 @@ uint32_t lovrSoundCopy(Sound* src, Sound* dst, uint32_t count, uint32_t srcOffse
 
   return frames;
 }
+
+void *lovrSoundGetCallbackMemo(Sound *sound) {
+  return sound->callbackMemo;
+}
```
bjornbytes, line 387: Sorry, this is incredibly minor: Can you configure your editor to add newlines to the end of files?  There is a file in the repository root `.editorconfig` that many editors (vim, emacs, vscode, sublime, etc.) can use to automatically do this when editing any of the lovr files.

src/modules/data/sound.c
bjornbytes, line 29: Note: A future plan that I may implement is adding a typed union (SoundType enum plus a union for callbackMemo/Destroy and stream and decoder).  This wouldn't have any impact on the public API

mcclure: Makes sense. I could imagine a callback making use of decoder, but blob less likely.

src/modules/data/sound.c
mcclure: I'll see what I can do sorry

src/modules/data/sound.h
```diff
@@ -18,9 +18,15 @@ typedef enum {
 } ChannelLayout;
 
 typedef struct Sound Sound;
+
+typedef uint32_t (*SoundCallback)(Sound* sound, uint32_t offset, uint32_t count, void* data);
+typedef void (*SoundDestroyCallback)(Sound* sound);
```
bjornbytes, line 23: Another minor thing!  Could these function pointers be declared without the "pointer" part, like this:

```c
typedef uint32_t SoundCallback(Sound* sound, uint32_t offset, uint32_t count, void* data);
typedef void SoundDestroyCallback(Sound* sound);
```

The types in the struct/prototypes just need a star like `SoundCallback* read`

I mention it because I won't be able to make this change later without unknowingly breaking your stuff

src/modules/data/sound.c
mcclure: You mean here or in a comment?

callbackMemo is for any arbitrary "state" associated with the read callback. For example imagine a Sound which is a sine wave generator. callbackMemo could be a pointer to a float representing the sine frequency.

callbackMemoDestroy is to free the memory used by callbackMemo.

In my current usage, I have a Sound whose callback refers to static [global] state for my sound effects. So callbackMemo is NULL, but callbackMemoDestroy is used (it sets a flag in the static structure noting that sound playback has stopped). Tomorrow I'm going to do PureData hookup and callbackData will point to a structure for the PureData playback channel associated with a particular Sound source (a buffer and a pointer to the generator, since one generator will generate several channels at once once per audio frame) but the callbackMemoDestroy might be blank because in that case the playback channels have their memory managed elsewhere.

I'll add a comment.

mcclure:
I think this new patch addresses all concerns Bjorn raised

bjornbytes:
Cool, thanks

mcclure:
Awesome, thanks! This closes down the biggest single divergence between Lovr and my fork ever since the start of my project.

PR #381, opened 2021-02-28
Fix MSVC build

mcclure:
- stdatomic.h had not defined ATOMIC_INT_LOCK_FREE, but it is now required
- It's _MSC_VER, not _MSVC_VER


bjornbytes:
Thanks

PR #382, opened 2021-03-05
Fix crash on startup on Oculus driver

mcclure:
I think the API for arr_t changed but the oculus driver code did not change to the new rules


mcclure:
I added a second commit that fixes a warning in headset_oculus. Because of some messiness in merging with my private fork, I'm putting the commit on this PR rather than making it its own PR. Sorry about this.

bjornbytes:
core/os is good since it uses os_window_set_vsync, thanks!

PR #384, opened 2021-03-15
Fix Android build

mcclure:
- Fix typos after "os" refactor
- Fix CMake finding javac

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -685,7 +685,7 @@ elseif(ANDROID)
     endif()
 
     if(ANDROID_PACKAGE)
-      set(PACKAGE_RENAME "--rename-manifest-package ${ANDROID_PACKAGE}")
+      set(PACKAGE_RENAME "--rename-manifest-package" "${ANDROID_PACKAGE}")
     endif()
 
     # Make an apk
--- a/src/core/os_android.c
+++ b/src/core/os_android.c
@@ -34,7 +34,7 @@ static int32_t onInputEvent(struct android_app* app, AInputEvent* event) {
     return 0;
   }
 
-  ButtonAction action;
+  os_button_action action;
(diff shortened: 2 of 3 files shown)


CMakeLists.txt
```diff
@@ -637,6 +637,7 @@ elseif(ANDROID)
   )
 
   if(LOVR_BUILD_EXE)
+    find_package(Java)
```
bjornbytes, line 640: This already exists on line 59, which confuses me

mcclure: Then I think this one is a merge error specific to my branch.

I will retest and remove this line.

src/core/os_android.c
```diff
@@ -280,27 +280,27 @@ void os_poll_events() {
   }
 }
 
-void os_on_quit(quitCallback callback) {
+void os_on_quit(fn_quit callback) {
```
bjornbytes, line 283: Also I think the function pointers here should have a star after them e.g. `fn_quit*`

mcclure: You are right, as I wrote them these functions would crash if called

mcclure:
I tested this new version with lovr-lodr. It addresses the issues above. I still was not able to get it to run without the `-0 so` -> `-0 ''` change we discussed in Slack.

bjornbytes:
The `-0 so` thing can be tracked here https://github.com/nothings/stb/pull/960#discussion_r596584133

But I am surprised that `0bb556554dd59ca706290cfad4796f410631924c` didn't fix it

mcclure:
I didn't test with 0bb5565. I tested with 6eef950aa84 (the parent of this patch)

mcclure:
I pushed the wrong commit on Friday :( :(

Third try!! How's this?

PR #387, opened 2021-03-22
Libluasocket support

brainrom:
Provides network functionalities. 
To build on android ndk requires some changes in lua5.0->lua5.1 compat layer, but finally it works.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,8 @@ option(LOVR_ENABLE_THREAD "Enable the thread module" ON)
 option(LOVR_ENABLE_TIMER "Enable the timer module" ON)
 
 option(LOVR_ENABLE_ENET "Bundle with lua-enet" ON)
+option(LOVR_ENABLE_LUASOCKET "Bundle with luasocket" ON)
+
 option(LOVR_ENABLE_JSON "Bundle with lua-cjson" ON)
 
 option(LOVR_USE_LUAJIT "Use LuaJIT instead of Lua" ON)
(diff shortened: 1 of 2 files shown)


bjornbytes:
It's cool to finally have this working.

Did you forget to add the `src/lib/luasocket` sources?

Either way, I don't think I will be merging this PR, because there is a fancy new plugin system that should allow this to be easily added to builds.  Plugins are currently on the `dev` branch, and I haven't documented them yet.  But the main workflow is to make a `plugins` folder in the repo root and copy/clone/submodule plugins into there.  Plugins are regular repositories with CMakeLists in them.  Here are 3 examples of plugins: [`bjornbytes/lua-cjson`](https://github.com/bjornbytes/lua-cjson), [`bjornbytes/lua-enet`](https://github.com/bjornbytes/lua-enet), [`bjornbytes/lua-deepspeech`](https://github.com/bjornbytes/lua-eepspeech).  Most of the time no changes need to be made to make stuff compatible as lovr plugins.  Let me know if I can help out with the process.

brainrom:
I tried to build libluasocket independently, but it heavily depend on lua machine, almost all protocol-specific code are written on lua and I haven't found another solution, than just compile it with the engine itself.

bjornbytes:
When built as a lovr plugin it will automatically have access to the Lua headers/libraries

brainrom:
@bjornbytes 
> Let me know if I can help out with the process.
Now I'm trying to rebuild it as LOVR [plugin](https://github.com/brainrom/lovr-luasocket), but encountered a problem (Error: no module 'luasocket.http' in file ./luasocket.so) to define submodules inside single .so/.dll file. AFAIK I should change luasocket.c someway to define all it's submodules but currently I can't figure out how exactly. Help is welcome.

brainrom:
So, with some workarounds now it can works. The final question is how to avoid symlinking libraries.

bjornbytes:
That is very cool.  So it looks like luasocket is using the "all-in-one" approach documented in [Lua's manual](https://www.lua.org/manual/5.1/manual.html#pdf-package.loaders), and lovr.filesystem's native library loader doesn't support that yet.  I can look into adding support for it, it would help match the functionality of Lua's built-in loader.

brainrom:
> (Implementation question I'm thinking about: Should it be a new loader that gets added or could it somehow be integrated into the existing one?)

Where it's located? AFAIU all loaders are placced inside `api/l_filesystem.c` and on line 429 metioned, that lovr should search native libraries next to executable, but instead of it, lovr search it in `pwd`. So, I think it's a bug.

PR #389, opened 2021-03-23
CMake build fix for Oculus Audio on Quest

mcclure:
Without this, building on Quest with CMake with OCULUS_AUDIO enabled will cause a crash when the apk is run.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -634,7 +634,8 @@ elseif(ANDROID)
 
   # Dynamically linked targets output libraries in raw/lib/<ABI> for easy including in apk with aapt
   set_target_properties(
-    lovr ${LOVR_ODE}
+    lovr
+    ${LOVR_ODE}
     PROPERTIES LIBRARY_OUTPUT_DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/raw/lib/${ANDROID_ABI}"
   )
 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -634,7 +634,8 @@ elseif(ANDROID)
 
   # Dynamically linked targets output libraries in raw/lib/<ABI> for easy including in apk with aapt
   set_target_properties(
-    lovr ${LOVR_ODE}
+    lovr
+    ${LOVR_ODE}
```
mcclure, line 638: Splitting this across two lines is not strictly relevant to oculus audio but would help a lot with forks (because it makes it possible to add additional items here without causing a merge conflict)

CMakeLists.txt
```diff
@@ -680,6 +681,11 @@ elseif(ANDROID)
       endif()
     endif()
 
+    if(LOVR_USE_OCULUS_AUDIO)
+      get_target_property(OCULUS_AUDIO_LIB ${LOVR_OCULUS_AUDIO} IMPORTED_LOCATION)
+      file(COPY ${OCULUS_AUDIO_LIB} DESTINATION raw/lib/${ANDROID_ABI})
+    endif()
+
```
mcclure, line 688: I don't like this, this is duplicative with the VRAPI case. The best way to do this would be to create a function like we have with move_dll, and use it with both the VRAPI/PICO lines and also this.

mcclure (commented):
Honestly not sure about this one:

mcclure:
TUP also needs to copy the ovraudio .so if it isn't already.

CMakeLists.txt
bjornbytes: Seems okay, there's like 4 `move_lib` functions that should try to get merged into one eventually.

bjornbytes:
tup is copying it already

PR #392, opened 2021-03-25
Allow boot.lua to run without debug object

mcclure:
This one is *very* silly and I'll understand if you don't take it…

I'm doing a [Rust port](https://github.com/mcclure/rovr) of Lovr, to learn Rust. I don't expect I'll finish this or that it will displace my use of the C version on my VR projects.

The Rust port uses the C version's exact `boot.lua` but does not, in its current early form, have the debug object. This means that if there is any error, the default lovr.errhand will crash with the brutal "error occurred while trying to process another error" message because it tries to call debug.traceback and fails. This patch simply checks for debug.traceback before calling it, which allows it to continue as normal and print the error message by itself.

Accepting this patch would be low cost (one if, triggered only when default lovr.errhand is called), would allow me to continue to use the *exact* boot.lua from the lovr repo possibly indefinitely and so avoid annoying repeated merges, and could also possibly help with other pathological situations, like other ports (we do talk sometimes about moving some of lovr to Zig) or someone for some reason I can't fathom specifically wanting to ship a game with the debug object omitted.

Note: I originally wrote this as "if debug and debug.traceback" for total armor plating but shortened it to "if debug" when I made the PR because I can't imagine even a hypothetical scenario where a vm includes an object named debug but doesn't implement some of the standard methods. I could change it back.

### DIFF
--- a/src/resources/boot.lua
+++ b/src/resources/boot.lua
@@ -238,7 +238,9 @@ end
 
 function lovr.errhand(message, traceback)
   message = tostring(message)
-  message = message .. formatTraceback(traceback or debug.traceback('', 4))
+  if debug then
+    message = message .. formatTraceback(traceback or debug.traceback('', 4))
+  end
   print('Error:\n' .. message)
   if not lovr.graphics then return function() return 1 end end
 


bjornbytes:
Can you explain why your Lua VM is not opening the standard `debug` library?

mcclure:
> Can you explain why your Lua VM is not opening the standard debug library?

No :(

It's just not :(

Embedding Lua in Rust doesn't use the normal C API, it uses this odd wrapper called mlua, it mostly is much easier than the C library but there have been 1 or 2 seemingly basic things that it's very nonobvious how to do them. I suspect there is just a flag I need to set somewhere to inject the debug object and I *plan* to fix that whenever I can get to it. Maybe I just fix it eventually and that resolves the problem that way, and we leave boot.lua as it is.

Realistically I *could* make *an* argument some people might want to intentionally disable the debug library in their shipping versions for security reasons, as it might enable breaking out of the lua sandbox. However this is a *very* weak argument since the lua sandbox is not much of a sandbox, the Lovr API functions are not hardened, and if you are using LuaJIT you don't need an exploit to break out of the sandbox because you can simply read and write memory.

bjornbytes:
I don't know anything about mlua or Rust but...how are you initializing it?  It seems it is able to control which libraries are opened, using either `new_with` or `load_from_std_lib`.  e.g. `Lua::new_with(StdLib::ALL)`.  This seems like the equivalent of doing `luaL_openlibs(L)` which is what LÖVR does.

I would prefer to either A) fully understand what is going on and avoid sprinkling caution where it is not needed or B) fully lean into the idea that no Lua standard library is guaranteed to be present and add similar checks to all uses of the standard libraries (string, debug, coroutine, maybe even baselib, etc.).  B seems like a lot of work though.

mcclure:
I *thought* that this isn't a concern with all libraries, just `debug`— I thought there was some kind of special configuration option where some copies of Lua ship with `debug` and some don't, and it's the one standard library that is an option.

But I am looking and I can't find anything about this in the documentation. Maybe I was just confused.

I have filed a support request with mlua to see why debug is missing in mlua…

bjornbytes:
It looks like mlua stuff was resolved with unsafe_new?  Please reopen if I was mistaken

mcclure:
Yeah.

I may come back later and ask for a slightly different patch. Basically, the trick is: mlua has a specific list of lua libraries it bans in "safe" mode, because they could be used to alter memory. The list is not documented. Long-term I think it would be interesting to offer support for safe mode, however I can't figure out how to support safe mode without a list of which libraries it drops! I have asked mlua guy for documentation but haven't heard back yet.

khvzak:
@mcclure In `mlua` [StdLib documentation](https://docs.rs/mlua/0.5.3/mlua/struct.StdLib.html) you could find a list of unsafe libraries. 
Only `ffi` (luajit) and `debug` are marked unsafe. 
Initially this behaviour has inherited from rlua origin where `debug` is not loaded by default.

mcclure:
@khvzak thank you for explaining!

PR #393, opened 2021-03-27
Phonon/SteamAudio support in CMake build

mcclure:
I tested this PR but only on Windows.


bjornbytes:
I think the 32 bit versions could be removed

mcclure:
I could do that. Do we know for a fact armv7 headsets are dead tho?

bjornbytes:
There isn't one officially supported by lovr.  32 bit is also generally decrepit in terms of optimization/testing here

mcclure:
Okay. All platforms including Windows? We're supporting 32 bit Windows for OculusAudio.

bjornbytes:
Yeah, same for windows

mcclure:
Removed 32-bit support

PR #397, opened 2021-03-28
Route ODE messages into LOVR log system

jmiskovic:
ODE errors, debugs and messages are redirected into LOVR's log system
by a callback mechanism for each.

The ODE submodule is updated to revision that does not crash when error
or debug occurs.

### DIFF
--- a/src/modules/physics/physics.c
+++ b/src/modules/physics/physics.c
@@ -39,11 +39,32 @@ static uint32_t findTag(World* world, const char* name) {
   return NO_TAG;
 }
 
+static void onErrorMessage(int num, const char* format, va_list args) {
+  char message[1024];
+  vsnprintf(message, 1024, format, args);
+  lovrLog(LOG_ERROR, "PHY", message);
+}
+
+static void onDebugMessage(int num, const char* format, va_list args) {
+  char message[1024];
+  vsnprintf(message, 1024, format, args);
+  lovrLog(LOG_DEBUG, "PHY", message);
+}
+
+static void onInfoMessage(int num, const char* format, va_list args) {
+  char message[1024];
+  vsnprintf(message, 1024, format, args);
+  lovrLog(LOG_INFO, "PHY", message);
+}
+
 static bool initialized = false;
 
 bool lovrPhysicsInit() {
   if (initialized) return false;
   dInitODE();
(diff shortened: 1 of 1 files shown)


jmiskovic:
Message callbacks will produce fully formatted message before passing them onto log. I saw lovrLog function could take format string and values as variadic function but AFAIK there's no way to dynamically generate a variadic function call from va_list.

src/modules/physics/physics.c
```diff
@@ -53,6 +56,27 @@ void lovrPhysicsDestroy() {
...
+  lovrLog(LOG_DEBUG, "ODE", composed_msg);
+}
+
+void onInfoMessage(int num, const char *msg, va_list args)
+{
+  char composed_msg[1024];
+  vsnprintf (composed_msg, 1024, msg, args);
+  lovrLog(LOG_INFO, "ODE", composed_msg);
+}
+
```
bjornbytes, line 79: I think these should be `static` functions and removed from the header, making them more "private".

src/modules/physics/physics.c
```diff
@@ -53,6 +56,27 @@ void lovrPhysicsDestroy() {
   initialized = false;
 }
 
+void onErrorMessage(int num, const char *msg, va_list args)
+{
+  char composed_msg[1024];
+  vsnprintf (composed_msg, 1024, msg, args);
+  lovrLog(LOG_DEBUG, "ODE", composed_msg);
```
bjornbytes, line 63: Should this be `LOG_ERROR` or maybe `LOG_WARN`?

src/modules/physics/physics.c
bjornbytes, line 79: Style notes:

- Opening braces for the functions should be on the same line, like `) {`
- Prefer full words for variables/parameters like `message`
- No spaces after function invocation `vsnprintf(`.
- Use camel case instead of snake case for variables e.g. `composedMessage`
- Pointer stars should be near the type instead of the variable `const char* message`

bjornbytes (commented):
This is great, thank you!

jmiskovic:
Good comments, thanks.
I also changed tag from ODE into more general PHY.

bjornbytes:
Awesome!

I will look into adding a `lovrLogv` later to avoid the copy

PR #398, opened 2021-04-01
Submo's for VR SDKs, and other changes

nevyn:
I'm gonna add a bunch of more changes; this is a start at least...


src/core/util.h
```diff
@@ -10,13 +10,21 @@
...
+  #define LOVR_NORETURN __declspec(noreturn)
+  #define LOVR_THREAD_LOCAL __declspec(thread)
+  #define LOVR_RESTRICT __restrict
+  #if defined(LOVR_BUILDING_SHARED) // building a dll; export symbols
+    #define LOVR_EXPORT __declspec(dllexport)
+  #elif defined(LOVR_BUILDING_EXE) // building exe: keep symbols internal
+    #define LOVR_EXPORT
+  #else // importing this header from a non-lovr project: import the symbols
+    #define LOVR_EXPORT __declspec(dllimport)
+  #endif
```
bjornbytes, line 22: Can you explain the reason why you're varying the __declspec?

nevyn: I added this to our fork a long time ago, trying to remember why it was needed :S it's probably related to us building a lovr.dll and a separate Alloverse.exe that loads it on Windows.  https://stackoverflow.com/a/58031/48125 says it's not required but is more efficient. you'll note enet, glfw, lua etc all do the same thing.

CMakeLists.txt
```diff
@@ -637,6 +637,7 @@ elseif(APPLE)
       MACOSX_BUNDLE TRUE
       MACOSX_RPATH TRUE
       BUILD_WITH_INSTALL_RPATH TRUE
+      INSTALL_NAME_DIR "@rpath"
```
bjornbytes, line 640: The `install_name` is only used by other applications that are trying to link LÖVR as a shared library, right?  Having a hard time understanding.

CMakeLists.txt
```diff
@@ -744,6 +745,7 @@ elseif(ANDROID)
         sign
         --ks ${ANDROID_KEYSTORE}
         ${ANDROID_APKSIGNER_KEYSTORE_PASS} ${ANDROID_KEYSTORE_PASS}
+        $<$<BOOL:${ANDROID_KEY_PASS}>:--key-pass> $<$<BOOL:${ANDROID_KEY_PASS}>:${ANDROID_KEY_PASS}>
```
bjornbytes, line 748: It would be nice to use a generator expression for the `--ks-pass` argument above as well

src/api/api.h
```diff
@@ -2,6 +2,7 @@
 #include <stdint.h>
 #include <stdarg.h>
 #include <stddef.h>
+#include "core/util.h"
```
bjornbytes, line 5: I'd like to avoid including util here if possible.

What are you using/exporting `lovrModules` for?  Could you use the new `luax_preload` function?  That function is meant to be an interface to have all the modules installed into a Lua state, without actually needing to know about the modules themselves.

src/core/util.h
bjornbytes: I'm not too concerned about the efficiency of LOVR_EXPORT because it is only used by the luaopen_* functions which are called only once at startup.

Although it's not needed right now, it can sometimes be useful to have `__declspec(dllexport)` when building an exe so you can dynamically load symbols from the executable (at one point this was being done to require lovr modules).

If (after potentially resolving the use of LOVR_EXPORT in api.h) you still need to use LOVR_EXPORT, maybe these two BUILDING_* defines could be replaced by just setting LOVR_EXPORT directly from CMakeLists.  The #define here could be guarded by an ifndef.  This would reduce the number of defines to keep track of and maybe simplify the macros here a bit.

I must admit that I'm getting caught up in the "poetry" of having the simpler set of platform-specific defines in this file, and am trying to avoid piling on all the extra windows complexity if possible.

src/api/api.h
nevyn: I'm replacing main.c with my own main function, so I'm using lovrModules to populate the Lua context. luax_preload sounds like exactly what I need :) I'll continue porting Alloverse to the dev branch, and potentially I can remove both the EXPORT stuff (it's only used for lovrModules right now anyway) AND the changes in this file. Watch this space :)

CMakeLists.txt
nevyn: I can't decide if I like it or not, but it IS nice with a oneliner, no? :sweat_smile: I'll set it up.

CMakeLists.txt
nevyn: I have written Mac apps for two decades and I'm still confused by install_name... Yes, I think it only applies to libraries. I see you took it out in latest dev; I'll play around with what's needed once I have this branch up and running on Linux and Windows.

src/core/util.h
nevyn: I moved it to this file with the "proper" declarations because I saw @mcclure also wanted export macros. I agree it's ugly... I'm not convinced defining it in CMake is less ugly since it's really source code, not a build setting; and it would be needed twice (cmake and tup). Windows just messes with everything, doesn't it :sweat_smile: 

The easiest solution for me is to pull it out once I remove the need for lovrModules, BUT that also feels contraproductive if mcc is gonna need it anyway. eh I'll just revisit once I get to the point where I'm messing with the Windows build...

CMakeLists.txt
```diff
@@ -35,6 +35,8 @@ option(LOVR_SYSTEM_OPENXR "Use the system-provided OpenXR" OFF)
 option(LOVR_BUILD_EXE "Build an executable (or an apk on Android)" ON)
 option(LOVR_BUILD_SHARED "Build a shared library (takes precedence over LOVR_BUILD_EXE)" OFF)
 option(LOVR_BUILD_BUNDLE "On macOS, build a .app bundle instead of a raw program" OFF)
+option(LOVR_BUILD_WITH_EXPORTED_INTERNALS "When building as a shared library, make all C symbols visible. This exposes internals, so be aware that any APIs you use from C may change at any time." OFF)
+option(LOVR_BUILD_EXTRA_PLUGINS "Any additional plugins you want to include in the build system" "")
```
bjornbytes, line 39: Maybe plugins could always be sourced from the top-level plugins folder instead of the one inside the lovr repo?

nevyn: @bjornbytes ah, from whatever the top CMakeLists.txt is? I like that idea!

nevyn:
I un-drafted this but then I realized I also have some quest mic permissions code I haven't upstreamed. Maybe you've already implemented it in some other way than me @bjornbytes? Anyways I'll go through this todo before I'm ready for merge:

- [ ] on-demand mic permissions on quest
- [ ] load plugins from root project instead of using a variable
- [ ] maybe also implement mic permissions on mac while I'm at it?
- [ ] go through https://github.com/alloverse/lovr/pull/2/files and see if there's anything else I've missed

bjornbytes:
Here is the current approach for microphone permissions:

- To start recording audio, you call `lovr.audio.start('capture')`.
- This may fail and return false.  One of the reasons for that failure may be permissions.
- If you want to properly support a platform with locked-down microphone permissions, first call `lovr.system.requestPermission('audiocapture')` and wait for the `lovr.permission` callback to fire.  You can implement custom logic for your application that integrates with this, like calling requestPermission at a good time (potentially on device input), rendering menus or spinners, etc.  Or you can just throw the requestPermission at the top of lovr.load and not even listen for the lovr.permission, it's up to you how detail-oriented you want to be.
- Once the `lovr.permission` callback is fired and grants audiocapture, retry `lovr.audio.start('capture')`.  It should no longer fail due to permissions.

CMakeLists.txt
nevyn: @bjornbytes This is implemented now :)

src/core/util.h
nevyn: This is gone now since I'm using LOVR_BUILD_WITH_EXPORTED_INTERNALS so I'm resolving this conversation

CMakeLists.txt
nevyn: okay it works fine without this addition so I'm leaving it out...

nevyn:
@bjornbytes okay this is ready for real. works on quest, mac, linux, windows :)

CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ option(LOVR_SYSTEM_OPENXR "Use the system-provided OpenXR" OFF)
 option(LOVR_BUILD_EXE "Build an executable (or an apk on Android)" ON)
 option(LOVR_BUILD_SHARED "Build a shared library (takes precedence over LOVR_BUILD_EXE)" OFF)
 option(LOVR_BUILD_BUNDLE "On macOS, build a .app bundle instead of a raw program" OFF)
+option(LOVR_BUILD_WITH_EXPORTED_INTERNALS "When building as a shared library, make all C symbols visible. This exposes internals, so be aware that any APIs you use from C may change at any time." OFF)
```
bjornbytes, line 38: Couple notes:

- The `LOVR_BUILD_*` options all drastically change the output format (exe, dll, app, etc.).  To me, choosing whether symbols are visible does not feel like it falls into this category.  Maybe it should just be a simpler `LOVR_EXPORT` or `LOVR_PUBLIC` or `LOVR_VISIBLE` option.  Also, it does not export internals, it exports everything.
- "When building as a shared library" -- it will also make symbols when building as an executable, right?  Perhaps that clause should be removed.
- The note about "be careful about using the C API", while accurate, feels out of place to me, I would consider removing.

CMakeLists.txt
```diff
@@ -214,9 +215,7 @@ endif()
 
 # Oculus SDK -- expects Oculus SDK 1.26.0 or later
 if(LOVR_ENABLE_HEADSET AND LOVR_USE_OCULUS)
-  if(NOT LOVR_OCULUS_PATH)
-    message(FATAL_ERROR "LOVR_USE_OCULUS requires the LOVR_OCULUS_PATH to be set to the location of the Oculus Desktop SDK (LibOVR) folder")
-  endif()
+  set(LOVR_OCULUS_PATH "${CMAKE_CURRENT_SOURCE_DIR}/deps/oculus-pc" CACHE STRING "location of the Oculus Desktop SDK folder")
```
bjornbytes, line 218: Very minor but 'location' should be capitalized...

CMakeLists.txt
```diff
@@ -214,9 +215,7 @@ endif()
 
 # Oculus SDK -- expects Oculus SDK 1.26.0 or later
 if(LOVR_ENABLE_HEADSET AND LOVR_USE_OCULUS)
```
bjornbytes, line 217: Note to self: this should probably yell at you if you're not on windows

CMakeLists.txt
```diff
@@ -225,12 +224,12 @@ if(LOVR_ENABLE_HEADSET AND LOVR_USE_OCULUS)
   endif()
   include_directories("${LOVR_OCULUS_PATH}/LibOVR/Include")
   link_directories("${LOVR_OCULUS_PATH}/LibOVR/Lib/Windows/${OCULUS_ARCH}/${OCULUS_BUILD_TYPE}/VS2017")
-  set(LOVR_OCULUS LibOVR)
+  set(LOVR_OCULUS "${LOVR_OCULUS_PATH}/LibOVR/Lib/Windows/${OCULUS_ARCH}/${OCULUS_BUILD_TYPE}/VS2017/LibOVR.lib")
```
bjornbytes, line 227: Just curious about this change: It seems like either the `link_directories` or the absolute path are necessary to get this to work, but not both.  I do like the absolute path better, maybe link_directories could get removed?  Or is it needed for the dll?  Long term this should really be an imported target like the other 3rd party libs.

CMakeLists.txt
```diff
@@ -239,7 +238,7 @@ endif()
 
 # Pico Native SDK (1.3.3)
 if(LOVR_ENABLE_HEADSET AND LOVR_USE_PICO)
-  set(LOVR_PICO_PATH "${CMAKE_CURRENT_SOURCE_DIR}/deps/pico" CACHE STRING "The path to the Pico SDK folder (unzipped aar)")
+  set(LOVR_PICO_PATH "${CMAKE_CURRENT_SOURCE_DIR}/deps/pico/aar" CACHE STRING "The path to the Pico SDK folder (unzipped aar)")
```
bjornbytes, line 241: Description is now incorrect -- `LOVR_PICO_PATH` is no longer an unzipped aar.

CMakeLists.txt
```diff
@@ -309,10 +308,11 @@ endif()
 # Plugins
 set(LOVR 1)
 link_libraries(${LOVR_LUA})
-file(GLOB LOVR_PLUGINS plugins/*)
+file(GLOB LOVR_PLUGINS ${CMAKE_SOURCE_DIR}/plugins/*)
+message("Using these Lovr Plugins: ${LOVR_PLUGINS}")
```
bjornbytes, line 312: Personally I don't think this provides useful-enough information to warrant a message, but if it stays please spell as "LÖVR plugins".

CMakeLists.txt
```diff
@@ -337,14 +337,18 @@ if(LOVR_USE_PICO)
 endif()
 
 if(LOVR_BUILD_SHARED)
+  list(REMOVE_ITEM LOVR_SRC src/main.c)
```
bjornbytes, line 340: On Android LÖVR is built as a shared library and `os_android.c` calls `main`, requiring this file to be compiled.  Am I missing something?

CMakeLists.txt
bjornbytes: Another thought on naming after reading further: Maybe `LOVR_VISIBILITY` that defaults to `hidden` so it can be directly plugged into `C_VISIBILITY_PRESET`?

CMakeLists.txt
```diff
@@ -337,14 +337,18 @@ if(LOVR_USE_PICO)
...
   add_library(lovr SHARED ${LOVR_SRC})
+  target_compile_definitions(lovr PRIVATE LOVR_BUILDING_SHARED)
```
bjornbytes, line 342: Same-ish comment as above -- why is the file getting omitted *and* a define getting added that accomplishes the same?  Maybe fallout from a previous change?

CMakeLists.txt
bjornbytes: If desktop platforms need main omitted (which seems sensible), then seems like `LOVR_BUILD_SHARED` is not flexible enough to express "should compile `main`".  Android should either not use/respect `LOVR_BUILD_SHARED` or there needs to be a separate `LOVR_NO_MAIN` for desktop platforms to use.

Tuprules.tup
```diff
@@ -17,6 +17,7 @@ CFLAGS += -Wall -Wextra
 CFLAGS += -Wno-unused-parameter
 CFLAGS_glad.c += -Wno-pedantic
 CFLAGS_os_android.c += -Wno-format-pedantic
+CFLAGS_os_macos.c += -xobjective-c
```
bjornbytes, line 20: :heart:

src/core/os_glfw.h
```diff
@@ -305,6 +307,11 @@ void os_on_quit(fn_quit* callback) {
 
 void os_on_focus(fn_focus* callback) {
   glfwState.onWindowFocus = callback;
+  if (callback && glfwState.window) {
+    // make sure whoever sets this callback also gets its initial state
+    int focused = glfwGetWindowAttrib(glfwState.window, GLFW_FOCUSED);
+    callback(focused);
+  }
```
bjornbytes, line 314: The changes here seem really surprising to me.  I don't expect registering an event handler to also fire the event.  What's the motivation here?  As a user of the VR simulator, you'd like `lovr.focus` to get fired at startup?  That is a reasonable request.  Could it be accomplished by adding `os_window_is_focused` and having `desktop_init` do `onFocus(os_window_is_focused());`?

src/core/os_macos.c
```diff
@@ -53,11 +57,38 @@ void os_sleep(double seconds) {
 }
 
 void os_request_permission(os_permission permission) {
-  //
+  if (permission == OS_PERMISSION_AUDIO_CAPTURE) {
+    // If on old OS without permissions check, permission is implicitly granted
+    if(![AVCaptureDevice respondsToSelector:@selector(authorizationStatusForMediaType:)]) {
+      if(state.onPermissionEvent) state.onPermissionEvent(permission, true);
```
bjornbytes, line 63: Should this have a `return`?  Is it supposed to run the `switch` below?

src/core/os_macos.c
```diff
@@ -53,11 +57,38 @@ void os_sleep(double seconds) {
...
+        ];
+        return;
+      case AVAuthorizationStatusDenied:
+        if(state.onPermissionEvent) state.onPermissionEvent(permission, false);
+        return;
+      case AVAuthorizationStatusRestricted:
+        if(state.onPermissionEvent) state.onPermissionEvent(permission, false);
+        return;
+    }
+  }
```
bjornbytes, line 87: Nits here:

- Would rather use `break` instead of `return` when nothing is below.
- `if (` should have a space before the opening paren.

src/core/os_macos.c
```diff
@@ -9,15 +9,19 @@
...
 #include <time.h>
 #include <pwd.h>
+#import <AVFoundation/AVFoundation.h>
 
 #include "os_glfw.h"
 
-static uint64_t frequency;
+static struct {
+  uint64_t frequency;
+  fn_permission* onPermissionEvent;
```
bjornbytes, line 18: Maybe `os_destroy` should `memset(&state, 0, sizeof(state))` so the permission callback doesn't leak across restarts

bjornbytes:
Any luck with AAudio capture?

CMakeLists.txt
nevyn: I like the latter option. Implemented.

CMakeLists.txt
nevyn: I... honestly have no idea. lemme boot into windows and see if I can just remove the change.

nevyn: Removing the absolute path in set doesn't work:

```  
LINK Pass 1: command "C:\PROGRA~2\MICROS~1\2019\COMMUN~1\VC\Tools\MSVC\1426~1.288\bin\Hostx64\x64\link.exe /nologo CMakeFiles\Alloverse.dir\src\allonet_loader.c.obj /out:Alloverse.exe /implib:Alloverse.lib /pdb:Alloverse.pdb /version:0.0 /machine:x64 /debug /INCREMENTAL /subsystem:console deps\lovr\lovr.lib deps\allonet\allonet.lib deps\lovr\luajit\src\luajit.lib deps\lovr\glfw\src\glfw3dll.lib deps\lovr\msdfgen\msdfgen.lib deps\lovr\ode\ode.lib ..\..\..\deps\lovr\deps\openvr\bin\win64\openvr_api64.lib LibOVR.lib deps\allonet\lib\enet\enet.lib deps\allonet\lib\cJSON\libcjson_utils.lib deps\allonet\lib\cJSON\libcjson.lib deps\allonet\lib\opus\opus.lib deps\allonet\lib\tinycthread\tinycthread.lib ws2_32.lib winmm.lib kernel32.lib user32.lib gdi32.lib winspool.lib shell32.lib ole32.lib oleaut32.lib uuid.lib comdlg32.lib advapi32.lib /MANIFEST /MANIFESTFILE:CMakeFiles\Alloverse.dir/intermediate.manifest CMakeFiles\Alloverse.dir/manifest.res" failed (exit code 1104) with the following output:
C:\Users\nevyn\Dev\allovisor-lovr\out\build\x64-Debug\LINK : fatal error LNK1104: cannot open file 'LibOVR.lib'```

 .. looks like most the stuff is using absolute addressing anyway, so might as well do that here too. Removing the link_directories works fine. Doing that and submitting patch.

File: CMakeLists.txt (hunk 11 above)
Comment (nevyn) [reply CMakeLists.txt:312]: sorry, that was a leftover debug print. gone now.

File: CMakeLists.txt (hunk 13 above)
Comment (nevyn) [reply CMakeLists.txt:342]: That's because you're defining a WinMain in os_win32.c which needs to be ifdef'd out so it doesn't reference main when trying to build without a main

File: CMakeLists.txt (hunk 12 above)
Comment (nevyn) [reply CMakeLists.txt:340]: ugh you're right. how did this work for me? It's not so much desktop platforms, but rather standalone desktop apps that use lovr as a library and have their own main. having a main in a shared library is really weird imho but if that's what android requires, that's that. I could define a LOVR_NO_MAIN, that's an okay workaround...

Comment (nevyn) [reply CMakeLists.txt]: I've added a commit with a proposed solution. I used LOVR_USE_MAIN instead of NO_MAIN to avoid double negation. It's not super pretty, and I'm inferring some stuff. If you don't like it I can also make it an always-on option and just override it from alloverse's cmake. lmk what you think.

File: src/core/os_glfw.h (hunk 15 above)
Comment (nevyn) [reply src/core/os_glfw.h:314]: Since there is no getter for whether the app is focused, I'm firing the handler so so that the game always knows what the focus state is. In Alloverse, I'm capping the framerate when backgrounded; but without this event, if you defocus before the app fully launches, it won't be capped properly. 

Maybe a better solution is to include an actual getter, e g lovr.system.isFocused(), and lovr.focus event is just used to tell you that that bool has changed. What do you think?

File: src/core/os_macos.c (hunk 16 above)
Comment (nevyn) [reply src/core/os_macos.c:63]: It should have a return. Thanks!

File: src/core/os_macos.c (hunk 18 above)
Comment (nevyn) [reply src/core/os_macos.c:18]: good call!

**Comment by nevyn** [comment]
@bjornbytes
> Any luck with AAudio capture?

no :/ I didn't respond in time so he closed the issue too. I'm going to investigate it some more now while waiting for your next review.

All comments above addressed as much as I could! Hope things are looking better now.

File: src/core/os_glfw.h (hunk 15 above)
Comment (bjornbytes) [reply src/core/os_glfw.h:314]: - To clarify something, `lovr.focus` does not have anything to do with the window focus.  It's the focus of the headset session.  If the headset module is disabled, `lovr.focus` should never get called.  However, the simulator does use window focus to emulate headset session focus.
- A `lovr.headset.isFocused` would be a welcome change, but might take a lot of work to wrangle all the drivers.
- An `os_window_is_focused` that does `glfwGetWindowAttrib` would be another good change and may be useful for pieces of this.
- A `lovr.system.isWindowFocused` thing would also be okay, but it sort of opens the floodgates and creates an expectation of exposing the entire window API in `lovr.system`, adding all of the necessary events, etc.  Might be good to hold off.
- A small change that seems like it's in the spirit of this one would be to try to get headset drivers to emit focus events when they're initialized (this was the idea I posted above).  If `lovr.headset.isFocused` is added then this can be handled in one place at a higher level, but in the meantime some of the headset drivers could start to do it.

File: CMakeLists.txt (hunk 13 above)
Comment (bjornbytes) [reply CMakeLists.txt]: I think I was confused and thought `int main` was getting ifdef'd out instead of the thing in os_win32.

File: CMakeLists.txt [hunk 19]
```diff
@@ -70,6 +73,10 @@ elseif(UNIX)
   endif()
 endif()
 
+if((LOVR_BUILD_SHARED AND NOT ANDROID) OR LOVR_USE_PICO)
+  set(LOVR_USE_MAIN OFF)
+endif()
```
bjornbytes, line 78: Just got another idea for this -- instead of the `LOVR_USE_MAIN` option, why not use `LOVR_BUILD_EXE`?

```cmake
if(NOT ANDROID AND LOVR_BUILD_SHARED)
  set(LOVR_BUILD_EXE OFF)
endif()

# later ...

if(LOVR_BUILD_EXE AND NOT LOVR_USE_PICO)
  list(APPEND LOVR_SRC src/main.c)
else()
  target_compile_definitions(lovr LOVR_OMIT_MAIN) # specifically for win32 WinMain
endif()
```

src/core/os_glfw.h
nevyn: I see. Okay, I'll pull this change out for now and we can discuss alternatives separately. Perhaps I should just use https://github.com/Papaew/lovr-window .

CMakeLists.txt
nevyn: yeah that's nice, less options. implemented.

nevyn:
@bjornbytes I believe all  your comments have been addressed and I'm ready for re-re-review :)

nevyn:
@bjornbytes Hey! Anything more you need from me on this one? We've merged it in our fork and started using it in production, seems to be working fine across platforms. LMK if you want me to split it up or do something else with it to make it easier to merge.

bjornbytes:
Hey, I got a bit distracted.  How viable do you think a squash would be here?  1 commit would probably be too squished, but there are a few revert/fixup/typo commits that might be good to combine.  If it's too convoluted, I can just merge as is, no worries.

nevyn:
I'll take a look asap!

nevyn:
@bjornbytes rebase and squash done. Should be much cleaner now. Closing this in favor of https://github.com/bjornbytes/lovr/pull/436 (didn't want to force-push, so I could compare on github.com and keep comments etc)

PR #399, opened 2021-04-01
SteamAudio and Phonon linkage for mac

mcclure:
These do not currently work, so a debug printf is added to spatializer_phonon.c to help in debugging.

Notice the awkward extra step for Oculus Audio.


CMakeLists.txt
```diff
@@ -266,8 +266,9 @@ if(LOVR_USE_STEAM_AUDIO)
   elseif(WIN32)
     set_target_properties(Phonon PROPERTIES IMPORTED_IMPLIB "${LOVR_STEAM_AUDIO_PATH}/lib/Windows/x64/phonon.lib")
     set_target_properties(Phonon PROPERTIES IMPORTED_LOCATION "${LOVR_STEAM_AUDIO_PATH}/bin/Windows/x64/phonon.dll")
-  elseif(APPLE) # This should be possible, there is a dylib in OSX/
-    message(FATAL_ERROR "LOVR_USE_STEAM_AUDIO is not currently supported on Apple platforms")
+  elseif(APPLE)
+    set_target_properties(Phonon PROPERTIES IMPORTED_LOCATION "${LOVR_STEAM_AUDIO_PATH}/lib/OSX/libphonon.dylib"
+                                            BUILD_RPATH "${LOVR_STEAM_AUDIO_PATH}/lib/OSX")
```
bjornbytes, line 271: I would prefer to `move_lib` the library and fix the rpath on regular macOS builds to be `@executable_path` (basically move `MACOSX_RPATH`, `BUILD_WITH_INSTALL_RPATH`, and `INSTALL_RPATH` properties outside of `if (LOVR_BUILD_BUNDLE)`).

It is a little bit more involved because `move_lib` needs to work properly on non-bundled builds.  At that point maybe we start dropping everything in a `bin` dir like the way linux CMake builds are currently set up.

mcclure: That approach would definitely sidestep the RPATH issues I am having with both phonon and steam.

bjornbytes: I made part of this change -- moved the rpath target properties to a place where they will always apply on macOS (untested, I gotta sleep).  The other part will be to modify `move_lib` so it is always called and writes to the correct place for bundled+nonbundled builds.

mcclure:
I figured out how to fix this PR without adding move_lib to the non-bundle builds. However, I think probably we want the move_lib approach anyway because it will work better with plugins. I will wait on that.
I'm back on Windows today so there's no urgency for me on this right now.

bjornbytes:
The rpath is now fixed and move_lib happens on all macOS builds.  So the stuff here will probably work better now, maybe get simplified a bit too.  Just let me know!

mcclure:
I'll take a pass at it, probably not today tho

mcclure:
With new commits, Oculus Audio works— but Steam Audio doesn't, and I don't know why.
I went to Stack Overflow with a [description of the new problem](https://stackoverflow.com/questions/67258171/target-soname-file-points-to-enclosing-directory-rather-than-file-but-only-on) to ask for help.

mcclure:
I literally found and filed two separate CMake bugs in the process of writing this but this patch is now ready to commit, it works anyway. I can squash into a single commit if you'd prefer.

bjornbytes:
Uh, wow, CMake.  Squash would be preferred if it reduces the number of un-buildable commits here

mcclure:
Refiled as #426 so I could squash without losing the mac-spatializer-test-2 branch

PR #400, opened 2021-04-02
Encapsulate "uses spatializer?" nature in lovrSourceUsesSpatializer

mcclure:
isSpatial() was removed because now the relevant concept is not "is it spatial?" but rather "does it have spatial effects?". Since all effects are currently spatial, this is the same as asking "does it have effects?".

However, the knowledge "no effects -> no spatializer use" is "magic", it's not obvious from the names and it isn't commented, and this knowledge is applied three times in audio.c. This patch encapsulates the knowledge into a UsesSpatializer() function and comments as appropriate.

Also, remove accessor prototype for old IsSpatial() function, which currently corresponds to no function.

### DIFF
--- a/src/modules/audio/audio.c
+++ b/src/modules/audio/audio.c
@@ -110,7 +110,7 @@ static void onPlayback(ma_device* device, void* out, const void* in, uint32_t co
       }
 
       // Read and convert raw frames until there's BUFFER_SIZE converted frames
-      uint32_t channels = source->effects == EFFECT_NONE ? 2 : 1;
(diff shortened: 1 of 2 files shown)


mcclure:
Wait, there is a typo.

mcclure:
Fixed, initial push was missing audio.h change.

bjornbytes:
I think this makes things easier to understand, thanks.  I am getting the sense that there are deficiencies of the design of the effect system, but I don't immediately know best way to rearchitect, and the way it exists now is at least pretty lightweight and allows some degree of control over the features of the spatializers.  Leaning towards getting it into people's hands and developing better understanding of how to improve it over time.

PR #402, opened 2021-04-06
WIP: Make bin directory AppImage ready

Wallbraker:
This at least skips the manual step of moving the files over into the directory. Not sure if the `lovr.ico` is supported.


bjornbytes:
Could these resource files exist directly inside `src/resources` instead of inside a subfolder?

bjornbytes:
I'm wondering if moving the appimage-related resources (and, later, potentially building the appimage itself) should be gated by the existing `LOVR_BUILD_BUNDLE` option.  It currently causes macOS to create a .app bundle instead of a lone executable, and maybe it would make sense for linux builds to use the same flow.

Wallbraker:
> Could these resource files exist directly inside `src/resources` instead of inside a subfolder?

Sure if you prefer that, I put them to organise them and make it clear what they were used for.

> I'm wondering if moving the appimage-related resources (and, later, potentially building the appimage itself) should be gated by the existing `LOVR_BUILD_BUNDLE` option. It currently causes macOS to create a .app bundle instead of a lone executable, and maybe it would make sense for linux builds to use the same flow.

That makes sense.

Wallbraker:
AppImage didn't like the `lovr.ico` file, so that's a no go.

src/resources/app-image/AppRun
```diff
@@ -0,0 +1,14 @@
...
+SELF_DIR="$(dirname "$(readlink -f "$0")")"
+LIBS_PATH="$SELF_DIR"
+
+if [ -z "$LD_LIBRARY_PATH" ]; then
+  LD_LIBRARY_PATH="$LIBS_PATH"
+else
+  LD_LIBRARY_PATH="$LIBS_PATH:$LD_LIBRARY_PATH"
+fi
+
+export LD_LIBRARY_PATH
```
bjornbytes, line 12: The rpath is currently being set to `$ORIGIN` which should allow libraries to be loaded from next to the executable, why was `LD_LIBRARY_PATH` also needed?

Wallbraker: Oh you are right, that works. Thanks.

bjornbytes:
Ok, I would still like to lift the files into the toplevel `resources` folder before merge.  I do think organizing the resources folder could be good but I'd prefer to tackle it in a holistic way.

bjornbytes:
It's great to finally have official Linux builds, thank you!

Wallbraker:
Thanks for merging :)

Issue #403, opened 2021-04-07
setGeometry() can crash if called "late" / while a sound is playing 


mcclure:
**Repro**
Check out the break-phonon-crash test here
https://github.com/mcclure/lovr-docs/tree/break-phonon-crash
And run showcase/Break

**Symptom**
This Break fork will once a second call `lovr.audio.setGeometry()` with a new mesh. Eventually, it will crash.
In my testing, I can call setGeometry() with different data as many times as I want in lovr.load(), but if I call it after the program has been running for a bit I often get a crash. Based on this I suspect the problem is caused by setGeometry while a sound is already playing.

In one crash, I attached MSVC debugger. The crash was on spatializer_phonon.c:245 `  IPLDirectSoundPath path = phonon_iplGetDirectSoundPath(state.environment, listener, forward, up, iplSource, radius, rays, occlusion, volumetric);`. I noticed "state.environment" was NULL.

mcclure:
This was on dev d2a9e230bf9e6

PR #405, opened 2021-04-12
Hold lock when changing geometry (prevents crash)

mcclure:
See issue #403


mcclure:
Note we might want to do something a LITTLE smarter later where the lock is deferred to the spatializer, then again maybe not.

Issue #411, opened 2021-04-15
Allow using alpha with the hexcode variant of lovr.graphics.setColor


bjornbytes:
Right now you can do `lovr.graphics.setColor(0xff00ff)`, but it is not possible to add an extra alpha argument after that.  This can just barely get squeezed in as a new variant since there is not currently a 2-argument version of setColor.  This would be a change to `luax_readcolor` in `api.c`.

PR #412, opened 2021-04-15
Allow using alpha with hexcode when setting color

jmiskovic:
Implements #411 proposal. I needed this variant before so I think this is good extension to setColor.

When setColor() is called without arguments it will throw `bad argument #1 to 'setColor' (number expected, got no value)"`. It does not convey all possible combinations of arguments, but I think this is acceptable to keep the error message concise.

### DIFF
--- a/src/api/api.c
+++ b/src/api/api.c
@@ -400,14 +400,12 @@ void luax_readcolor(lua_State* L, int index, Color* color) {
     color->g = luax_checkfloat(L, index + 1);
     color->b = luax_checkfloat(L, index + 2);
     color->a = luax_optfloat(L, index + 3, 1.);
-  } else if (lua_gettop(L) == index) {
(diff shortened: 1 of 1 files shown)


jmiskovic:
I've tested all functions that use `luax_readcolor`. All work (lg.newMaterial, lg.setBackgroundColor, lg.setColor, material.setColor) except the lg.clear. That one is already complex enough because it accepts various combinations of numbers and booleans.

Also I'm not sure what does it mean to have transparency in background color? Does it ever affect rendering results?

bjornbytes:
For background color / lg.clear, having an alpha doesn't blend with previous contents, it just "assigns" to alpha channel.  You might want a transparent background if you're rendering an overlay.  clear can't really support this variant due to the ambiguity.

Implementation looks good.  Thanks for implementing this!

jmiskovic:
I'll also update the docs.

PR #422, opened 2021-04-22
Oculus spatializer reverb effect support (attempt)

mcclure:
This adds support for setGeometry and the "reverb" effect to the oculus spatializer. There are multiple problems with this PR. I am posting it as a "request for comments".

The first problem with this PR is it does not work. In my testing with the Break demo (will post an issue shortly with the updated Break demo) enabling setGeometry with Oculus Spatializer and this patch has no effect whatsoever. Simply put the reverb is not audible, and I don't know why.

The second problem is that this patch changes the API. setGeometry now takes an additional "geometry mode" enum, which can be "disable", "box", or "mesh". This is for three reasons:
* There is a difficulty with the Oculus spatializer and the current Lovr API. Lovr has "effects" you can enable and disable on a on a per-sound basis, which Steam Audio supports. Oculus however does not have per-source effects. Oculus has a global effects setting and you can only choose whether or not to have the Oculus spatializer manage your sounds. Therefore a method is needed in Oculus spatializer for turning reverb on and off.
* At present, it is possible to upload a geometry to Steam, but it is not possible to "erase" that geometry. That is awkward. You can turn reverb on and off but you can't get back to the pre-setGeometry state. The "disable" mode works with both Oculus and Steam in this patch.
* Oculus has a "box model" option. For testing purposes, it would be nice to be able to switch back and forth between mesh and box models. We discussed adding a setBoxModel() or an AudioMesh(), but we don't have that API yet. As a stopgap, requesting "box" in the Oculus spatializer with this patch computes the AABB of the submitted mesh and uses that as the box. (Requesting box mode in the steam spatializer with this patch gives you normal mode).

I don't know if this API change is ideal, but I think supporting "interpret mesh as box" would be a helpful option to offer *somewhere* in the API eventually.

I don't know what to do to get this to work. The Oculus Audio API is very very scantily documented. There are lots of enums and I have tried setting some of them but maybe I missed something. I have submitted a request for more docs to Oculus.

### DIFF
--- a/src/api/api.h
+++ b/src/api/api.h
@@ -110,6 +110,7 @@ int luax_setconf(struct lua_State* L);
 void luax_setmainthread(struct lua_State* L);
 void luax_atexit(struct lua_State* L, void (*destructor)(void));
 void luax_readcolor(struct lua_State* L, int index, struct Color* color);
+// Note: Vertices are packed as 3-vectors, not 4-vectors.
 int luax_readmesh(struct lua_State* L, int index, float** vertices, uint32_t* vertexCount, uint32_t** indices, uint32_t* indexCount, bool* shouldFree);
 
 // Module helpers
--- a/src/api/l_audio.c
+++ b/src/api/l_audio.c
@@ -33,6 +33,13 @@ StringEntry lovrAudioMaterial[] = {
   { 0 }
 };
 
+StringEntry lovrGeometryMode[] = {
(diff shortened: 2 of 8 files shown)


bjornbytes:
Awesome, it's really cool to have the pieces of this worked out even if it's not 100% complete yet.  I'll be posting more once I have time to give this a good look.

bjornbytes:
Archiving this (and the other 2) due to inactivity

Issue #427, opened 2021-04-28
All-in-one Plugin Support


bjornbytes:
Lua's stock loaders offer 2 ways of loading native libraries.  For `require 'a.b.c'`, it will try:

- `a/b/c.so`, `a/b/c.dll`, `/usr/local/a/b/c/init.so` etc.
- `a.so`, `a.dll`, `/usr/local/a/init.so` etc. (all-in-one)

In both cases this is controlled by `package.cpath` and the symbol loaded will be `luaopen_a_b_c`.  The purpose of all-in-one is to package multiple Lua libraries in one native library file.  luasocket notably uses this approach to package e.g. `luasocket.ftp`, `luasocket.http`, `luasocket.dns` modules into a single dll file.

The plugin loader that `lovr.filesystem` uses currently supports the first type of library, but not the second.  This issue is about adding the second.

When confronted with `require 'a.b.c'`, the loader should, in addition to trying e.g. `<exepath>/a/b/c.dll`, it would also try `<exepath>/a.dll`.  That's it.

It isn't yet clear if this should be a separate loader or stuffed into the existing one.  Will likely need to just feel it out during implementation.  Seems like latter would be simpler/reduce duplication.

brainrom:
LOVR can load `socket`, `socket.ftp`, `socket.http` from `socket.so`, using `luaopen_socket`, `luaopen_socket_ftp`, `luaopen_socket_http` symbols respectively, but can't load `mime` from `socket.so` (and `socket` from `libluasocket.so`, where CMake builds module by default, too!), even if `luaopen_mime` symbol presents in `socket.so`, workaround is symlinking/copying `mime.so -> socket.so` . Maybe, when `require` is called. LOVR should search symbols through all available `.so`/`.dll` inside lovr's executable directory (`build/bin` for example) and throw an error only in case of unsuccess.

bjornbytes:
> but instead of it, lovr search it in `pwd`. So, I think it's a bug.

Where are you seeing that native libraries are searched for in the working directory?

In l_filesystem, the base path is `lovrFilesystemGetExecutablePath` (`/path/to/lovr`).  The executable gets stripped off -> `/path/to/`.  Finally the module and extension is appended -> `/path/to/module.so`.

brainrom:
Here :cry:  :
```
main.lua:2: module 'socket.http' not found:
        no field package.preload['socket.http']
        no file './socket/http.lua'
[...]

PR #429, opened 2021-04-30
All-in-one plugins lua loader

brainrom:
Adds #427 functionality. Almost no code duplication.

### DIFF
--- a/src/api/l_filesystem.c
+++ b/src/api/l_filesystem.c
@@ -411,7 +411,7 @@ static int luaLoader(lua_State* L) {
   return 0;
 }
 
-static int libLoader(lua_State* L) {
+static int libLoaderCommon(lua_State* L, bool allInOneFlag) {
 #ifdef _WIN32
   const char* extension = ".dll";
 #else
@@ -454,6 +454,7 @@ static int libLoader(lua_State* L) {
 #endif
 
   for (const char* m = module; *m && length < sizeof(path); m++, length++) {
+    if (allInOneFlag && *m == '.') break;
     *p++ = *m == '.' ? LOVR_PATH_SEP : *m;
   }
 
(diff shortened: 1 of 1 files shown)


src/api/l_filesystem.c
```diff
@@ -484,6 +485,18 @@ static int libLoader(lua_State* L) {
...
+static int libLoader(lua_State* L) {
+  const char* module = lua_tostring(L, 1);
+  const bool allInOneFlag = false;
+  return(libLoaderCommon(L, allInOneFlag));
+}
+
+static int libLoaderAllInOne(lua_State* L) {
+  const char* module = lua_tostring(L, 1);
+  const bool allInOneFlag = true;
+  return(libLoaderCommon(L, allInOneFlag));
```
bjornbytes, line 497: Small requests:

- Could `allInOneFlag` just be named `allInOne`?
- `const bool` should just be `bool` (There's nothing _wrong_ with const there, it is just a little unconventional for the codebase style)
- Please remove parentheses around returns: `return libLoaderCommon(...);`)

bjornbytes:
This is a good way to accomplish this, looks good!  So with this change, does your luasocket plugin work properly?  Does it still need the symlinks?

Could you rebase the commits on this branch so it has only 1 commit?

In the future I might look into using closures to hold the `allInOne` flag, but that's just a fancy organization thing.

Thanks for the PR.

src/api/l_filesystem.c
brainrom: 1. Yes. But IMHO `allInOneFlag` more suitable, because it just switches function behavior and don't provide any additional information.
2. OK
3. OK

brainrom:
Unfortunately, it's still need symlinks. I'm still thinking about loader, that searches for requested symbols in every single `.so`/`.dll` next to lovr executable, but it doesn't sounds great. There is should be more fancy solution. Maybe just split (with no excess size) it to `socket.so`, `ltn12.so` and `mime.so`. Or even just copy `ltn12.lua` and `mime.lua` as is, because it doesn't use C functions. Also there is a problems with Windows compatibility in my luasocket binding. On Linux everything is awesome, on Windows it doesn't wants to import module. I'll install MSVC on VM and fix it. Cross-compiling LOVR Lin->Win using MinGW works with a bit of magic, but not suitable at all.

PR #436, opened 2021-05-31
Submo's for VR SDKs, and other changes (rebased)

nevyn:
This is https://github.com/bjornbytes/lovr/pull/398 but with squashed fixups and rebased linearly onto latest dev.


nevyn:
@bjornbytes hmm. Perhaps I should try to disable LFS for the submos and push again so we're not at mercy of bandwidth limitation, eh?

nevyn:
@bjornbytes Done, git LFS removed from `oculus-mobile`. It was only used for VrSamples anyway, which is not something that is needed for a submo that provides CI support.

bjornbytes:
Awesome, thanks!

nevyn:
@bjornbytes omg that was fast!! aaahh what a relief, this is finally done!! thank you!! for the first time in years, Alloverse can use pure upstream lovr as-is ;D

PR #437, opened 2021-06-04
Tupfile.lua

bjornbytes:
This converts the tup build system to use tup's Lua parser.  It's a pretty big rework, but has a lot of benefits:

- Currently, to build using tup, you have to first build all the dependencies with CMake.  This sucks, at that point why not just use CMake.  On this branch, dependencies can be built with tup, so it's possible to build LÖVR without any CMake for the first time.
- Lua is a more expressive configuration syntax.  The build configuration is provided in a single Lua table at the top of Tupfile.lua, and is more expressive than the existing `tup.config` (though that may still be useful for variants).  We can also have more friendly asserts, and a lot of the rules are easier to write using Lua.
- There is now only 1 tup-related file instead of 3.
- Add a flag called `supercharge` for doing lto.
- Allow cross compiling lovr to windows using mingw.

It can build on all platforms.  There are some things that haven't been worked out yet though:

- [ ] LuaJIT has a complicated build system and I haven't been able to build that with tup yet.
- [ ] The OpenXR loader has a complicated build system and I haven't been able to build that with tup yet.


bjornbytes:
I don't think there's a reason not to merge this.  This doesn't impact CMake and I think I'm the only one using tup.  There's a few snags to be resolved here (LuaJIT, OpenXR, tup variants) but I like this structure a lot more than the current one and would like to start using and maintaining it.

PR #442, opened 2021-06-10
Texture:replacePixels: Add support for string and lightuserdata

Voxar:
Moves pixels into a texture without copying the c buffer. 

Not sure how to handle mipmaps. If the goal is to copy pixels quickly one probably doesn't want to spend time generating mipmaps.

The use case is video textures.


bjornbytes:
I'm not going to be merging this in its current form because it duplicates too much code.  I think it's mostly the fault of the current replacePixels implementation.  This will be on my mind on the gpu branch and I'll make sure there's a way to get raw pointer data into Textures there.

PR #445, opened 2021-06-15
MinGW support

brainrom:
Now LOVR can be built (and even cross-compiled Linux->Windows) via MinGW GCC compiler too. Because Windows and MSVC are not case-sensetive, os_win32.c changes do not affect to classical Windows build process.


bjornbytes:
Thanks!  I've also got MingW support in #437, so it should be universal once that gets merged.  Exciting.

PR #450, opened 2021-06-24
Fix compile on old macos

mcclure:
Our permissions code currently uses AVCaptureDevice, which only became available on MacOS with 10.14. The code attempts to deal with this with a [AVCaptureDevice respondsToSelector] check, which will work at runtime, but this check will never run because the code will fail to compile (I tested on 10.13). This change #ifs out the section that can't compile on 10.13. (I set it to throw an assert if the check is compiled out but found at runtime to be needed, on the assumption failing to run is better than running with incorrect permissions.)

There are various other ways we could do this, like late-binding tricks. But in tests this works.

### DIFF
--- a/src/core/os_macos.c
+++ b/src/core/os_macos.c
@@ -12,6 +12,7 @@
 #import <AVFoundation/AVFoundation.h>
 
 #include "os_glfw.h"
+#include "util.h"
 
 static struct {
   uint64_t frequency;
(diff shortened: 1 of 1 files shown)


bjornbytes:
I'd like to change the assert to `lovrThrow("Unreachable");` to match other places, but I think it is too minor to request changes from you so I will do it after merging.  Thanks for the fix.

mcclure:
Maybe a good idea to also throw a #warning in there now that I think about it. `#warning Compiling without permissions code. Do not distribute this build.` Something like that.

PR #451, opened 2021-06-25
Fix copy so that plugins can work on mac

mcclure:
Tried to build on mac with cjson plugin. Got this error.

Screen Shot 2021-06-24 at 4 01 54 PM

Bjorn suggested the fix in this patch based on the UNIX code. I tried it. It worked.


bjornbytes:
:purple_heart:

Issue #303, opened 2020-10-27
Vulkan


bjornbytes:
This issue tracks the work to add the ability to render graphics with Vulkan and WebGPU, in addition to or instead of OpenGL.

You can follow along on the [`gpu`](https://github.com/bjornbytes/lovr/compare/dev...gpu) branch.

New Features
---

- Explicit Pass objects
- Headless rendering without a window
- ffi pointers to GPU memory
- Generic Buffer objects (#88 #208 #209)
- Indirect Rendering and Dispatch
- No global graphics pipeline state
- More pipeline states (depth nudge, depth clamp, more stencil control)
- Raw GPU IO operations like copies, blits, asynchronous readbacks
- Improved validation and error handling
- Support for 6 camera views for single pass cubemaps
- Compute shader vertex animation
- Occlusion queries
- Texture Views
- BC6/BC7 texture support, KTX2 support

Removed Features
---

- Canvas is no longer an object, instead you pass in canvas info as a table
- No more linestrip/trianglestrip/trianglefan draw modes (use an index buffer)
- Removed 'line' DrawStyle
- Removed `lovr.graphics.arc`
- ShaderBlock and Mesh are replaced by Buffer

bjornbytes:
Ideas for exposing multithreaded rendering with varying levels of difficulty/control.

```lua
-- Draw a 10x10 grid of cubes, single-threaded
for i = 1, 10 do
  for j = 1, 10 do
    pass:cube('fill', i, j, -1, .5)
  end
end

-- Draw a 10x10 grid of cubes, in parallel
-- Internally it creates 10 jobs, each with a batch,
-- waits for them to finish recording, and submits
-- them all, in a well-defined order based on id
pass:run(function(id, batch)
  for i = 1, 10 do
    batch:cube('fill', i, id, -1, .5)
  end
end, 10)

-- This is the same as above, but the batches are
-- managed by the user.  You can have full control over
-- when and how they run, where waiting happens, what
-- order they're submitted in, and when submission takes
-- place.  Allows you to do other work while the batches
-- are recording or change submission order or fill out
-- multiple batches in a single task if that is better for you.
-- You could even fill out batches for different passes all
-- at the same time and then run the passes all at once when
-- the entire workload is finished.
-- This uses the job system which isn't designed yet, so
-- things are a bit fuzzy on how that would work.  You could
-- also imagine this pushes the batches onto Channels for a
-- thread pool to consume.
local batches = {}
local jobs = {}
[...]

PR #455, opened 2021-07-06
Image:setBlob() feature

brainrom:
This is useful for fast reading textures from external plugins, while LOVR doesn't have a stable C API to return Image directly. For example, in video player plugin (FFMpeg-based) that I'm currently working on. Example usage:
```
rawblob=lovr.data.newBlob(ffmpeg.getframe())
img:setBlob(rawblob)
```

### DIFF
--- a/src/api/l_data_image.c
+++ b/src/api/l_data_image.c
@@ -82,6 +82,13 @@ static int l_lovrImageGetBlob(lua_State* L) {
   return 1;
 }
 
+static int l_lovrImageSetBlob(lua_State* L) {
+  Image* image = luax_checktype(L, 1, Image);
+  Blob* blob = luax_checktype(L, 2, Blob);
+  image->blob=blob;
+  return 0;
+}
+
 const luaL_Reg lovrImage[] = {
   { "encode", l_lovrImageEncode },
   { "getWidth", l_lovrImageGetWidth },
@@ -92,5 +99,6 @@ const luaL_Reg lovrImage[] = {
   { "getPixel", l_lovrImageGetPixel },
   { "setPixel", l_lovrImageSetPixel },
   { "getBlob", l_lovrImageGetBlob },
+  { "setBlob", l_lovrImageSetBlob },
   { NULL, NULL }
 };


nevyn:
This seems closely related to https://github.com/bjornbytes/lovr/pull/442

brainrom:
Hm, maybe. But now #442 looks like excessive code copy-pasting. Anyway, my video decoder plugin can work with both of videotexture options.

bjornbytes:
This is missing the necessary refcounting (the old Blob will never get freed, the new Blob might accidentally get freed), and validation (the Blob might not be big enough to hold all the pixels, causing crashes or exploit opportunities).  Something doesn't feel quite right about swapping out backing storage like this.

Maybe `newImage(w, h, format, blob)` could use the input Blob directly instead of always creating a new piece of memory and copying.  However, there are some times when you actually want that copy to take place (so that you can use your Blob for something else) and other times when you don't.  So it's not obvious that switching to 'nocopy' would be the right approach.

Depending on how ffmpeg works, maybe the ffmpeg Lua bindings could be modified to accept a lightuserdata/Blob.  That way you could read the video data directly into the Image: `ffmpeg.getFrame(image:getBlob():getPointer())`.

I am still looking at #442, there's duplication there but that's not really its fault, it is simply bumping up against some gnarly architecture problems in the existing design.

brainrom:
Okaaay, one more undocumented LOVR feature for me. Actually, `image:getBlob():getPointer())` way works.
However because of some unknown reasons I'm unable to use just `Blob* blob = luax_checktype(L, 2, Blob);`
Just adding a function with this string (even if it's not used in any place and not added to any `luaL_Reg`) breaks module loading: `module 'ffmpeg' not found` (.so file builds without any problems) and now I'm forced to use direct way without any checks. I'll do extra research later.
```
Proxy* p = lua_touserdata(L, 2);
Blob* blob = p->object;
```
Now my current goal is to make the module completely non-blocking to use it in main thread and provide `ffmpeg->Image->Texture->Shader->GPU` pipeline.

brainrom:
Now I'm faced the problem, that the `lovrTextureReplacePixels` is the bottleneck in all LOVR rendering process, FPS may decrease down to 30 when replacing at least 4 textures in one `lovr.update()` call. I'll do additional research with the #442. Maybe it's even possible to call `ffmpeg.getFrame(Texture)` which then call in C side some modified version of `lovrTextureReplacePixels` to copy frame directly to GPU. Maybe #303 is necessary to effective and fast texture update.

bjornbytes:
So for Vulkan you will be able to get a temporary GPU buffer, write data into there (via Lua, C, or FFI), and then issue an async copy.  This is probably the fastest way to get data into a Texture, and will be able to be done on any thread.  In OpenGL it's called "pixel unpack buffer" and isn't currently exposed.  So yes, I do think the upcoming Vulkan work will expose a few tools to improve this.

(btw in Vulkan there are also dedicated transfer/compute queues, where you can upload/process data while other frames are rendering.  no plans to offer that currently due to complexity, but it could get added as a future improvement).

PR #458, opened 2021-07-14
Allow get sample rate in conf.lua, read back with lovr.audio.getSampleRate()

mcclure:
This patch has been submitted before and rejected before, and that's okay, I'm just submitting an updated version so you can close the PR and there will be a record of it. :)

This patch allows setting the sample rate. The sample rate still cannot be changed once the program starts, but it can be set in lovr.conf with t.audio.sampleRate and queried with lovr.audio.getSampleRate().

I maintain this patch because sometimes in my fork I need to run at 44.1khz sample rate, but not all the time (so changing audio.h is slightly inconvenient).

### DIFF
--- a/src/api/l_audio.c
+++ b/src/api/l_audio.c
@@ -202,7 +202,7 @@ static int l_lovrAudioGetSpatializer(lua_State *L) {
   return 1;
 }
 
-static int l_lovrAudioGetSampleRate(lua_State *L) {
+static uint32_t l_lovrAudioGetSampleRate(lua_State *L) {
   lua_pushnumber(L, lovrAudioGetSampleRate());
   return 1;
 }
@@ -319,7 +319,7 @@ int luaopen_lovr_audio(lua_State* L) {
     spatializer = lua_tostring(L, -1);
     lua_pop(L, 1);
 
-    lua_getfield(L, -1, "sampleRate");
+    lua_getfield(L, -1, "samplerate");
     int userSampleRate = luaL_optnumber(L, -1, 0);
     if (userSampleRate > 0)
       sampleRate = userSampleRate;
--- a/src/api/l_audio.c
+++ b/src/api/l_audio.c
@@ -311,7 +311,7 @@ int luaopen_lovr_audio(lua_State* L) {
 
   bool start = true;
   const char *spatializer = NULL;
-  int sampleRate = DEFAULT_SAMPLE_RATE;
+  int sampleRate = 48000; // Set default here
   luax_pushconf(L);
   lua_getfield(L, -1, "audio");
   if (lua_istable(L, -1)) {
--- a/src/api/l_audio.c
+++ b/src/api/l_audio.c
@@ -202,6 +202,11 @@ static int l_lovrAudioGetSpatializer(lua_State *L) {
   return 1;
 }
 
+static int l_lovrAudioGetSampleRate(lua_State *L) {
+  lua_pushnumber(L, lovrAudioGetSampleRate());
+  return 1;
+}
+
 static int l_lovrAudioGetAbsorption(lua_State* L) {
   float absorption[3];
   lovrAudioGetAbsorption(absorption);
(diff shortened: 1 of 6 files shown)


bjornbytes:
Might be able to make this work.  Since it's in conf.lua and not setDevice, there is no need to worry about recreating data converters or spatializers, which is nice.

Can we restrict to 24khz, 44.1khz, 48khz?  Phonon is documented as only working with these rates.  Or should it be more flexible -- sample rate can be anything, but phonon will pout and fail to initialize if it doesn't support your sample rate?

Trying to remember if there was any other reason for the harcoded sample rate...

src/api/l_audio.c
```diff
@@ -305,20 +311,27 @@ int luaopen_lovr_audio(lua_State* L) {
...
   const char *spatializer = NULL;
+  int sampleRate = DEFAULT_SAMPLE_RATE;
   luax_pushconf(L);
   lua_getfield(L, -1, "audio");
   if (lua_istable(L, -1)) {
     lua_getfield(L, -1, "spatializer");
     spatializer = lua_tostring(L, -1);
     lua_pop(L, 1);
 
+    lua_getfield(L, -1, "sampleRate");
```
bjornbytes, line 322: For better or worse conf.lua fields use the Lua lowercase convention, so it should be `samplerate` for consistency

src/modules/audio/audio.h
```diff
@@ -72,6 +72,7 @@ void lovrAudioGetPose(float position[4], float orientation[4]);
 void lovrAudioSetPose(float position[4], float orientation[4]);
 bool lovrAudioSetGeometry(float* vertices, uint32_t* indices, uint32_t vertexCount, uint32_t indexCount, AudioMaterial material);
 const char* lovrAudioGetSpatializer(void);
+int lovrAudioGetSampleRate();
```
bjornbytes, line 75: Would prefer if sample rate was represented as `uint32_t` everywhere to match miniaudio and `Sound`.

src/api/l_audio.c
```diff
@@ -305,20 +311,27 @@ int luaopen_lovr_audio(lua_State* L) {
 
   bool start = true;
   const char *spatializer = NULL;
+  int sampleRate = DEFAULT_SAMPLE_RATE;
```
bjornbytes, line 314: I'd rather move the 48000 here, since no one else uses the define.

mcclure:
> Might be able to make this work. Since it's in conf.lua and not setDevice, there is no need to worry about recreating data converters or spatializers, which is nice.
> 
> Can we restrict to 24khz, 44.1khz, 48khz? Phonon is documented as only working with these rates. Or should it be more flexible -- sample rate can be anything, but phonon will pout and fail to initialize if it doesn't support your sample rate?
> 
> Trying to remember if there was any other reason for the harcoded sample rate...

At the moment, Oculus spatializer effectively requires 48k (it works at other frequencies, but reverb breaks). The way I have dealt with this (in my reverb branch) is by having Oculus spatializer print a "Oculus spatializer not compatible with this sample rate" warning at startup. Maybe it would make more sense to have Phonon print a warning and disable itself outside those three? Although I can't imagine a REASON to have a sample rate other than those three, it IS possible to use "simple" spatializer, so I don't see a reason to restrict simple spatializer because one of our three spatializers doesn't support it.

src/api/l_audio.c
mcclure: Made the change but—
I usually think it's better, rather than putting a magic number in code, to put the magic number in a constant and use the constant. What's concerning to me here specifically is this means the default will be set in an l_ file— if someone hypothetically were using the "C part" of the Lovr API standalone, they'd have no sense of what the default would be.

mcclure:
Made all changes above except the 24k/44.1k/48k restriction. I can do that one too if you're sure.

bjornbytes:
I agree that phonon should print a warning and bail.  Putting that off is fine for now, especially since it works with the default sample rate.  Consider it thrown onto the pile of phonon stuff for later.

bjornbytes:
This is great!

src/api/l_audio.c
bjornbytes: Opps, I missed this comment, but yeah that's pretty reasonable.

PR #459, opened 2021-07-20
Fix distance attenuation in Oculus spatializer

mcclure:
Oops. There's literally been no distance attenuation in the Oculus Spatializer since we merged it. This one-line fix addresses this.


bjornbytes:
yay

PR #462, opened 2021-07-23
Add new XML attributes required for Oculus

mcclure:
Oculus Store is currently requiring these XML attributes for submission.

By the way, in my fork I've broken out all these various attributes onto their own indented lines— this makes merging less of a problem if individual things have changed (like, you want to change excludeFromRecents because you're making a release build, or you want to bump your version number on the top stanza). Would you be interested in this formatting chagne upstream?


bjornbytes:
I prefer single line for now, but not by much.  Thanks for PR

PR #470, opened 2021-08-09
Find Vulkan SDK via environment or CMake variable

mcclure:
Without this I do not find vulkan.h on Windows.

The Windows installer for Vulkan sets this VULKAN_SDK variable. I don't know what happens on Linux.

Setting it this way may lead to a problem where if a new Vulkan SDK is installed, CMake does not notice until you clean rebuild.

There might also be linker flags needed, I'm not that far yet.


mcclure:
We're doing it with FindPackage instead

Issue #344, opened 2020-12-06
Vector math API should make mul() less confusing


jmiskovic:
The `mul` function in vector math API has inconsistencies which make it hard to remember and use correctly. I find it hardest part of LOVR API to remember and it results in most of my bugs due to unintended mutation.

The "mul" as function name is more about implementation (multiplying things together) and doesn't convey semantic meaning of vector operations. There are few different operations that all use the same name:

- `vec3:mul(n)` is vector scaling, semantically same as `vec3:div(1/n)` and similar to non-mutating `vec3 * n` variant
- `vec3:mul(vec3)` is component-wise multiplication (Hadamard product), I don't know if it's useful for small-dimensional vectors?
- `mat4:mul(mat4)` and `quat:mul(quat)` stacks (combines) two matrices on top of each other
- `mat4:mul(vec3)` and `quat:mul(vec3)` applies transformation to vector, without mutating mat4 / quat

The last usage is especially confusing. The rest of API (`add`, `scale`, `rotate`...) follows the convention that only object on which the operation is called gets mutated. The `mat4:mul(vec3)` will return transformed vector, but it will also mutate passed-in vector which is pretty big exception from rest of API.

I'm thinking how it could be improved at this point. Few suggestions:

1. Add function `transform()` for vec3 and vec4, which would take mat4 / quat and it would mutate only object on which it's called (it could also be called `apply()` for brevity)
2. Describe `quat:mul(vec3)` and `mat4:mul(vec3)` variants as deprecated and redirect readers to `vec3:transform()`
3. Add function `scale(n)` for vec2 / vec3 / vec4 and deprecate `mul()` and `div()` for those vector types

This would make vector operations more consistent and memorable, so that users could focus on what they are trying to accomplish. For me personally even implementing just first suggestion would be useful as I could ignore `mul` operations that don't fit the mental model. Other suggestions would IMO benefit new users making sense of API and to avoid frustrating bugs.

Does this make sense? Has the ship already sailed for substantially changing vector math?

bjornbytes:
The ship hasn't sailed!  There's also #169 which I think about pretty regularly.

I can explain the reasoning behind the current API: it matches GLSL.  That was the main inspiration behind the feature set, overloads, and naming of the current API.

- In GLSL, `*` with two vectors is component-wise, and `*` with a vector and a number is scaling.
- For matrices, `*` with two matrices is composition, but `*` with a vector is transformation.
- quaternions in LOVR just followed matrices

[...]

Issue #169, opened 2019-11-21
Do cross, normalize, etc. need to mutate the original vectors?


bvisness:
I've been bitten several times over the past few days by vector mutations leaking into places I didn't expect. It seems that some operations yield new temporary vectors, while others just mutate the first argument and return it for chaining. It is not clear which operations do this, so I've found myself wrapping lots of things in `vec3` out of paranoia.

While I understand that a mutation-based API might be good for performance reasons, I don't think it's a great default behavior, since it leads to sneaky issues that are hard to debug. I also don't think it's really necessary for most things, since you have this temporary vector system that reduces allocation and GC load.

At the moment, my preferred API design would be:

- All vector operations that return a vector yield a new temporary vector, both for permanent and temporary vectors. (This includes operator overload stuff.)
- The existing set of mutate-y behaviors is renamed to `:mutCross`, `:mutNormalize`, etc. Or something along those lines. This makes it easy to opt into mutation for cases where it matters, like with permanent vectors that you are frequently changing.

However, I understand that this would be a big change that is liable to break, well everything. I just want to bring it up for discussion and get your thoughts on it.

nevyn:
This bit me too; I prefer non-mutating defaults with explicit mutating variants! I like the ruby convention of having “!” variants be mutating.

For compatibility... Maybe a separate package, or some module level flag to opt into new behavior?

bjornbytes:
Yeah, I think this is a tricky area.  The two different systems (mutable and immutable) each have tradeoffs, and I don't know if one is "better".  I think at this point I'm used to the mutations but have probably just internalized the paranoia you describe and developed a bit of stockholm syndrome.  If I take a step back, I can see how avoiding mutations altogether would make the API easier to use in some ways.  On the other hand, it's a bit nice knowing that vectors are only created when `vec3` is called (metamethods are an exception to this though...).

There's also a limited number of temporary vectors (16k at the moment).  It's possible that encouraging more vector allocations would cause projects to bump up against this limit more often, and I'd want to make sure that the workaround for this isn't more annoying than just having mutable vectors be the default.  Mutable variants of functions is one possibility, another one might be optional out-args?  (It's worth saying that the mutable vectors can run out of temp space too).

I'd also like to do some simple profiling to see what the performance characteristics between mutable and immutable are.  As you said, the temp vector system should mean that they're pretty much the same, but sometimes there can be surprises in real projects.

[...]

PR #481, opened 2021-09-08
Move vector rotate/transform to Vec3/4; Breaks API

jmiskovic:
As discussed in #344, the `Mat4:mul(Vec3)` and `Quat:mul(Vec3)` don't
follow the convention of vector library. Those forms mutate the passed
in vector and not the Mat4/Quat object on which method was called.

The `Mat4:mul(Vec3/4)` is moved to `Vec3/4:transform(Mat4)` and the
`Quat:mul(Vec3)` is moved to `Vec3:rotate(Quat)`.

The `Mat4:mul(x,y,z)` seemed like useful form, and it returns number
components without mutation. That form is now expanded to more useful
`Mat4:mul(x,y,z,w)` to be able to treat input vector as directional.
If not supplied, w defaults to 1 so that vector is treated as point. The
function now returns also the w component.

Because Mat4 has `mul(x,y,z,w)`, the similar form is added to Quat for
consistency: `Quat:mul(x,y,z)`.

This change breaks backward compatibility by removing three forms
`Mat4:mul(Vec3)`, `Mat4:mul(Vec4)` and `Quat:mul(Vec3)`.

Because this is a breaking change, maybe we should update docs, examples and the showcase code to be ready before this is actually merged in.


bjornbytes:
Still thinking about #169 -- if add/sub/mul/div returned new vectors (potentially with out-args to keep mutation possible), then everything would be consistent: `Mat4:mul(x, y, z, [w])`/`Mat4:mul(v)` and `Vec3:add()` could all return temp vectors.  That would be a huuuge breaking change though.  It's not something immediately relevant to this PR, and I think this PR does a nice job of making things consistent under the current vector system.  Just something I am keeping in mind.

jmiskovic:
This proposal would take care of `Mat4:mul(Vec3)` mutating the vec3 instead of mat4, but the other issue is that `mul` performs semantically different operations. I would still rather apply rotation to the vector, instead of passing vector to quat.

bjornbytes:
I changed the merge base to `dev` since this is a breaking change.

jmiskovic:
If you are still considering this PR, I'd like modify it so that no API is broken. I would just add the new methods and keep the `Mat4:mul(Vec3)`, `Mat4:mul(Vec4)` and `Quat:mul(Vec3)` around. It makes sense that M x V multiplication can be found under `mat4:mul(vec3)`.

We could also add a warning note in docs about possible confusion.

bjornbytes:
Ok here is a slightly modified plan:

- Add `Vec3/Vec4:transform` as in this PR, except using `luax_readmat4` so numbers can be used as well.
- Add `Vec3:rotate` as in this PR, except using `luax_readquat` so angle/axis numbers can be used as well.

This is an easy non-breaking change.  I'll rebase and commit this.

---

For `Mat4/Quat:mul(vec)`, right now they are the only operation `x:op(y)` that modifies `y` and it's surprising/confusing.  I think they should either return numbers or temp vectors.  (Also I think `Quat:mul(x, y, z)` is missing and should be added).

How to decide between vector/number return type?

- number: saves you a temp vector and function call if you were going to immediately `:unpack`
- vector: saves you a function call and `:set` overhead if you were going to use as vector

`Mat4:mul(x, y, z)` returns numbers currently, but there is also `Quat:direction` that returns a temp vec3.  There are also `Mat4:getTranslation/Rotation/Scale` which return numbers, but I think numbers are required here so e.g. `Mat4:getPose` can be added in the future.

I kind of lean towards temp vector return type, since it seems more likely that you're going to use the result for more vector operations, and the cost of unpacking is a little smaller than the cost of :set.

bjornbytes:
Pushed the rest.  Longest PR ever!

PR #484, opened 2021-09-12
Font:getWidth and lovr.graphics.print now support a first-line indent

mcclure:
**Motivation**

I want to be able to do some really basic text layout. In particular, I want to highlight exactly one word as red in a text printout. The best way I can think of to do this is to print the text before the word in white, the word in red, and then the text after the word in white. In order to do this, I must `print` each of the three substrings, then after each `print` use `getWidth` to "advance" my text pointer. The new "last line" feature in `getWidth` makes this possible.

**Problem**

The problem is once I have the "last line" width, I cannot do anything with it. If I add that width to the `print` location, then the entire paragraph will be pushed right, not just the line I'm trying to advance.

**Solution**

The patch allows the "last line" number to be passed back in to `print` or `getWidth` as an "indent". The indent is in in-font sizing (IE it is before the fontScale transformation).

**Test**

Copy this [main.lua.txt](https://github.com/bjornbytes/lovr/files/7150020/main.lua.txt) into a folder as main.lua and run it with the patch.

Screen Shot 2021-09-12 at 1 19 34 PM

**Other approaches?**

lovr.graphics.print is already calculating the values `getWidth` computes (it calls Measure()), so it could return those values when it's called and save a call and a little bit of time.

### DIFF
--- a/src/api/l_graphics.c
+++ b/src/api/l_graphics.c
@@ -1047,7 +1047,8 @@ static int l_lovrGraphicsPrint(lua_State* L) {
   float wrap = luax_optfloat(L, index++, 0.f);
   HorizontalAlign halign = luax_checkenum(L, index++, HorizontalAlign, "center");
   VerticalAlign valign = luax_checkenum(L, index++, VerticalAlign, "middle");
-  lovrGraphicsPrint(str, length, transform, wrap, halign, valign);
+  float indent = luax_optfloat(L, index++, 0.f);
+  lovrGraphicsPrint(str, length, transform, wrap, halign, valign, indent);
   return 0;
 }
 
--- a/src/api/l_graphics_font.c
+++ b/src/api/l_graphics_font.c
@@ -15,7 +15,8 @@ static int l_lovrFontGetWidth(lua_State* L) {
   float height;
   uint32_t lineCount;
   uint32_t glyphCount;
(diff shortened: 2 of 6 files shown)


bjornbytes:
This solves the problem but I'd prefer to add something to Font that lets you measure out the individual lines. 
 Font just has a lot of arguments already, and it's nice to separate out the layout functionality so you can do other stuff like switch fonts, insert icons between ranges of text, or render a range of lines.

The algo to do this would be something like this

```lua
local x, y = 0, 0
for i, entry in ipairs({ { font, text }, { font2, text2 } }) do
  local font, text = unpack(entry)
  local width, lines, final = font:getWidth(text, wrap)
  lovr.graphics.setFont(font)

  -- if this line is indented, just draw rest of line and reset x to zero
  if x ~= 0 then
    local tail = text:sub(font:getLines(text, wrap - x)[1]) -- get text of remainder of line, hypothetically
    lovr.graphics.print(tail, x, y, ...)
    x = 0
    y = y + font:getHeight()
  end

  -- draw remaining lines in text chunk, if any
  lovr.graphics.print(text, x, y, ..., wrap)
  x = final
  y = y + lines * font:getHeight()
end
```

I'd also like to add a LÖVE-style variant of print that lets you print text with different colors by passing a table of colors/strings instead of a single string.

Issue #491, opened 2021-10-06
Vector: Implement __eq


ottworks:
![image](https://user-images.githubusercontent.com/5253130/136124869-36f05209-0bf7-4779-bf88-5595d5f0613e.png)

bjornbytes:
Something to consider: should this test for exact equality or approximate equality?  In e.g. Unity, `v1 == v2` returns true if the vectors are within `1e-5` of each other.

ottworks:
I think something like `Vec3:approximateEquals(Vec3, threshold=0.00005)`
should be a method in its own right.

jmiskovic:
You can currently compare equality like this: ` vecA:distance(vecB) < 1e-5`. Comparing floats without epsilon margin of error is useless, for example the `0.9` won't be equal to `1.2 - 0.3` in either C or Lua.

If `__eq` is implemented to test approximate equality we would lose the ability to test for identity (if you modify one vector, the other also changes), but I don't think it matters much. One could also meaningfully implement `__lt` and `__le` by comparing the lengths. The change should be made to all vectors and `Quat` and `Mat4`.

ottworks:
I don't want 2*N squares and two square roots involved in the equality
check. Vector equality should simply stem from the equality of its parts.

Voxar:
Our use case is just "did this mat/vec change at all since last frame" for optimisation so I would be happy with an `equals` that just does a pointer check and memcmp.

For vec I think a (b-a):lengthSquared() < e is sufficient for other cases. I would not mind if there's an optional `epsilon` argument for convenience tho, perhaps even defaulted to larger than 0. Depends on the direction of LÖVR I think.

PR #493, opened 2021-10-08
Add :equals method to vector objects;

bjornbytes:
Still discussion going on in #491 so consider this TBD.  An `__eq` metamethod won't work currently because temporary vectors are lightuserdata, which doesn't support `__eq`.


bjornbytes:
Maybe it needs an optional threshold argument.

PR #509, opened 2021-11-23
Fix stencil color mask;

bjornbytes:
- When calling lovr.graphics.stencil, the color mask is initially
  disabled, and gets restored to its initial state afterwards.
- However, when it's restored, it uses lovrGraphicsSetColorMask, which
  just sets shadow state that doesn't make it all the way to GL until
  another draw is done.
- The consequence of this is that if you call .stencil and then don't do
  a draw, any clears that happen will use the old (disabled) color mask,
  preventing the color buffer from being cleared.
- The solution here is to lower the color mask change down into opengl.c
  where it can directly hit OpenGL.

Need to look over this to make sure there aren't any unintended consequences.

### DIFF
--- a/src/modules/graphics/opengl.c
+++ b/src/modules/graphics/opengl.c
@@ -1500,6 +1500,11 @@ void lovrGpuPresent() {
 
 void lovrGpuStencil(StencilAction action, int replaceValue, StencilCallback callback, void* userdata) {
   lovrGraphicsFlush();
+
+  uint8_t lastColorMask = state.colorMask;
+  state.colorMask = 0;
+  glColorMask(GL_FALSE, GL_FALSE, GL_FALSE, GL_FALSE);
+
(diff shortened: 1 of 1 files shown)


jmiskovic:
Doesn't work well in normal use. I'm testing with this code:

```lua
function lovr.draw()
  lovr.graphics.stencil(
    function()
      for x = -1, 1, 0.2 do
      for y =  0, 2, 0.2 do
      for z = -2, 0, 0.2 do
        lovr.graphics.box('fill', x, y, z, 0.05)
      end
      end
      end
    end)
  lovr.graphics.setStencilTest('less', 1)
  lovr.graphics.sphere(0, 0, -10, 5)
end
```

This is the correct render by the master branch:

image

This is render with this fix applied:
image

bjornbytes:
Both pieces of state needed to be updated.

PR #515, opened 2021-12-26
Clarifying comments in GPU-graphics files

mcclure:
This is an ongoing project, I'm not expecting this to be merged immediately or necessarily even at all.


src/modules/graphics/graphics.h
```diff
@@ -185,6 +185,12 @@ typedef enum {
   WINDING_CLOCKWISE
 } Winding;
 
+// i | u | f for signed integer, unsigned integer, floating point
+// <bitcount>
+// n for normalized (signed integer goes to -1 to 1 in shader, unsigned integer goes to 0 - 1 in shader)
+// x<componentcount>
+// (or) matN for NxN matrix
+
```
bjornbytes, line 193: A thing bugging me slightly is that the normalized versions (`i8nx4` and `u8nx4`) don't really make sense.  It's not "signed integer normalized", it's a distinct thing -- there are 8 bits encoded as "signed normalized instead of signed" and "unsigned normalized instead of unsigned".  Maybe `sn8x4` and `un8x4` make more sense, but those start to get really weird.  Other options

- LÖVE: `unorm8vec4` and `snorm8vec4`
- WebGPU: `unorm8x4` and `snorm8x4` (so just un/sn expanded)
- Metal: `uchar4Normalized` and `char4Normalized` (has the same signed/unsigned connotation)
- Vulkan: `R8G8B8A8_UNORM` and `R8G8B8A8_SNORM`, as distinct from `SINT` and `UINT`

PR #516, opened 2021-12-30
pbr-fix to display glass-like materials

coderofsalvation:
the alpha-value of 'roughness' get assigned to lovrDiffuse (which is always 1) unless you set the alpha on the pbr-shader:

image

(above is a screenshot of blender before exporting to glTF)

### DIFF
--- a/src/resources/shaders.c
+++ b/src/resources/shaders.c
@@ -267,7 +267,7 @@ const char* lovrStandardFragmentShader = ""
 "  result = tonemap_ACES(result * lovrExposure); \n"
 "#endif \n"
 
-"  return lovrGraphicsColor * vec4(result, 1.); \n"
+"  return lovrGraphicsColor * vec4(result, lovrDiffuseColor.a ); \n"
 "}"
 
 // Helpers


src/resources/shaders.c
```diff
@@ -267,7 +267,7 @@ const char* lovrStandardFragmentShader = ""
 "  result = tonemap_ACES(result * lovrExposure); \n"
 "#endif \n"
 
-"  return lovrGraphicsColor * vec4(result, 1.); \n"
+"  return lovrGraphicsColor * vec4(result, lovrDiffuseColor.a ); \n"
```
bjornbytes, line 270: Should this also take the base color texture's alpha value into account?

If so, rather than doing `texture(lovrDiffuseTexture, lovrTexCoord)` again (which I think would read from the texture a second time?), the `baseColor` variable above could be turned into a `vec4`, the places using it could append `.rgb`, and then the alpha value here could be `baseColor.a`.  Lemme know if that makes sense!

coderofsalvation: > update: `lovrGraphicsColor * vec4(result, 1.0) * lovrDiffuseColor;` seems to look good/better.

btw. I'm not sure whether I understand what you mean, so I'll summarize what I do understand:
```
vec3 baseColor = texture(lovrDiffuseTexture, lovrTexCoord).rgb * lovrDiffuseColor.rgb;
```
here the texture\s base color texture is read, and indeed without alpha (vec3).
So I changed it to vec4 (and changed future references of `baseColor` to `baseColor.rgb`) and tried:
```
return lovrGraphicsColor * vec4(result, baseColor.a) * lovrDiffuseColor;
```
This worked, however it did result in the environmentMap not being visible anymore on the 'glass' material (alpha:0.3 and low roughness, high metallic)

WDYT?

bjornbytes: I think it would be

```
return lovrGraphicsColor * vec4(result, baseColor.a);
```

Multiplying by `lovrDiffuseColor` would be applying the color a second time

bjornbytes:
I applied this fix in `99a4c3`, let me know if anything still looks off.

PR #525, opened 2022-01-18
Sound:setPlaybackRate

nevyn:


bjornbytes:
Some ideas

- Sources should set their playback rate to 1.0 when created.
- There should be a getter so you can see what the playback rate is.
- Should clones inherit the playback rate from their parent?
- Is there a converter-related race condition between `:setPlaybackRate` and `:clone`?  I think as long as the clone reads the playback rate from its parent once and then configures its converter using that value, things would be okay.

bjornbytes:
Note: enabling dynamic sample rate on a miniaudio data converter disables passthrough optimizations, but this is okay for us since we skip creating the converter altogether when no conversion is necessary.

bjornbytes:
Applied in d8c23bacec733482afbf56fd57749132de14aa8d as `Source:setPitch`.  Thanks for the PR for this.

PR #526, opened 2022-01-29
Thumbrest support;

bjornbytes:
This adds a new button called `thumbrest` (supported on touch, quest 2, and vive focus 3 controllers).  The expectation is that this will usually only return data for `isTouched`.  This needs to be tested.


src/modules/headset/headset_vrapi.c
```diff
@@ -284,6 +284,7 @@ static bool vrapi_isTouched(Device device, DeviceButton button, bool* touched) {
   switch (button) {
     case BUTTON_TRIGGER: *touched = input->Touches & ovrTouch_IndexTrigger; return true;
     case BUTTON_THUMBSTICK: *touched = input->Touches & ovrTouch_Joystick; return true;
+    case BUTTON_THUMBREST: *touched = input->Touches & ovrTouch_ThumbRest; return true;
```
bjornbytes, line 287: Should this be `ThumbRest` or `LThumbRest and `RThumbRest`?  VrApi docs don't say what the difference is.

bjornbytes:
It's unclear if OpenVR supports thumbrests on the touch controllers.

PR #530, opened 2022-02-14
Fix/device buttons

jmiskovic:
Ability to distinguish which controller has A,B,  and which one has X,Y buttons.

Edit: I made two commits because one is VrApi and other is l_module and they are separate issues; can squash if needed.

### DIFF
--- a/src/api/l_headset.c
+++ b/src/api/l_headset.c
@@ -479,8 +479,7 @@ static int l_lovrHeadsetIsDown(lua_State* L) {
       return 1;
     }
   }
-  lua_pushboolean(L, false);
-  return 1;
+  return 0;
 }
 
 static int l_lovrHeadsetWasPressed(lua_State* L) {
@@ -521,8 +520,7 @@ static int l_lovrHeadsetIsTouched(lua_State* L) {
       return 1;
     }
   }
-  lua_pushboolean(L, 1);
-  return false;
+  return 0;
 }
 
 static const int axisCounts[MAX_AXES] = {
--- a/src/modules/headset/headset_vrapi.c
+++ b/src/modules/headset/headset_vrapi.c
@@ -247,6 +247,14 @@ static bool vrapi_isDown(Device device, DeviceButton button, bool* down, bool* c
     return false;
   }
 
+  if (device == DEVICE_HAND_LEFT && (button == BUTTON_A || button == BUTTON_B)) {
+    return false;
+  }
+
+  if (device == DEVICE_HAND_RIGHT && (button == BUTTON_X || button == BUTTON_Y)) {
+    return false;
+  }
+
   if (state.hands[device - DEVICE_HAND_LEFT].Type != ovrControllerType_TrackedRemote) {
     return false;
   }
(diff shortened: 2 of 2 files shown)


bjornbytes:
Should wasPressed/wasReleased get the same treatment?

Also this PR made me notice that OpenXR doesn't have A/B/X/Y actions?!  I'll add that soon.

There's also the question of, if buttons get to return `nil` to signal "no data", should axes and poses do it too?  But I think that causes too much pain, and there is already `isTracked` to check for valid poses.  Still, something like `lovr.headset.isValid` that lets you see if anything (pose, button, axis, skeleton, haptics) is active might be nice.  This discussion comes up a lot though.

jmiskovic:
For wasPressed/Released I saw the docs warn that _Some headset backends are not able to return pressed/released information_, so did not want to mess with headsets I could not test. In this case the returned nil could mean both the lack of button and the lack of support for reading button state change.

Axes currently return zero results when not available, so it is possible to see the control is missing from device. Could be set to nil to follow the convention.

jmiskovic:
As for OpenXR, the poses of controllers are rotated by 90deg along 1,0,0 compared to VrApi. I made a controller visualizer that could be used to check the differences between the two, and maybe between different headset controllers. It is adapted to Quest Touch but tries to scan through all buttons / axes.

```Lua
local hand_names = {'hand/left', 'hand/right'}
local buttons = { 'trigger', 'thumbstick', 'b', 'y', 'a', 'x', 'touchpad', 'menu', 'grip' }
local axes_in_button = { trigger = 1, thumbstick = 2, touchpad = 2, grip = 1 }

local palette = { 0xfbbbad,  -- https://lospec.com/palette-list/twilight-5
                  0xee8695,  -- 2
                  0x4a7a96,  -- 3
                  0x333f58,  -- 4
                  0x292831 } -- 5

lovr.graphics.setBackgroundColor(palette[5])

function lovr.draw()
  local pose = mat4()
  for i, hand in ipairs(hand_names) do
    local side = hand == 'hand/left' and 1 or -1
    -- hand pose block
    pose:set(lovr.headset.getPose(hand))
    lovr.graphics.setColor(palette[4])
    local block_pose = mat4(pose):scale(0.08)
    block_pose:rotate(math.pi/6,  1,0,0)
    block_pose:translate(0, 0, 0.5)
    block_pose:rotate(-math.pi / 2,  0,0,1)
    lovr.graphics.cylinder(block_pose, 0.02, 0.01, true, 3)
    local bones = lovr.headset.getSkeleton(hand)
    if bones then -- hand bones
      lovr.graphics.setColor(palette[3])
      for j, bone in ipairs(bones) do
        pose:set(unpack(bone)):scale(0.005, 0.003, 0.010)
        lovr.graphics.sphere(pose)
      end
    else -- controller buttons bones
      local button_pose = mat4()
      local count = 0
      for j, button in ipairs(buttons) do
        button_pose:set(pose)
        button_pose:rotate(-math.pi / 3,  1,0,0)
        button_pose:translate(side * 0.05, 0, 0)
        button_pose:rotate(math.pi / 8,  0,side,0)
        button_pose:translate(0, -0.01 * count, 0)
        local touched = lovr.headset.isTouched(hand, button)
        local down = lovr.headset.isDown(hand, button)
        local axes = axes_in_button[button]
        if down ~= nil then
          count = count + 1
          if axes and axes == 1 then -- displace text by one or two axes
            local a = lovr.headset.getAxis(hand, button)
            button_pose:translate(0, 0, a * -0.02)
            button_pose:rotate(a * math.pi / 6,  0,side,0)
          elseif axes and axes == 2 then
            local x, y = lovr.headset.getAxis(hand, button)
            button_pose:translate(0, 0, -0.01)
            button_pose:rotate( x * math.pi / 4,  0,1,0)
            button_pose:rotate(-y * math.pi / 4,  1,0,0)
            button_pose:translate(0, 0, 0.01)
          end
          if down then
            button_pose:translate(0, 0, -0.005):scale(1.2)
            lovr.graphics.setColor(palette[1])
          elseif touched then
            button_pose:scale(1.2)
            lovr.graphics.setColor(palette[3])
          else -- idle
            lovr.graphics.setColor(palette[3])
          end
          lovr.graphics.print(string.upper(button), mat4(button_pose):scale(0.01))
        end
      end
    end
  end
end

```

PR #532, opened 2022-02-16
Fix NaN handling when calculating angle for mat4

jmiskovic:
The arccos was incorrectly calculated when cosangle was larger than 1.0


bjornbytes:
This makes sense.

PR #535, opened 2022-02-18
Add Vec2/3/4:angle(other)

jmiskovic:
We could also consider moving vec2 and vec4 implementations to `maf.h`. With the current layout I had to repeat some of code (lenght and dot calculation) because existing Lua-facing functions could not be reused.

Some testing code:

```lua
local function lua_angle(v1, v2)
  return math.acos(v1:dot(v2) / (v1:length() * v2:length()))
end

local v1 = vec2(1, 2)
local v2 = vec2(3, 4)
print(v1:angle(v2))
print(lua_angle(v1, v2))


local v1 = vec3(1, 2, 3)
local v2 = vec3(4, 5, 6)
print(v1:angle(v2))
print(lua_angle(v1, v2))

local v1 = vec4(1, 2, 3, 4)
local v2 = vec4(5, 6, 7, 8)
print(v1:angle(v2))
print(lua_angle(v1, v2))


lovr.event.push('quit')
```

### DIFF
--- a/src/api/l_math_vectors.c
+++ b/src/api/l_math_vectors.c
@@ -253,7 +253,7 @@ static int l_lovrVec2Distance(lua_State* L) {
     uvec[1] = luax_checkfloat(L, 3);
     u = uvec;
   } else {
-    u = luax_checkvector(L, 2, V_VEC2, NULL);
+    u = luax_checkvector(L, 2, V_VEC2, "vec2 or number");
   }
   float dx = v[0] - u[0];
   float dy = v[1] - u[1];
@@ -270,7 +270,7 @@ static int l_lovrVec2Dot(lua_State* L) {
     uvec[1] = luax_checkfloat(L, 3);
     u = uvec;
   } else {
-    u = luax_checkvector(L, 2, V_VEC2, NULL);
+    u = luax_checkvector(L, 2, V_VEC2, "vec2 or number");
   }
   lua_pushnumber(L, v[0] * u[0] + v[1] * u[1]);
   return 1;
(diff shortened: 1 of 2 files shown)


src/api/l_math_vectors.c
```diff
@@ -296,6 +296,25 @@ static int l_lovrVec2Lerp(lua_State* L) {
...
+  float* v = luax_checkvector(L, 1, V_VEC2, NULL);
+  float* u;
+  float uvec[2];
+  float dot, length_v, length_u;
+  if (lua_type(L, 2) == LUA_TNUMBER) {
+    uvec[0] = lua_tonumber(L, 2);
+    uvec[1] = luax_checkfloat(L, 3);
+    u = uvec;
+  } else {
+    u = luax_checkvector(L, 2, V_VEC2, NULL);
```
bjornbytes, line 309: Very minor: changing the NULL here (and in the Vec4 method) to "vec2 or number" would change the error message to say "vec2 or number expected" instead of "vec2 expected".  I'm not even consistent about doing it everywhere though.

src/core/maf.h
```diff
@@ -122,6 +122,11 @@ MAF vec3 vec3_max(vec3 v, const vec3 u) {
   return v;
 }
 
+MAF float vec3_angle(vec3 v, const vec3 u) {
```
bjornbytes, line 125: `v` could be `const` here.

bjornbytes:
These currently return `nan` if the denominator is zero, maybe that's okay or maybe there's a better result.  I've seen other libraries handle this and return both 0 and `math.pi / 2`

src/api/l_math_vectors.c
jmiskovic: I've now corrected all such cases in this source file.

jmiskovic:
Yeah, if either vector is zero-length then the angle doesn't exist at all and `nan` is mathematically correct. I would prefer `nan` to 0 or pi / 2, to give me a hint that I should handle specific edge case better.

jmiskovic:
The conclusion from Slack discussion was to return pi / 2 for zero-length input vectors. PR is now updated to reflect this.

bjornbytes:
Looks good!

Issue #536, opened 2022-02-26
Is it possible to set the filter mode on Fonts?


muratg:
Looks like `lovr.graphics.setDefaultFilter` does not apply to font scaling.  I'd like to use nearest neighbor.  Was able to get it working for textures, but not TTFs. Any ideas?

bjornbytes:
It's not possible right now.  Right now fonts are msdf -- do you want nearest neighbor filtering with those, or do you want the font rasterized with a pixelated appearance without sdf?  Non-sdf fonts are likely going to get added soon.

I'm wondering if a function like `Font:getTexture` would be the way to go.  It might allow for other things beside customizing the filter.  Having it affected by the default filter could work too, but maybe it would be confusing?

muratg:
No, I don't need sdf stuff at all, just the pixelated rasterization.  I have some fonts suitable for pixel games (small, all white pixels, easy to sale in a pixel-perfect way.)  (These work fine in love2d.)

I actually looked for a way to see if I can retrieve the texture out of a Font.  That would work for me as long as I can set the filtering mode on that (i.e. that's not read-only or something.)  I'd be fine with having the default filter affecting it too.  Perhaps it affects it by default, but you can override it, in case you want a different filtering mode there?

I think `Font:getTexture` is the winner as it can enable other scenarios too as you say.

Is this a big change or could there be a simple(r) workaround?

(BTW thank you for the amazing library! I really enjoy playing with it!)

bjornbytes:
Adding `Font:getTexture` would be a pretty easy change.  In fact, there is already a C function for it, so just the Lua bindings would need to get written.

There's one problem with exposing the Texture though: when the Font's texture atlas gets filled up, a new (larger) texture gets created.  This means that textures returned by `Font:getTexture` would become out of date as the atlas is resized.  This maybe suggests that a higher-level approach like respecting the default filter and/or something like `Font:setFilter` would be better, so that Font can internally ensure the filter is still applied as it creates new Textures for its atlas.  I think doing both makes sense, it's also congruent with LÖVE which is nice.  I can probably add this soon, unless you have any interest in sending a pull request for it?

Adding support for non-sdf font rasterizers (and potentially bmfont) is a more involved change, but one that is currently in progress as part of the vulkan work.  It sounds like the use case you described will need to wait for that.

muratg:
> I can probably add this soon, unless you have any interest in sending a pull request for it?
 
[...]

PR #537, opened 2022-02-27
Font filtering now respects the default setting.

muratg:
Per the discussion on https://github.com/bjornbytes/lovr/issues/536.  The default filter setting is now flowed through and used as the font filter.

However, there's a caveat: previously (before this PR) font filter was set to bilinear, as default and always. With this change, it's now configurable, _but_ the default is now trilinear (which is the default for the textures.)

Probably not an issue, but just wanted to bring this up in case it is.
 
I tested this on Linux and Windows and it fixes my issue.


bjornbytes:
Thanks!  I think trilinear is okay because font textures don't have mipmaps.

PR #538, opened 2022-03-04
Use Texture's format when creating an Image from Canvas

lun-4:
Instead of creating images hardcoded as RGBA, always use the selected texture's format.


bjornbytes:
Thanks!

Issue #485, opened 2021-09-12
Word Wrapping Improvements


mcclure:
Awhile ago we had [this bug](https://github.com/bjornbytes/lovr/issues/230), but then it got fixed.

But it is back. See the test program (or just look at the screenshots) in [this PR](https://github.com/bjornbytes/lovr/pull/484). The test program draws a box around the wrap area of a string (the bottom line is fixed at six lines). As you can see text regularly exceeds the right line by exactly one word.

The test program assumes the patch in #484 but you can quickly make the test program work in vanilla lovr by commenting out the second and third `print()` calls.

bjornbytes:
These are different issues.  In #230, new words were starting past the wrap limit if there was a space straddling the wrap limit, which was not intended behavior.  In the #484 screenshot, after a word goes past the wrap limit, the next word is wrapped onto the next line, which is the current intended behavior.  (In addition, it looks like "downstairs" and "live" end right up against the wrap limit which would have triggered #230, so it looks like that's still fixed).

bjornbytes:
Word wrapping probably works this way because it was simpler to implement -- you can start pasting vertices into a buffer immediately, and when you hit a space decide if you want to wrap.

While porting Font to Vulkan I'll see how easy it is to improve on this -- would need to first measure a word, then decide where to put it based on how long it is, then paste the vertices.

bjornbytes:
This is improved on the dev branch (text is guaranteed to not go past the wrap limit).

PR #539, opened 2022-03-11
Add Font:getMeasure

Voxar:
Takes same arguments as getWidth(). 
Returns all the results from `lovrFontMeasure`

### DIFF
--- a/src/api/l_graphics_font.c
+++ b/src/api/l_graphics_font.c
@@ -22,6 +22,26 @@ static int l_lovrFontGetWidth(lua_State* L) {
   return 3;
 }
 
+static int l_lovrFontGetMeasure(lua_State* L) {
+  Font* font = luax_checktype(L, 1, Font);
+  size_t length;
+  const char* string = luaL_checklstring(L, 2, &length);
+  float wrap = luax_optfloat(L, 3, 0.f);
+  float width;
+  float lastLineWidth;
+  float height;
+  uint32_t lineCount;
+  uint32_t glyphCount;
+  lovrFontMeasure(font, string, length, wrap, &width, &lastLineWidth, &height, &lineCount, &glyphCount);
+  height /= lovrFontGetPixelDensity(font);
+  lua_pushnumber(L, width);
+  lua_pushnumber(L, height);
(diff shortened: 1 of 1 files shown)


bjornbytes:
Thanks for sending this.  Can you describe what you need it for (i.e. do you need the glyph count?  is this just for convenience?  performance?)?  The font measurers are a bit messy atm -- we just added `lastLineWidth` to satisfy a feature request, may add `:getWrap` to return each wrapped line in a table.  It helps to know what everyone's needs are to come up with the best design.

Voxar:
We need the total height of the wrapped text as it will later be rendered. We calculate and cache the bounds of text elements for culling.
Don't need the glyphCount but I though why not include it as it's there. Perhaps a silly thing to return. 

The returned height should probably be include the `Font:getLineHeight()` tho. 

I didn't discover until later that the height returned from `lovrRasterizerMeasure` is the same as `Font:getHeight() * lineCount`. To deduce the later you need to know a bit about the internals tho so I think it's a bit cleaner api perhaps?

Voxar:
Testing this a bit closer. It behaves really weird so I'm closing this for now

Voxar:
Demo 

https://user-images.githubusercontent.com/45087/158560685-59e4ba90-1c40-4cac-a060-cac0dd1a4bbb.mov

White square is width and height as returned from getMeasure. 
Red square is the width sent as the `wrap` argument and the height calculated with `font:getHeight() * font:getLineHeight() * numberOfLines`.
`getMeasure` returns a better height estimate when using lineSpacing != 1, otherwise the same result. 

(the weirdness experienced yesterday was because I removed the `lua_pushnumber(L, glyphCount);` without decrementing the number of returned values -.-)

bjornbytes:
#485 starts to discuss some improvements to word wrapping -- right now words can extend past the wrap limit because "width > wrap" is checked only on spaces.  I think this explains the difference in width.

The height difference does look surprising though.

bjornbytes:
Looks like `Font:getLineHeight` uses `lua_pushinteger`, which is a mistake.

bjornbytes:
Fixed!

PR #542, opened 2022-03-15
Fix crash when raycasting meshshape

Voxar:
https://ifyouwannabemylovr.slack.com/archives/C59QZ4V6Y/p1647379743806479


Voxar:
I'll push a proper that also returns all hits

PR #550, opened 2022-04-23
Allow passing a file to lovr;

bjornbytes:
The file will be treated as main.lua.
Its directory will be treated as the source.

### DIFF
--- a/src/modules/filesystem/filesystem.c
+++ b/src/modules/filesystem/filesystem.c
@@ -148,6 +148,22 @@ bool lovrFilesystemInit(const char* archive) {
   if (archive) {
     state.source[LOVR_PATH_MAX - 1] = '\0';
     strncpy(state.source, archive, LOVR_PATH_MAX - 1);
+
+    // If the command line parameter is a file, use its containing folder as the source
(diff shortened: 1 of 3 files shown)


src/modules/filesystem/filesystem.c
```diff
@@ -148,6 +148,22 @@ bool lovrFilesystemInit(const char* archive) {
   if (archive) {
     state.source[LOVR_PATH_MAX - 1] = '\0';
     strncpy(state.source, archive, LOVR_PATH_MAX - 1);
+
+    // If the command line parameter is a file, use its containing folder as the source
+    size_t length = strlen(state.source);
+    if (length > 4 && !memcmp(state.source + length - 4, ".lua", 4)) {
```
bjornbytes, line 154: Hm, this could use `fs_stat` instead of a weird extension test, but that would be at odds with what boot.lua is doing.

jtoy: love this idea, I was going to ask about this feature. I write lots of test scripts and it is getting painful copying around all the other files in the dir around to ech test script.

Issue #551, opened 2022-04-24
Master has a compile on my windows system


jascase901:
```
cmake ../
-- Selecting Windows SDK version 10.0.19041.0 to target Windows 10.0.19043.
-- Looking for pthread.h
-- Looking for pthread.h - not found
-- Found Threads: TRUE  
-- Using Win32 for window creation
-- The ASM compiler identification is MSVC
-- Found assembler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.28.29910/bin/Hostx64/x64/cl.exe
-- Check if the system is big endian
-- Searching 16 bit integer
-- Looking for sys/types.h
-- Looking for sys/types.h - found
-- Looking for stdint.h
-- Looking for stdint.h - found
-- Looking for stddef.h
-- Looking for stddef.h - found
-- Check size of unsigned short
-- Check size of unsigned short - done
-- Using unsigned short
-- Check if the system is big endian - little endian
-- Performing Test NO_STACK_PROTECTOR_FLAG
-- Performing Test NO_STACK_PROTECTOR_FLAG - Failed
-- Check target system architecture: x86_64
-- Check target fpu: FPU IS Hard
-- Check target fpu abi: FPU ABI IS Hard
-- Looking for alloca
-- Looking for alloca - not found
-- Looking for clock_gettime
-- Looking for clock_gettime - not found
-- Looking for gettimeofday
-- Looking for gettimeofday - not found
-- Looking for isnan
-- Looking for isnan - not found
-- Looking for isnanf
-- Looking for isnanf - not found
-- Looking for _isnan
-- Looking for _isnan - found
-- Looking for _isnanf
-- Looking for _isnanf - found
-- Looking for pthread_attr_setstacklazy
-- Looking for pthread_attr_setstacklazy - not found
-- Looking for include file alloca.h
-- Looking for include file alloca.h - not found
-- Looking for include file malloc.h
-- Looking for include file malloc.h - found
-- Looking for include file inttypes.h
-- Looking for include file inttypes.h - found
-- Looking for include file sys/time.h
-- Looking for include file sys/time.h - not found
-- Looking for include file unistd.h
-- Looking for include file unistd.h - not found
-- Found PythonInterp: C:/msys64/mingw64/bin/python3.exe (found suitable version "3.9.10", minimum required is "3") 
-- Found OpenGL: opengl32   
-- Enabling OpenGL support
[...]

PR #552, opened 2022-04-24
use static assert macro instead of keyword

jascase901:
This change fixes a linker error I noticed when compiling on my windows
sytstem. see #551

According to this link the macro is prefered to the keyword.
https://docs.microsoft.com/en-us/cpp/c-language/static-assert-c?view=msvc-170#:~:text=The%20_Static_assert%20keyword%2C%20and%20the%20static_assert%20macro%2C%20both,runtime%20and%20incur%20a%20runtime%20cost.%20Microsoft-specific%20behavior


bjornbytes:
This didn't work for me on Linux until I added `#include <assert.h>` to get the

PR #553, opened 2022-04-24
Compiler intrinsic no retrun on mvsc

jascase901:
Im not sure if this is good way of solving this or not. I personally prefer to avoid compiler intrinsic. And I am planning on upgrading my version of mvsc when I get less lazy. 

edit: this code should add support for Microsoft (R) Build Engine version 16.9.0+5e4b48a27 for .NET Framework

### DIFF
--- a/Tupfile.lua
+++ b/Tupfile.lua
@@ -180,7 +180,7 @@ end
 troublemakers = {
   glad = '-Wno-pedantic',
   os_android = '-Wno-format-pedantic',
-  miniaudio = '-Wno-unused-function',
+  miniaudio = '-Wno-unused-function -Wno-pedantic',
 }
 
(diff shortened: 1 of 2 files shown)


bjornbytes:
Isn't this unnecessary since MSVC supports `_Noreturn` when compiling as C11?

jascase901:
I wouldn't merge this.  Updating my cmake version from `cmake version 3.16.4` to `cmake version 3.23.1`  makes this code change unnecessary.  If cmake  versions less than 3.18   don't support c11 with mvsc, changing code to make lovr build with eariler versions of cmake is playing with fire,.

jascase901:
I should also say I ended upgrading msvc as well.

PR #555, opened 2022-04-27
Add `World:setStepIterations(num)` and its getter

jmiskovic:
Physics world's "quick step" is executed in multiple iteration steps.
The getter and setter for this value is now made available as two new
methods in the World object.

This is allows user to balance between the less accurate but quick
simulations, and more stable behavior of physics.

Something similar was already possible, by reducing the delta time and
running the sim multiple times per frame. However, any force user applies
to collider is zeroed after each step. User would thus have to keep track
of applied forces, and re-apply them inside the physics iteration loop.

By default ODE uses 20 iterations in quick step.

### DIFF
--- a/src/api/l_physics_world.c
+++ b/src/api/l_physics_world.c
@@ -367,6 +367,20 @@ static int l_lovrWorldIsCollisionEnabledBetween(lua_State* L) {
   return 1;
 }
 
+static int l_lovrWorldGetStepCount(lua_State* L) {
+  World* world = luax_checktype(L, 1, World);
+  int iterations = lovrWorldGetStepCount(world);
+  lua_pushnumber(L, iterations);
+  return 1;
+}
+
+static int l_lovrWorldSetStepCount(lua_State* L) {
+  World* world = luax_checktype(L, 1, World);
+  int iterations = luaL_checkinteger(L, 2);
+  lovrWorldSetStepCount(world, iterations);
+  return 0;
+}
+
 const luaL_Reg lovrWorld[] = {
   { "newCollider", l_lovrWorldNewCollider },
   { "newBoxCollider", l_lovrWorldNewBoxCollider },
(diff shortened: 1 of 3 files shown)


src/api/l_physics_world.c
```diff
@@ -367,6 +367,20 @@ static int l_lovrWorldIsCollisionEnabledBetween(lua_State* L) {
...
+static int l_lovrWorldGetStepIterations(lua_State* L) {
+  World* world = luax_checktype(L, 1, World);
+  int iterations = lovrWorldGetStepIterations(world);
+  lua_pushnumber(L, iterations);
+  return 1;
+}
+
+static int l_lovrWorldSetStepIterations(lua_State* L) {
+  World* world = luax_checktype(L, 1, World);
+  float iterations = luaL_checkinteger(L, 2);
```
bjornbytes, line 379: This should be int

src/modules/physics/physics.c
```diff
@@ -191,6 +191,14 @@ void lovrWorldUpdate(World* world, float dt, CollisionResolver resolver, void* u
   dJointGroupEmpty(world->contactGroup);
 }
 
+int lovrWorldGetStepIterations(World* world) {
+  return dWorldGetQuickStepNumIterations (world->id);
```
bjornbytes, line 195: nitpick: space before parens

bjornbytes:
It will be good to expose this.  My thought process while looking it over:

- Should it be part of newWorld?  There are kind of too many parameters already...
- Should it be an input to `update`?  It doesn't seem like the kind of thing you'd want to change every frame, so this would maybe just be annoying.
- Naming?  Since this is dealing with the *number* of iterations/steps, it should probably use a "count" suffix instead of a plural, like `getStepIterationCount` or just `getStepCount`.

jmiskovic:
Good points. Most often you'd experimentally find the best value and init the world with it. Some advanced users might dynamically lower #iterations when FPS drops.

I agree that newWorld() already has too many params of different types. Having it as optional parameter in update() would work, except in this case:

```
world:update(dt)  -- the default value of 20 is used
world:update(dt, 30)
world:update(dt)  -- be stateful and use 30? or revert to 20?
```

My thinking was to make it a World method so that most people can safely ignore the feature and it's there when you reach for it. I'd keep it as separate method.

As for rename I vote for get/setStepCount; less typn.

bjornbytes:
:+1: to separate method and get/setStepCount

bjornbytes:
Thanks!

bjornbytes:
[Added docs](https://github.com/bjornbytes/lovr-docs/commit/f20546911d416d825294d39fce9446dd9522e0f5)

PR #556, opened 2022-04-28
Hardware raytracing acceleration

mcclure:
This is probably a bad idea.

This is a feature branch in which I'm attempting to add RTX support to the Vulkan branch. Current status is it can check if raytracing is supported, enable it if so, and report back to Lua what it did.

This feature is problematic for reasons including RTX being questionably practical at VR framerates, unlikelihood of support on Android and raytracing being overall simply complex. But, I want to experiment with it. Long term, if we choose to add this feature it should probably have a shader-based fallback implementation for systems that lack RTX.

### DIFF
--- a/src/core/gpu_vk.c
+++ b/src/core/gpu_vk.c
@@ -2833,3 +2833,16 @@ bool gpu_raytrace_get_buildsize(gpu_raytrace_acceleration_type rat, uint32_t str
 
   return true;
 }
+
+bool gpu_raytrace_build(gpu_stream *stream, gpu_raytrace_acceleration_type rat, uint32_t structCount, void *geometry, uint32_t rangeCount, void *ranges) {
+  VkAccelerationStructureBuildGeometryInfoKHR input;
+  input.sType = VK_STRUCTURE_TYPE_ACCELERATION_STRUCTURE_BUILD_GEOMETRY_INFO_KHR;
+  input.type = rat==GPU_RAYTRACE_ACCELERATION_TYPE_TOP?
+    VK_ACCELERATION_STRUCTURE_TYPE_TOP_LEVEL_KHR :
+    VK_ACCELERATION_STRUCTURE_TYPE_BOTTOM_LEVEL_KHR;
+  input.mode = VK_BUILD_ACCELERATION_STRUCTURE_MODE_BUILD_KHR;
+  input.geometryCount = structCount;
+  input.pGeometries = (VkAccelerationStructureGeometryKHR *)geometry;
+
+  vkCmdBuildAccelerationStructuresKHR(stream->commands, 1, &input,)
+}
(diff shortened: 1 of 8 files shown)


src/core/gpu_vk.c
```diff
@@ -144,7 +147,8 @@ static struct {
 enum { CPU, GPU };
 enum { LINEAR, SRGB };
 
-static void hash32(uint32_t* hash, void* data, uint32_t size);
+// Had to rename vHash to include util.h-- can be removed before merge
+static void vHash32(uint32_t* hash, void* data, uint32_t size);
```
mcclure, line 151: util.h also has a hash32() as a #define. Although the lovrLogs in this patch are not necessary and the util.h include can be removed, probably it's worth it to keep this function name changed to avoid clashes if util.h is ever added for a good reason.

src/core/gpu_vk.c
```diff
@@ -1932,8 +1937,40 @@ bool gpu_init(gpu_config* config) {
...
     };
 
+    VkPhysicalDeviceAccelerationStructureFeaturesKHR enableAccelerationStructure = { // Raytracing (will be linked in later)
+      .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_ACCELERATION_STRUCTURE_FEATURES_KHR,
+      .accelerationStructure = true
+    };
+
     if (config->features) {
-      VkPhysicalDeviceShaderDrawParameterFeatures supportsShaderDrawParameter = { .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_DRAW_PARAMETER_FEATURES };
+      // Check required extensions for raytracing
```
mcclure, line 1946: Multiple problems with this next section: Does not give the user the ability to disable raytracing support; goes behind Lovr's back and gets extensions directly from vkEnumerateDeviceExtensionProperties instead of integrating with the existing check-extensions system (although I think bjorn said he was removing that when he merges with dev).

bjornbytes: - re disabling raytracing: I think it's okay if you can't disable.  As a single data point, mesa's AMD driver totally ignores whether you enable/disable raytracing.  At first I built some complicated system to let Lua request individual features, but it turns out drivers don't really care that much what you enable (with a few exceptions like the 'robustBufferAccess' feature).

shakesoda:
re questionably practical at vr framerates: i have to contest this on a couple points:

1. this is unlikely to be true for too much longer (i.e. a hardware generation or two if i'm being a bit more conservative)
2. vr eye buffer resolutions are only a subset of relevant use cases.
3. vr framerates at the bottom end of the scale are only around 45fps if we're accounting for tech like ASW - and i believe that's not outlandish on current gen hw

i agree that there should ideally be a fallback path, though, i'm going to have useful systems without RT hardware for quite some years to come where it'd still be very useful for experimentation.

bjornbytes:
`dev` branch is probably in a mature-enough state for this feature now

mcclure:
This is now rebased on the dev branch and "works" (can successfully report raytrace capability when run with https://github.com/mcclure/lovr-gpu-sample test "1-info").

bjornbytes:
Poke, wondering if this is still active and should be kept open?  Still excited about it and would be interested in contributing design/code at some point.

PR #557, opened 2022-04-30
Add piston joint to physics

jmiskovic:
The piston is similar to slider joint, but it allows rotation around the joint axis.

I noticed that the slider joint behaves erratically and unstable when colliders are far from each other. I hoped that piston joint would not have that behavior. It turns out that piston joint is slightly better, but still not perfect. The way to properly fix this is avoiding quickStep, or increasing iteration count.

Maybe having both slider joint and piston joint would confuse new users? To me piston is slightly more versatile, with some torque applied it can be made to work like slider joint. Anyway, the code is working so I'm archiving it as a PR.

### DIFF
--- a/src/api/l_physics.c
+++ b/src/api/l_physics.c
@@ -18,6 +18,7 @@ StringEntry lovrJointType[] = {
   [JOINT_DISTANCE] = ENTRY("distance"),
   [JOINT_HINGE] = ENTRY("hinge"),
   [JOINT_SLIDER] = ENTRY("slider"),
+  [JOINT_PISTON] = ENTRY("piston"),
   { 0 }
 };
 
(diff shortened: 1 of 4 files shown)


bjornbytes:
Hm, I do think it will be somewhat confusing to have 2 "single axis translation" joints.  It's interesting how the APIs are different between the two joint types in ODE.

bjornbytes:
Planning on trying to make this an `allowrotation` flag on `SliderJoint` (I still need to fully investigate if slider/piston are compatible in ODE but I *think* it'll work out).  Just writing down the intent here, I should get to it eventually but it keeps dropping in priority.

bjornbytes:
If I'm reading correctly, Jolt only has slider joint and it doesn't support rotation, right?  If so, maybe this can be closed then.

jmiskovic:
Yes, Jolt's slider joint fixes the rotation. Piston could be done with `SixDOFConstraint` but I haven't experimented with that. No one specifically asked for piston joint and I made this just to avoid the behavior of ODE's slider joint.

PR #559, opened 2022-05-06
Add support for setting the display refresh-rate.

sgosselin:
## Details

OpenXR provides APIs to enumerate the list of supported refresh-rates, and
selecting a new refresh-rate. This is useful on Quest 2 because the headset
supports 60/72/80/90Hz.

This patch adds two new APIs to the lovr.headset module:
  - `lovr.headset.getDisplayFrequencies()`:
	This function returns a table containing the supported refresh-rates on
	success; nil otherwise.
  - `lovr.headset.setDisplayFrequency(refreshRate:number)`:
	This function takes the new refresh-rate to select as an argument and returns
        true on success, false otherwise.

Only the OpenXR backend has support for this feature and it is gated by the
"refreshRate" OpenXR feature flag, similarly to what the `getDisplayFrequency()`
API does.

## Testing

```lua
function lovr.load()
  supportedFrequencies = lovr.headset.getDisplayFrequencies()
  if supportedFrequencies ~= nil then
    selectedFrequency = math.max(unpack(supportedFrequencies))
  else
    selectedFrequency = 0.0
  end

  print('selected frequency: ' .. selectedFrequency)
  print(lovr.headset.setDisplayFrequency(0.0))
end

function lovr.update(dt)
  print(dt)
end
```
- On Quest 2, `lovr.headset.getDisplayFrequencies()` returns `{60,72,80,90}`
  and `lovr.update(dt)` is called every ~11ms, as expected since the code
  select 90Hz.
- On Linux, `lovr.headset.getDisplayFrequencies()` returns `nil` and the call
  to `lovr.headset.setDisplayFrequency(0.0)` returns `false`.

### DIFF
--- a/src/modules/headset/headset_openxr.c
+++ b/src/modules/headset/headset_openxr.c
@@ -114,7 +114,9 @@ EGLConfig os_get_egl_config(void);
   X(xrDestroyHandTrackerEXT)\
   X(xrLocateHandJointsEXT)\
   X(xrGetHandMeshFB)\
-  X(xrGetDisplayRefreshRateFB)
+  X(xrGetDisplayRefreshRateFB) \
+  X(xrEnumerateDisplayRefreshRatesFB) \
+  X(xrRequestDisplayRefreshRateFB)
 
 #define XR_DECLARE(fn) static PFN_##fn fn;
 #define XR_LOAD(fn) xrGetInstanceProcAddr(state.instance, #fn, (PFN_xrVoidFunction*) &fn);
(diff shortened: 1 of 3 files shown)


src/modules/headset/headset_openxr.c
```diff
@@ -1019,6 +1021,20 @@ static float openxr_getDisplayFrequency(void) {
...
+  XR(xrEnumerateDisplayRefreshRatesFB(state.session, 0, count, NULL));
+  float *frequencies = malloc(*count * sizeof(float));
+  lovrAssert(frequencies, "Out of Memory");
+  XR(xrEnumerateDisplayRefreshRatesFB(state.session, *count, count, frequencies));
+  return frequencies;
+}
+
+static bool openxr_setDisplayFrequency(float frequency) {
+  if (!state.features.refreshRate) return false;
+  XR(xrRequestDisplayRefreshRateFB(state.session, frequency));
```
bjornbytes, line 1035: I wonder if this should manually handle `XR_ERROR_DISPLAY_REFRESH_RATE_UNSUPPORTED_FB` and return `false` for that instead of throwing an error.

bjornbytes:
This is useful, thanks!  I think eventually it might be nice to expose the 'refresh rate changed' event, but maybe we should wait until the extension becomes EXT/KHR for that.  I do kind of prefer the term "refresh rate" instead of "display frequency" but it's not worth making a breaking API change.

src/modules/headset/headset_openxr.c
sgosselin: Good idea! I updated the patch.

sgosselin:
@bjornbytes yeah I thought about calling these news APIs "xRefreshRate" but I figured I should keep this consistent with the existing `getDisplayFrequency()` API. I updated the patch with your suggestion, thanks for the review!

PR #560, opened 2022-05-06
Allow XR_KHR_android_create_instance to be optional

rpavlik:
It's essentially replaced by KHR_loader_init_android, and will eventually not be provided by new/updated runtimes, so don't require it.

That said, it's safe to pass its struct even if the extension isn't available, since it will just be ignored if unrecognized.


bjornbytes:
Makes sense.  Thanks!

PR #569, opened 2022-06-27
Rationalize, and properly cast , sizes in dev graphics/gpu code (incomplete)

mcclure:
The goal of this PR is to remove the warnings you currently get when compiling the dev branch on x64. Almost all of those warnings occur when a size_t is stored in a u32.

Currently the dev branch stores all graphics-related sizes in uint32_t. This is considered safe because of asserts preventing buffers from being larger than 1GB so theoretically we just have to add casts. However I decided to create new types `graphics_size` and `gpu_size` (there have to be two because gpu.h and graphics.h do not either include the other) for two reasons: first off, if the limit is ever increased above 4GB then without a specific typedef it will be much much harder once the casts are added to go back and identify every place a size is squished into 32 bits; second off, I believe defining semantic types for sizes makes it easier to notice when bugs are present in size assignment code. As evidence for this argument, when I started going through and adding the graphics_size casts everywhere, I found two separate overflow bugs in size assignments, one potentially serious.

So now this PR has three goals:
1. Replace the `uint32_t`s that store size currently with `gpu_size` or `graphics_size` as appropriate
2. Add proper casts whenever `size_t` or `VkDeviceSize` is assigned to `graphics_size` or `gpu_size`
3. Make sure all places sizes are assigned to `graphics_size`. `gpu_size` or `uint32_t` really are safe

The gpu_size/graphics_size code was not exactly clean once I started writing it so maybe there is room to disagree on whether we should do (1). However, I think we do need to do (3), because I did find these two bugs (or at least, failures to assert where the code is currently attempting to assert):

1. `measureTexture` returns a size_t and then immediately gets assigned to uint32_t. It should return a 32-bit type to match texture size elsewhere in lovr, and either measureTexture or its call sites should check for overflow.
2. The `lovrCheck` asserts in `lovrGraphicsGetBuffer`, `lovrBufferCreate` and `lovrTextureCreate` that sizes are under 1<<30 are actually unsafe, because the first two multiply and store in u32 before testing for 1<<30, so they could overflow, and the third relies on measureTexture\*. For example lovrBufferCreate with size 270000000 and stride 16 would overflow, pass the `lovrCheck` as success, and then do something wrong.

I partially implemented this patch but got stuck because these two decisions need to be made:

1. How should measureTexture handle the potential for overflowing the 1GB limit? lovrCheck? If we expect , how do we handle u32/u64 multiply overflow? I think my vote for now would be lovrCheck.
2. Because we're potentially going to be using it in more places, I would like to put the 1 <<30 size limit number in some kind of constant, and Bjorn in Slack recommended including it in GraphicsLimits. Where should it live? Or a different way of asking is, (a) should it be exposed in graphics.h? (b) should there be a copy of it in gpu_vk.c and/or gpu.h? From discussion with Bjorn in slack, I think the correct way to do it is to put the constant only in graphics.c (and report it up through GraphicsLimits\*\*), because gpu_vk.c does not have anywhere it enforces the 1GB size limit itself (that is the job of graphics.c).

\* The fact measureTexture returns a size_t almost saves us on 64-bit systems because multiplying two u32s will never overflow a u64, but measureTexture in some paths multiplies three u32s...
\*\* However if it turns out any `l_*.c` files need to independently check if something exceeds 1GB, I might want to revisit this, because GraphicsLimits is a pretty big struct so declaring it on the stack every time just to call lovrGraphicsGetLimits could get annoying.


src/api/l_graphics.c
```diff
@@ -1062,7 +1062,7 @@ static int l_lovrGraphicsNewShader(lua_State* L) {
   }
 
   info.flags = flags.data;
-  info.flagCount = flags.length;
+  info.flagCount = (graphics_size)(flags.length);
```
bjornbytes, line 1065: Note that flagCount is not a size, it's a count.  It does need to get narrowed to a u32 at some point, but it should not use `graphics_size` since it does not represent the size or offset of a piece of GPU memory.

bjornbytes:
I am super on board with the goal here.  It's really important that there's no unintentional truncation, and passing a 4GB+ Blob to the graphics stuff should error properly.  I've been attentive to this, and adopted the 1GB limit as a constraint to help simplify it a bit (fewer 32 <-> 64 conversions).  But some things fell through the cracks, and it's good to fix them.

My initial impression about adding the size type: To be honest, I think the size types will add cognitive overhead while writing code (I'll need to second guess every time I write uint32_t), and I guess that is probably the point.  There is something lost by moving away from the simplicity of "everything is uint32_t".  I will probably forget to make something a gpu_size instead of a uint32_t a few times.

To answer your 2 questions:

- The measureTexture thing is difficult.  One idea is for measureTexture to consider a u32/u64 overflow "unreachable code", and add nicer errors to lovrTextureCreate and the transfer function ("Texture too big", "Copy range exceeds texture size").
- The 1GB limit is separate-from-but-related-to the integer narrowing.  There are 3 people that have their own limits for this:
  - The GPU driver has a limit of the maximum amount of memory it can allocate.  This is given by the `maxMemoryAllocationSize` limit, but it's in an extension (`VK_KHR_maintenance3`) so I'm too lazy to expose it right now.  It will always be at least 1GB.
  - core/gpu has its own limit of the maximum amount of "size" it supports.  Right now that's 4GB, because it uses u32 for sizes.  So if core/gpu was to expose the Vulkan limit to the graphics module, it would need to cap it to 4GB first.
  - The graphics module has a limit of the max GPU size it supports.  Right now it's arbitrarily set to 1GB to align with the hypothetical Vulkan limit.  So if the graphics module wants to expose the limit to Lua, it needs to cap it to 1GB.
  - Ultimately, my recommendation to "ossify" this constant somewhere would be to use a limit, since it will eventually become one.  The laziest way to do it would be to add it to GraphicsLimits hardcoded to 1GB, but nowhere else.  The least-lazy way to do it would be to add the Vulkan extension and bubble it all the way up to gpu_limits and GraphicsLimits, `MIN`ing it each time.  The moderately lazy way would be to add it to gpu_limits and GraphicsLimits, but not add the Vulkan extension.

mcclure:
>consider a u32/u64 overflow "unreachable code", and add nicer errors to lovrTextureCreate and the transfer functions

The one that confuses me is lovrPassCopyImageToTexture. It calls measureTexture like six or seven times and I don't at a glance understand what it's for. Are these ranges inside *existing* textures? Does that mean they're all "proven safe" already?

bjornbytes:
The copy size is proven safe for the texture using `checkTextureBounds`, which should transitively mean the measureTexture call is safe (but it doesn't hurt for measureTexture to check for overflow anyway)

src/api/l_data_modelData.c
```diff
@@ -536,7 +537,7 @@ static int l_lovrModelDataGetAnimationChannelKeyframe(lua_State* L) {
   for (uint32_t i = 0; i < count; i++) {
```
mcclure, line 537: Hey so for the record what guarantees that channel->property <= 3?

src/modules/graphics/graphics.c
```diff
@@ -26,6 +26,11 @@ const char** os_vk_get_instance_extensions(uint32_t* count);
 #define MAX_SHADER_RESOURCES 32
 #define MATERIALS_PER_BLOCK 1024
 
+// 1GB is a very specific number because the gpu.h limits are always at least this large.
+// If this constant ever goes up, we will need to fetch the gpu limit and compare.
+#define MAX_BUFFER_SIZE (MAX_FRAME_MEMORY)
+#define MAX_BUFFER_SIZE_STR "1 GiB"
```
mcclure, line 32: I regret to report that the IEC officially defines 1 GB as 1000*1000*1000 and 1 "Gibibyte", a word no one ever uses, as 1024*1024*1024. I don't agree with this but error messages should probably be exact

src/modules/graphics/graphics.c
```diff
@@ -1412,7 +1431,7 @@ Shader* lovrShaderCreate(ShaderInfo* info) {
     spv[i].attributes = tempAlloc(spv[i].attributeCount * sizeof(spv_attribute));
     spv[i].resources = tempAlloc(spv[i].resourceCount * sizeof(spv_resource));
 
-    result = spv_parse(info->stages[i]->data, info->stages[i]->size, &spv[i]);
+    result = spv_parse(info->stages[i]->data, (uint32_t)info->stages[i]->size, &spv[i]);
```
mcclure, line 1434: Hey question, you call spv_parse on what appears to be the exact same data twice in this function. Why?

mcclure:
I think this is ready for review now. There is still time to withdraw graphics_size/gpu_size but I think it will lead to more accurate code long run.

I did not fully thread buffer size through the limits because it is "safe" to assume the gpu correction will always be greater than or equal to 1GB anyway (there is a comment documenting this assumption) and I did not do any fixups inside measureTexture because it turned out to not be necessary (?!?). It's possible I missed a measureTexture external safety check however.

mcclure:
Oh, and I did NOT fix the various multiply overflow sites I mentioned. Sorry. Maybe in another patch.

src/modules/graphics/graphics.c
bjornbytes: First call returns the amount of memory to allocate

src/api/l_data_modelData.c
bjornbytes: It's an enum

src/modules/graphics/graphics.c
bjornbytes: Hm, I'd almost rather do `1073741824`

src/api/l_data_modelData.c
```diff
@@ -336,7 +336,8 @@ static int l_lovrModelDataGetMeshVertex(lua_State* L) {
     ModelAttribute* attribute = mesh->attributes[i];
     if (!attribute) continue;
 
-    uint32_t stride = model->buffers[attribute->buffer].stride;
+    lovrCheck(model->buffers[attribute->buffer].stride < UINT32_MAX, "Model buffer stride is impossibly large");
+    uint32_t stride = (uint32_t)model->buffers[attribute->buffer].stride;
```
bjornbytes, line 340: Can we just type this as `size_t` and avoid the check?

src/api/l_data_modelData.c
```diff
@@ -536,7 +537,7 @@ static int l_lovrModelDataGetAnimationChannelKeyframe(lua_State* L) {
   for (uint32_t i = 0; i < count; i++) {
     lua_pushnumber(L, channel->data[keyframe * count + i]);
   }
-  return count + 1;
+  return (int)(count + 1);
```
bjornbytes, line 540: Can we just make `counts` and `count` and `i` an `int` and forgo the cast?

src/api/l_graphics.c
```diff
@@ -721,7 +721,7 @@ static int l_lovrGraphicsGetBuffer(lua_State* L) {
     default: {
       Blob* blob = luax_totype(L, 1, Blob);
       if (blob) {
-        info.length = blob->size / info.stride;
+        info.length = (graphics_size)blob->size / info.stride;
```
bjornbytes, line 724: I don't think `graphics_size` is the correct type to use for a buffer length.  A buffer *size*, the size of its memory in bytes, should be a graphics_size, but the length is not a memory size and `uint32_t` would be more appropriate.

src/api/l_graphics.c
```diff
@@ -754,7 +754,7 @@ static int l_lovrGraphicsNewBuffer(lua_State* L) {
     default: {
       Blob* blob = luax_totype(L, 1, Blob);
       if (blob) {
-        info.length = blob->size / info.stride;
+        info.length = (graphics_size)(blob->size / info.stride);
```
bjornbytes, line 757: Ditto

src/api/l_graphics.c
```diff
@@ -1050,6 +1050,7 @@ static int l_lovrGraphicsNewShader(lua_State* L) {
           case LUA_TNUMBER: flag.id = lua_tointeger(L, -2); break;
           default: lovrThrow("Unexpected ShaderFlag key type (%s)", lua_typename(L, lua_type(L, -2)));
         }
+        lovrCheck(flags.length < UINT32_MAX, "Too many flags");
```
bjornbytes, line 1053: Wouldn't `<=` be a more appropriate condition?  Also, the shader compiler only supports 2000 flags and LÖVR is currently reserving IDs 1000+, so maybe there could be a `MAX_FLAGS` constant since that's a more realistic limit.

src/api/l_graphics_buffer.c
```diff
@@ -146,8 +146,8 @@ void luax_readbufferdata(lua_State* L, int index, Buffer* buffer, char* data) {
   uint32_t dstIndex = luax_optu32(L, index + 2, 1) - 1;
 
   if (blob) {
-    uint32_t limit = MIN(blob->size / stride - srcIndex, info->length - dstIndex);
-    uint32_t count = luax_optu32(L, index + 3, limit);
+    graphics_size limit = (graphics_size)(MIN(blob->size / stride - srcIndex, info->length - dstIndex));
+    graphics_size count = luax_optgraphics_size(L, index + 3, limit);
```
bjornbytes, line 150: Same sort of comment here.  I don't think `count` should be a `graphics_size` since it's not a number of bytes.  At the point these quantities are turned into sizes (basically post-multiplication down below), graphics_size would become the correct type.

src/api/l_graphics_pass.c
```diff
@@ -680,11 +680,11 @@ static int l_lovrPassCopy(lua_State* L) {
 
   if (blob) {
     Buffer* buffer = luax_checktype(L, 3, Buffer);
-    uint32_t srcOffset = luax_optu32(L, 4, 0);
-    uint32_t dstOffset = luax_optu32(L, 5, 0);
+    graphics_size srcOffset = luax_optgraphics_size(L, 4, 0);
```
bjornbytes, line 683: `srcOffset` should be a `size_t` since it represents an offset into CPU memory.  It should be fine to copy a small range of bytes to a buffer, even if those bytes come from 100GB into a blob.

src/api/l_graphics_shader.c
```diff
@@ -28,11 +28,12 @@ static int l_lovrShaderClone(lua_State* L) {
       case LUA_TNUMBER: flag.id = lua_tointeger(L, -2); break;
       default: lovrThrow("Unexpected ShaderFlag key type (%s)", lua_typename(L, lua_type(L, -2)));
     }
+    lovrCheck(flags.length < ~((uint32_t)0), "Too many flags");
```
bjornbytes, line 31: I'd prefer writing this constant as `~0u` to match the rest of the codebase (though this particular one might be better as a smaller constant limit as mentioned above).

src/modules/graphics/graphics.c
```diff
@@ -275,7 +281,7 @@ typedef struct {
   void* next;
   gpu_bundle_pool* gpu;
   gpu_bundle* bundles;
-  uint32_t cursor;
+  graphics_size cursor;
```
bjornbytes, line 284: This should be a `uint32_t` since it's an index into a CPU array.

src/modules/graphics/graphics.c
```diff
@@ -314,17 +320,25 @@ static struct {
...
   Allocator allocator;
 } state;
 
+// GPU pointers may be of a different size from CPU ABI pointers.
+// Currently lovr limits GPU sizes to a 32-bit safe 1GB, but in case
+// this increases later, graphics-related sizes are represented
+// using graphics_size here, a separate gpu_size in gpu, and
+// a luax_checkgraphics_size/luax_optgraphics_size in api.h.
+// graphics.c is responsible for ensuring these types are in sync:
+_Static_assert (sizeof(graphics_size) == sizeof(gpu_size), "graphics_size and gpu_size types must always be the same size");
```
bjornbytes, line 340: Function/macro calls should not have spaces before their parentheses.

src/modules/graphics/graphics.c
```diff
@@ -992,7 +1007,7 @@ Texture* lovrTextureCreate(TextureInfo* info) {
   lovrCheck(info->depth == 1 || info->type != TEXTURE_2D, "2D textures must have a depth of 1");
   lovrCheck(info->depth == 6 || info->type != TEXTURE_CUBE, "Cubemaps must have a depth of 6");
   lovrCheck(info->width == info->height || info->type != TEXTURE_CUBE, "Cubemaps must be square");
-  lovrCheck(measureTexture(info->format, info->width, info->height, info->depth) < 1 << 30, "Memory for a Texture can not exceed 1GB"); // TODO mip?
+  lovrCheck(measureTexture(info->format, info->width, info->height, info->depth) < MAX_BUFFER_SIZE, "Memory for a Texture can not exceed " MAX_BUFFER_SIZE_STR); // TODO mip?
```
bjornbytes, line 1010: I'd prefer if this constant were named something other than `MAX_BUFFER_SIZE` since that makes it seem like it's the maximum size of a buffer when it's really the maximum size of a GPU memory allocation.  Maybe `MAX_GPU_ALLOCATION_SIZE`

src/modules/graphics/graphics.c
```diff
@@ -1015,21 +1030,24 @@ Texture* lovrTextureCreate(TextureInfo* info) {
...
     lovrCheck(info->type != TEXTURE_3D || levelCount == 1, "Images used to initialize 3D textures can not have mipmaps");
 
-    uint32_t total = 0;
+    graphics_size total = 0;
     for (uint32_t level = 0; level < levelCount; level++) {
       levelOffsets[level] = total;
       uint32_t width = MAX(info->width >> level, 1);
       uint32_t height = MAX(info->height >> level, 1);
-      levelSizes[level] = measureTexture(info->format, width, height, info->depth);
+      levelSizes[level] = (graphics_size)measureTexture(info->format, width, height, info->depth);
```
bjornbytes, line 1046: Casts should have spaces after them to match the rest of the code.

src/modules/graphics/graphics.c
```diff
@@ -1225,6 +1243,7 @@ const SamplerInfo* lovrSamplerGetInfo(Sampler* sampler) {
 Blob* lovrGraphicsCompileShader(ShaderStage stage, Blob* source) {
   uint32_t spirv = 0x07230203;
 
+  lovrCheck(source->size < INT_MAX, "Shader cannot be more than 2 GiB");
```
bjornbytes, line 1246: The condition and the error message do not match

src/modules/graphics/graphics.c
```diff
@@ -1402,7 +1421,8 @@ Shader* lovrShaderCreate(ShaderInfo* info) {
   spv_result result;
   spv_info spv[2] = { 0 };
   for (uint32_t i = 0; i < stageCount; i++) {
-    result = spv_parse(info->stages[i]->data, info->stages[i]->size, &spv[i]);
+    lovrCheck(info->stages[i]->size < UINT32_MAX, "Shader too large");
+    result = spv_parse(info->stages[i]->data, (uint32_t)info->stages[i]->size, &spv[i]);
```
bjornbytes, line 1425: I think it would be cleaner to change `spv_parse` to take a `size_t` instead of a `uint32_t` and remove the check/cast.

src/modules/graphics/graphics.c
```diff
@@ -1584,9 +1604,10 @@ Shader* lovrShaderCreate(ShaderInfo* info) {
   shader->info = *info;
   shader->layout = getLayout(slots, shader->resourceCount);
 
+  // Size conversion checked safe earlier in function
   gpu_shader_info gpu = {
-    .stages[0] = { info->stages[0]->data, info->stages[0]->size },
-    .stages[1] = { info->stages[1]->data, info->stages[1]->size },
+    .stages[0] = { info->stages[0]->data, (uint32_t)info->stages[0]->size },
+    .stages[1] = { info->stages[1]->data, (uint32_t)info->stages[1]->size },
```
bjornbytes, line 1610: Ditto, gpu should take a `size_t`

src/modules/graphics/graphics.c
```diff
@@ -1660,15 +1681,16 @@ Material* lovrMaterialCreate(MaterialInfo* info) {
...
       if (block->head != ~0u && gpu_finished(block->list[block->head].tick)) {
-        state.materialBlock = i;
+        state.materialBlock = (uint32_t)i;
         found = true;
         break;
       }
     }
 
     if (!found) {
+      lovrAssert(state.materialBlocks.length < UINT32_MAX, "Too many material blocks in system");
```
bjornbytes, line 1691: The correct limit is `UINT16_MAX` since Material stores this number in a `uint16_t`.

src/modules/graphics/graphics.c
```diff
@@ -2045,16 +2067,17 @@ float lovrFontGetWidth(Font* font, ColoredString* strings, uint32_t count) {
 }
 
 void lovrFontGetLines(Font* font, ColoredString* strings, uint32_t count, float wrap, void (*callback)(void* context, const char* string, size_t length), void* context) {
-  size_t totalLength = 0;
+  uint32_t totalLength = 0;
```
bjornbytes, line 2070: I think it would be cleaner to change the cursor to `size_t` instead of making this a `uint32_t`.

src/modules/graphics/graphics.c
```diff
@@ -3852,7 +3875,7 @@ void lovrPassCopyTallyToBuffer(Pass* pass, Tally* tally, Buffer* buffer, uint32_
 
   for (uint32_t i = 0; i < count; i++) {
     uint32_t index = srcIndex + i;
-    lovrCheck(tally->masks[index / 64] & (1 << (index % 64)), "Trying to copy Tally slot %d, but it hasn't been marked yet", index + 1);
+    lovrCheck(tally->masks[index / 64] & ((uint64_t)1 << (index % 64)), "Trying to copy Tally slot %d, but it hasn't been marked yet", index + 1);
```
bjornbytes, line 3878: I'd prefer `1ull`

src/modules/graphics/graphics.c
```diff
@@ -3930,7 +3953,7 @@ void lovrPassCopyImageToTexture(Pass* pass, Image* image, Texture* texture, uint
   layerOffset += measureTexture(texture->info.format, srcOffset[0], 1, 1);
   size_t pitch = measureTexture(texture->info.format, lovrImageGetWidth(image, srcOffset[3]), 1, 1);
   gpu_buffer* buffer = tempAlloc(gpu_sizeof_buffer());
-  char* dst = gpu_map(buffer, totalSize, 64, GPU_MAP_WRITE);
+  char* dst = gpu_map(buffer, (gpu_size)totalSize, 64, GPU_MAP_WRITE);
```
bjornbytes, line 3956: If `totalSize` exceeds the maximum value representable by `gpu_size`, it seems like this would lead to memory safety issues.  Ideally `measureTexture` returns a `gpu_size` and ensures the result fits in one.

src/modules/graphics/graphics.c
```diff
@@ -4027,14 +4050,15 @@ void lovrPassTock(Pass* pass, Tally* tally, uint32_t index) {
 // Helpers
 
 static void* tempAlloc(size_t size) {
+  lovrCheck(size < SIZE_MAX-MAX_FRAME_MEMORY, "Allocation over " MAX_FRAME_MEMORY_STR);
```
bjornbytes, line 4053: Although our error messages aren't great, I think they should be slightly nicer than "Shader too big" and "Allocation over 1GiB".  Also, binary operators should have spaces around them.  If possible this check should be combined into the one below, or replace it.

src/modules/graphics/graphics.c
```diff
@@ -4027,14 +4050,15 @@ void lovrPassTock(Pass* pass, Tally* tally, uint32_t index) {
...
   while (state.allocator.cursor + size > state.allocator.length) {
     lovrAssert(state.allocator.length << 1 <= MAX_FRAME_MEMORY, "Out of memory");
     os_vm_commit(state.allocator.memory + state.allocator.length, state.allocator.length);
     state.allocator.length <<= 1;
   }
 
-  uint32_t cursor = ALIGN(state.allocator.cursor, 8);
-  state.allocator.cursor = cursor + size;
+  graphics_size cursor = ALIGN(state.allocator.cursor, 8);
+  state.allocator.cursor = (graphics_size)(cursor + size);
```
bjornbytes, line 4061: These should not be `graphics_size`s because they represent CPU memory.

bjornbytes (commented):
Well, I was very pedantic, but this seems like one of the rare times when it's appropriate.  I'm concerned that a lot of places that should be using `graphics_size` are not using it.  Basically every time there is a `gpu_map` call, there would need to be a bounds check or cast on that, in addition to buffer/texture creation.

bjornbytes:
I fixed all the graphics.c warnings in a2ee485d4130ed9ed14e69c03cae2dcafc485c0d and marked you as a co-author

bjornbytes:
Since all but 2 of these are resolved, and I'm not opting for the `gpu_size` change atm, I think this can be closed...

PR #570, opened 2022-07-01
Fix various compiler warnings in MSVC (non graphics edition)

mcclure:
- Put in casts/checks in audio code when assigning size_t to 32 bit
- () is different from (void)
- Turned off warnings for anonymous unions and negating unsigned integers which were technically accurate but unhelpful (and interfered with bit conversion and a weird bit math construct in audio.c) (CMakeLists only)

Left out graphics related warnings because that's in the other patch.

Left out weirdo image.c use-undefined-variable bug because Bjorn said he was handling it.


CMakeLists.txt
```diff
@@ -597,7 +597,7 @@ if(WIN32)
   target_sources(lovr PRIVATE src/core/os_win32.c)
   target_sources(lovr PRIVATE etc/lovr.rc)
   if (MSVC)
-    set_target_properties(lovr PROPERTIES COMPILE_FLAGS "/wd4244 /MP")
+    set_target_properties(lovr PROPERTIES COMPILE_FLAGS "/wd4244 /wd4146 /wd4116 /MP")
```
bjornbytes, line 600: From Microsoft's description, I'm having trouble understanding what `C4116` actually is.  Was there somewhere this warning was appearing?

Since 4146 is sometimes useful, it might be good to silence the warning only in audio.c with `set_source_files_properties`.  Feel free to skip that though.

src/api/l_data_sound.c
```diff
@@ -51,8 +51,9 @@ static int l_lovrSoundGetSampleRate(lua_State* L) {
 
 static int l_lovrSoundGetByteStride(lua_State* L) {
   Sound* sound = luax_checktype(L, 1, Sound);
-  uint32_t stride = lovrSoundGetStride(sound);
-  lua_pushinteger(L, stride);
+  size_t stride = lovrSoundGetStride(sound);
+  lovrCheck(stride < UINT32_MAX, "Sound contains impossibly many channels");
```
bjornbytes, line 55: I don't know, it feels really weird to me to check this on egress.  Like, I create a Sound and everything's great, but then I call a getter on it and it throws an error.  Also I'm not convinced that testing against `unit32_t` is correct here, since `lua_pushinteger` takes a `lua_Integer` which is a `ptrdiff_t`.

CMakeLists.txt
mcclure: Sorry, C4116 was relevant to util.c not audio.c. util.c has:

`uint32_t u = ((union { float f; uint32_t u; }) { f }).u;`

The union here is being used for punning, but because it is anonymous, Microsoft apparently believes that there is no way it could have an effect. Either they literally forgot unions existed or, more likely, they introduced this warning for C++ (where union type punning works differently) and then forgot to remove it for C11. Or maybe CMake isn't building properly in C11 mode to start with, I haven't tested.

I can do the set_source_files_properties thing.

src/api/l_data_sound.c
mcclure: 1. This is, uh, a typo actually. I meant to add the check to a different function but I got a line number confused. Will remove.
2. The "check on egress" problem remains because of the below patch to sound:getFrames. I can fix that by switching to lua_optinteger instead of lua_optu32 if you're good with that.

    Overall there is a problem that we have API code that has inherent bounds or has unnecessary bounds due to the current typing, and we have data classes which are capable of representing anything up to SIZE_MAX, and it would make the most sense for either the data classes or the constructors to enforce a contract that they are no bigger than the "safe" size but there's no... like, there's no indication of who enforces those contracts or where. Maybe it could go in the header files? If the sizes can't be enforced by the type system I think it's best it at least be written down somewhere.
3. You're right about ptrdiff_t (again, this one change is outright a mistake) but there is a little bit of complexity there because any number which is returned to lua DOES have an upper bound and it's less than SIZE_MAX. PTRDIFF_MAX is usually lower than SIZE_MAX, and on luajit I would expect the upper bound to be something like 2^52 because it never(?) creates 32 bit integer types. I'm not sure if we need to worry about that (I can believe someone would load a 5 GB wav file either maliciously or in a pathological case. But 2^52 bytes is... 4 petabytes?) but if we choose to worry about it, I actually don't know how to query what the max value is.

mcclure:
Should address all comments

src/modules/data/sound.c
```diff
@@ -68,8 +68,10 @@ static uint32_t lovrSoundReadMp3(Sound* sound, uint32_t offset, uint32_t count,
   }
 
   uint32_t channels = lovrSoundGetChannelCount(sound);
-  size_t samples = mp3dec_ex_read(sound->decoder, data, count * channels);
-  uint32_t frames = samples / channels;
+  uint32_t readSamples = (uint32_t)(count * channels);
+  lovrCheck(readSamples/count == channels, "Cannot request more than 2^32-1 samples of sound data at a time"); // Check for overflow
```
bjornbytes, line 72: It's not safe to throw in this function since it can run in the audio thread.  It could cap the number of frames it tries to read and return a smaller value, which would cause the mixer to retry with the remaining number of frames.

PR #573, opened 2022-07-31
Fix cylinder's triangle winding and normals

jmiskovic:
Issue became apparent with back face culling:
```
function lovr.draw(pass)
  pass:setCullMode('back')
  local pose = mat4(0, 1.5, -1)
        :scale(0.2)
        :rotate(lovr.timer.getTime(), 0,1,0)
  pass:cylinder(pose, true, nil, nil, 12)
end
```
image


bjornbytes:
Thank you thank you

PR #578, opened 2022-08-18
Fix SPV type conversion for vec

jmiskovic:


src/core/spv.c
```diff
@@ -464,24 +464,24 @@ static spv_result spv_parse_push_constants(spv_context* spv, const uint32_t* op,
     if (OP_CODE(type) == 22 && type[2] == 32) { // OpTypeFloat
       if (columnCount > 2 && columnCount < 4 && componentCount == columnCount) {
```
bjornbytes, line 465: One more here

jmiskovic:
uuf I was sleeping already, let me fix this quickly

bjornbytes:
Thanks!

PR #582, opened 2022-09-21
Fix sending strings with \0 through channels

jmiskovic:
A null-char is valid part of Lua string. When such a string is sent through the channel, its length should be stored as well to be able to correctly reconstruct it on the other thread.

The bug was triggered with this code:

    s1 = 'a \0 b'
    print(#s1) -- 5
    ch:push(s1)
    s2 = ch:pop()
    print(#s2) -- 2


bjornbytes:
Got it, thanks!

PR #585, opened 2022-10-06
Use  the internal glslang submodule for compiling shaders

### LINKED ISSUE #579: Possibly build glslangValidator before shader compilation
As communicated in Slack before opening this issue, the Vulkan merge means that CMake uses the system's `glslangValidator` executable, even though `glslang` is vendored in the source tree. It would be welcome to build that and use the vendored executable to prevent system inconsistencies (my system's `glslang` is 8.13.3743, which is over 2 years ago, and so, incompatible with the CMake script's usage of glslangValidator, for example).

Leaving it as an issue just in case someone gets unfortunate enough to have the same problem. :P

rpavlik:
Fixes #579 and makes it build on Debian 11 (Bullseye) - I did no other testing. I suspect it's safe but maybe should test runtime shader compilation since I did touch that bit.


bjornbytes:
Thank you!  :purple_heart:

PR #587, opened 2022-10-14
LOVR can be used as a sumbodule

brainrom:
Massive project with plugins, modules, etc. can be build by CMake, where LOVR is just one of modules.
The proposed directory structure:
```
- project
  - main.lua
  - CMakeLists.txt (CMakeLists.txt.project)
  - lovr
    - CMakeLists.txt (LOVR's CMakeLists)
```


bjornbytes:
I merged parts of this in 7d3cc45cc23f29785cf30a12e7cd87967802bf0f which should hopefully be all the CMakeLists changes needed.  I'm not interested in making the struct internals public in headers right now, but if there are any internals you need access to that are missing from the C API I'd still like to try to expose those.  For the example project CMake file, if you put it in a gist I can reference it from the docs.  Although it's not everything proposed in this PR hopefully it makes your project a little easier to maintain.  Thanks for opening this.

PR #592, opened 2022-10-24
Add physics heightfield shape

jmiskovic:
This is sitting on my drive for too long. It works, but needs bit more effort before it's ready to merge.

It is not possible to change the pose of heightfield after creating it. I believe the `dGeomSetPosition` needs to be used (need to test), so the `setPose()` would have to be specialized for the heightfield collider case. I also skipped on using the recommended `dGeomHeightfieldDataSetBounds()` as it didn't behave correctly in my tests.

I've only tested on "r32f" heightmap image format as I can store actual terrain height into each pixel. Other formats that support `getPixel()` should be OK to use, with the `verticalScaling` parameter mapping the 0 to 1 pixel value to actual height. Changing the terrain is not implemented at the moment.

I would like to also support the callback mode where ODE constantly queries the Lua side about height at coordinates. Useful for sea with waves, but it could pan out to be too slow to use.

### DIFF
--- a/src/api/l_physics.c
+++ b/src/api/l_physics.c
@@ -10,6 +10,7 @@ StringEntry lovrShapeType[] = {
   [SHAPE_CAPSULE] = ENTRY("capsule"),
   [SHAPE_CYLINDER] = ENTRY("cylinder"),
   [SHAPE_MESH] = ENTRY("mesh"),
+  [SHAPE_TERRAIN] = ENTRY("terrain"),
   { 0 }
 };
 
--- a/src/api/l_physics_shapes.c
+++ b/src/api/l_physics_shapes.c
@@ -13,6 +13,7 @@ void luax_pushshape(lua_State* L, Shape* shape) {
     case SHAPE_CAPSULE: luax_pushtype(L, CapsuleShape, shape); break;
     case SHAPE_CYLINDER: luax_pushtype(L, CylinderShape, shape); break;
     case SHAPE_MESH: luax_pushtype(L, MeshShape, shape); break;
+    case SHAPE_TERRAIN: luax_pushtype(L, TerrainShape, shape); break;
     default: lovrUnreachable();
   }
 }
--- a/src/api/l_physics_world.c
+++ b/src/api/l_physics_world.c
@@ -1,5 +1,6 @@
 #include "api.h"
 #include "physics/physics.h"
+#include "data/image.h"
 #include "util.h"
 #include <lua.h>
 #include <lauxlib.h>
@@ -40,6 +41,16 @@ static void raycastCallback(Shape* shape, float x, float y, float z, float nx, f
   lua_pushnumber(L, ny);
   lua_pushnumber(L, nz);
   lua_call(L, 7, 0);
+}
+
+static float terrainCallback(lua_State * L, int fn_index, float x, float z) {
+  lua_pushvalue(L, fn_index);
+  lua_pushnumber(L, x);
+  lua_pushnumber(L, z);
+  lua_call(L, 2, 1);
+  float height = luax_checkfloat(L, -1);
+  lua_remove(L, -1);
+  return height;
 }
 
 static int l_lovrWorldNewCollider(lua_State* L) {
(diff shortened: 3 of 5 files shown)


jmiskovic:
```
-- Example for generating procedural terrain with physical simulation
local shaderCode = {[[
/* VERTEX shader */
layout(location = 0) out vec4 fragmentView;
layout(location = 1) out vec3 worldView;

vec4 lovrmain() {
  worldView = VertexPosition.xyz;
  fragmentView = Projection * Transform * VertexPosition;
  return DefaultPosition;
} ]], [[
/* FRAGMENT shader */
layout(location = 0) in vec4 fragmentView;
layout(location = 1) in vec3 worldView;

Constants {
  vec3 fogColor;
};

vec4 lovrmain() {
  float fogAmount = 1.f - exp(-length(fragmentView) * 0.01);
  vec3 surface = Color.rgb * getPixel(ColorTexture, worldView.xz * 0.01).rgb;
  vec4 color = vec4(mix(surface, fogColor, fogAmount), Color.a);
  return color;
}]]}

local shader = lovr.graphics.newShader(unpack(shaderCode))


local function grid(size, subdivisions) -- construct vertices for subdivided plane
  local vertices = {}
  local indices  = {}
  local step = size / (subdivisions - 1)
  for z = -size / 2, size / 2, step do
    for x = -size / 2, size / 2, step do
      table.insert(vertices, {x, 0, z})
      table.insert(vertices, {x, 0, z + step})
      table.insert(vertices, {x + step, 0, z})

      table.insert(vertices, {x, 0, z + step})
      table.insert(vertices, {x + step, 0, z + step})
      table.insert(vertices, {x + step, 0, z})
    end
  end
  return vertices
end


local function terrain(x, z) -- calculate terain height at any location
  local lacunarity = 3
  local gain = 0.5
  local octaves = 3
  local sum = 0
  local freq = 0.005
  local amp = 30
  local h
  for i = 1, octaves do
    -- h = 1 - math.abs(lovr.math.noise(x * freq, z * freq))
    h = lovr.math.noise(x * freq, z * freq)
    sum = sum + amp * h
    freq = freq * lacunarity
    amp = amp * gain
  end
  return -30 + sum
end


local terrain_scale = 200
local grid_subdivisions = 100

-- create mesh for rendering terrain
local vertices = grid(terrain_scale, grid_subdivisions)
for i, v in ipairs(vertices) do -- adjust subdivided grid heights to the terrain
  local x, h, z = unpack(v)
  v[2] = terrain(x, z)
end
local mesh_buffer = lovr.graphics.newBuffer(vertices, { 'vec3:VertexPosition' })

-- create heightmap for physical collider
local image_size = 256
local image = lovr.data.newImage(image_size, image_size, 'r32f')
for px = 0, image_size - 1 do
  for pz = 0, image_size - 1 do
    local x = ((px+1) / image_size - 0.5) * terrain_scale
    local z = ((pz+1) / image_size - 0.5) * terrain_scale
    local h = terrain(x, z)
    image:setPixel(px, pz, h)
  end
end

local terrain_collider = world:newHeightfieldCollider(image, terrain_scale, 1)
local boxes = {}
for i = 1,100 do
  local box = world:newBoxCollider(
    lovr.math.randomNormal(20, 0),
    lovr.math.randomNormal(5, 20),
    lovr.math.randomNormal(20, 0),
    1,1,1)
  table.insert(boxes, box)
end


function lovr.update(dt)
  world:update(1 / 60)
end


function lovr.draw(pass)
  pass:setShader(shader)
  lovr.graphics.setBackgroundColor(0x74A2FF)
  pass:send('fogColor', {0, 0, 0})
  pass:setColor(0x508547)
  pass:mesh(mesh_buffer, 0, 0, 0)
  pass:setWireframe(true)
  pass:setColor(0,0,0, 0.1)
  pass:mesh(mesh_buffer, 0, 0.2, 0)
  pass:setWireframe()
  pass:setColor(0x6A5639)
  for i, box in ipairs(boxes) do
    pass:cube(mat4(box:getPose()))
  end
  pass:setShader()
end

```

bjornbytes:
Naming bikeshed:

- Heightfield
- Heightmap
- Terrain
- Ground

Heightfield/Heightmap are good, and the most generic/correct, but TerrainShape is kind of cute.

bjornbytes:
More thoughts/notes from testing it:

- Calling :setPose on the heightfield collider worked for me
- It's kind of weird to have a heightfield on a Collider.  You'll almost always want it to be static, and it doesn't make sense to apply forces, attach joints, extra shapes, etc.  Calling :setMass on it doesn't make sense and will accidentally make it non-kinematic again.  However, Colliders are the only place where you can set friction/restitution right now, which might be important for the terrain.  I thought about making the heightmap some sort of weird global property like `World:setHeightmap(image)`, but this just seems too confusing.  So I think the current approach is better.
- Even without support for ode's dynamic callback, it could be nice to have a callback option for creating the heightfield, as long as it provides a speedup over creating an extra image and calling setPixel a bunch.  Internally this would still just allocate some memory and call the Lua callback repeatedly.
- Today, it's pretty common to create a "ground" collider from a carefully-positioned box shape.  Maybe there could be a shorthand for "I just want an infinite flat ground plane" for heightfields.  It could maybe use the plane geom.  Maybe too complicated/unnecessary.
- Would be nice to get newHeightfieldShape and newMeshShape eventually, but that is a nitpick.

Overall it's really good, seems merge-able to me.

jmiskovic:
You are right about changing the pose, I used it incorrectly with mat4. This was my main obstacle for merging it, thanks for testing.

The example has too many steps to get it working. It would be amazing if Lua could access the lovr's subdivided grid and not have to re-implement (although most likely the hex grid would be better for collisions and for hiding the underlying geometry). I also like the idea of easily creating the ground. At very least we can make the terrain flat if nil is passed for image, with size specified in horizontalScale.

Setting mass is hilarious but the terrain does work as dynamic collider. Maybe somebody finds use for it, for making a snow globe or something.

How about SurfaceShape for a name? It covers all the cases I can think of: terrain, seabed, sea surface, ceiling of caves and bumpy wall.

jmiskovic:
Renamed from heightfield shape to surface shape.

Removed thickness param, now it is always 10 meters. Increasing it would prevent fast bullets from tunneling through but would also increase number of AABB false positives. This thickness can resist up to ~1000 m/s impacts on 1 / 60 update delta.

Added support for terrain height callback, static mode only.

Changed API param ordering:

```
-- make a 100x100 flat surface
world:newSurfaceCollider(100)

-- read 100x100 terrain from image data
world:newSurfaceCollider(100, r32f_image)

-- read 200x200 terrain from R channel and scale it to 100 height
world:newSurfaceCollider(200, rgb_image, 100)

-- 15x15 terrain sampled and cached from callback fn in 50x50 points
world:newSurfaceCollider(15, function(x, z) return lovr.math.noise(x, z) end, 50)
-- x,z are in world coordinates, from -size / 2 to size / 2
-- if not specified, sampling density defaults to 100 points per each axis
```

bjornbytes:
I think Terrain is preferred to Surface since "surface" could mean the surface of any object, and surface is a bit more overloaded (window/display surface is in gpu internals, `Surface` is also a struct in shaders) so it risks clashing with other names.  Terrain isn't quite as generic (weird to use it for e.g. a wall collider) but does make it immediately clear what it's for.

jmiskovic:
Renamed

bjornbytes:
Thanks!  This is a really exciting feature

PR #600, opened 2022-11-09
fix wrong loop limit in lovr.graphics.newShader

s-ol:
stumbled over this odd-looking line while taking a look at the PBR Lighting sample.


bjornbytes:
This was intentional -- it iterates until the `{ 0 }` 0-length entry is hit.  It was necessary at one point because there were some "internal" DefaultShaders that didn't have Lua enumerants, so it was necessary to loop over only the entries in the `lovrDefaultShader` string table instead of all of the DefaultShader values.  However, this is no longer the case and `DEFAULT_SHADER_COUNT` works just as well and is a bit more clear.

PR #603, opened 2022-11-13
GLTF fixes

s-ol:
from matrix chat:

> I've noticed some off-by-ones with GLTF model node indices
> but I'm not sure which is the right place to fix it:
> * the GLTF model parser puts its nodes into the modelData->nodes array starting at zero (of course) and the root node last, but in modelData->nodeMap the indices that are stored start counting from 1
> * luax_checknode in l_data_modelData.c doesn't account for this and therefore all getNode...() APIs from there are off-by-one when using strings to access the nodes
>
>the question is, should the map in the GLTF model parser be fixed or should the map output be used - 1 for indexing model->nodes[] (and if so, why?)
> also: the animation & material maps have the same map_set call / index logic, so they might be off-by-one as well? Haven't checked their Lua APIs yet to see if there is proper conversion happening here
>
> ah, model->materialCount - i starts from 0, and I'm guessing this is true for animations too
> for model->nodeCount - i this is probably off-by-one because the rootNode (with highest index) is not accounted for

### DIFF
--- a/src/api/l_graphics_model.c
+++ b/src/api/l_graphics_model.c
@@ -30,7 +30,12 @@ static uint32_t luax_checkanimation(lua_State* L, int index, Model* model) {
       lovrCheck(animationIndex != MAP_NIL, "ModelData has no animation named '%s'", name);
       return (uint32_t) animationIndex;
     }
-    case LUA_TNUMBER: return lua_tointeger(L, index) - 1;
+    case LUA_TNUMBER: {
+      uint32_t animation = luax_checku32(L, index) - 1;
+      ModelData* data = lovrModelGetInfo(model)->data;
+      lovrCheck(animation < data->animationCount, "Invalid animation index '%d'", animation + 1);
+      return animation;
+    }
     default: return luax_typeerror(L, index, "number or string"), ~0u;
   }
 }
@@ -45,7 +50,12 @@ uint32_t luax_checknodeindex(lua_State* L, int index, Model* model) {
       lovrCheck(nodeIndex != MAP_NIL, "ModelData has no node named '%s'", name);
       return (uint32_t) nodeIndex;
     }
-    case LUA_TNUMBER: return lua_tointeger(L, index) - 1;
+    case LUA_TNUMBER: {
+      uint32_t node = luax_checku32(L, index) - 1;
+      ModelData* data = lovrModelGetInfo(model)->data;
+      lovrCheck(node < data->nodeCount, "Invalid node index '%d'", node + 1);
+      return node;
+    }
     default: return luax_typeerror(L, index, "number or string"), ~0u;
   }
 }
--- a/src/api/l_graphics_model.c
+++ b/src/api/l_graphics_model.c
@@ -143,14 +143,12 @@ static int l_lovrModelGetNodeScale(lua_State* L) {
   Model* model = luax_checktype(L, 1, Model);
   uint32_t node = luax_checknodeindex(L, 2, model);
   OriginType origin = luax_checkenum(L, 3, OriginType, "root");
-  float position[4], scale[4], rotation[4], angle, ax, ay, az;
+  float position[4], scale[4], rotation[4];
   lovrModelGetNodeTransform(model, node, position, scale, rotation, origin);
-  quat_getAngleAxis(rotation, &angle, &ax, &ay, &az);
-  lua_pushnumber(L, angle);
-  lua_pushnumber(L, ax);
-  lua_pushnumber(L, ay);
-  lua_pushnumber(L, az);
-  return 4;
+  lua_pushnumber(L, scale[0]);
+  lua_pushnumber(L, scale[1]);
+  lua_pushnumber(L, scale[2]);
+  return 3;
 }
 
 static int l_lovrModelSetNodeScale(lua_State* L) {
--- a/src/modules/data/modelData_gltf.c
+++ b/src/modules/data/modelData_gltf.c
@@ -888,7 +888,7 @@ ModelData* lovrModelDataInitGltf(ModelData* model, Blob* source, ModelDataIO* io
           scale[2] = NOM_FLOAT(json, token);
         } else if (STR_EQ(key, "name")) {
           gltfString name = NOM_STR(json, token);
-          map_set(&model->nodeMap, hash64(name.data, name.length), model->nodeCount - i);
+          map_set(&model->nodeMap, hash64(name.data, name.length), node - model->nodes);
           memcpy(model->chars, name.data, name.length);
           node->name = model->chars;
           model->chars += name.length + 1;
--- a/src/modules/graphics/graphics.c
+++ b/src/modules/graphics/graphics.c
@@ -2797,13 +2797,11 @@ const ModelInfo* lovrModelGetInfo(Model* model) {
 
 uint32_t lovrModelGetNodeDrawCount(Model* model, uint32_t node) {
   ModelData* data = model->info.data;
-  lovrCheck(node < data->nodeCount, "Invalid model node index %d", node + 1);
   return data->nodes[node].primitiveCount;
 }
 
 void lovrModelGetNodeDraw(Model* model, uint32_t node, uint32_t index, ModelDraw* mesh) {
   ModelData* data = model->info.data;
-  lovrCheck(node < data->nodeCount, "Invalid model node index %d", node + 1);
   lovrCheck(index < data->nodes[node].primitiveCount, "Invalid model node draw index %d", index + 1);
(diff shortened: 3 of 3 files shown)


s-ol:
Something weird is going on with Model:getNodePosition/Rotation/Scale as well, still investigating that

s-ol:
Okay, I figured out one bit: Model:getNodeScale() was a copy of Model:getNodeOrientation(), so those bogus values threw me off for a bit. But I'm still not managing to copy a node transform exactly:

```Lua
function lovr.load()
  model = lovr.graphics.newModel('test_cube.glb')

  rn = model:getRootNode()
  print("ORIGINAL TRANSFORM")
  print(model:getNodeTransform(rn, 'root'))

  -- copy and apply transform matrix
  local pos = vec3(model:getNodePosition(rn, 'root'))
  local scale = vec3(model:getNodeScale(rn, 'root'))
  local rot = quat(model:getNodeOrientation(rn, 'root'))
  local basexf = lovr.math.newMat4(pos, scale, rot)
  -- print('pos', pos)
  -- print('scale', scale)
  -- print('rot', rot)
  model:setNodeTransform(rn, basexf)

  print("RESTORED TRANSFORM")
  print(model:getNodeTransform(rn, 'root'))
end

function lovr.draw(pass)
  pass:draw(model)
end
```

with a test model of a transformed cube made in blender (see attachment), I'd expect the transform before/after the `Model:setNodeTransform()` to match, but it doesnt:

```
ORIGINAL TRANSFORM
1	3	-2	0.40000000596046	0.6000000834465	0.5	1.6285449266434	0.58154702186584	0.54790300130844	-0.60133630037308
RESTORED TRANSFORM
1	3	-2	0.4995265007019	0.42231246829033	0.58491510152817	1.9710187911987	0.56878536939621	0.57249844074249	-0.59053260087967
```

Lua + GLB file: [test_nodexform.zip](https://github.com/bjornbytes/lovr/files/9997277/test_nodexform.zip)

Is this a transform order issue between lm.mat4() and the way the GLTF specification specifies transforms? Or a bug in the way the matrix is decomposed into TRS components?

If its the first, lvor definitely needs to provide some tools to copy the original transformation matrix so that things can be manually animated relative to their original position (since Model:setNode...() are all in modelroot-space). In any case, I think a Model:getNodeTransformMatrix() function would be very useful.

s-ol:
Seems like order is part of it; with this snippet it's already closer (correct scale), but the rotation is still off:

```lua
  basexf = lovr.math.newMat4()
  basexf:translate(pos)
  basexf:rotate(rot)
  basexf:scale(scale)
  model:setNodeTransform(rn, basexf)
  local basexf = lovr.math.newMat4(pos, scale, rot)
```

```
ORIGINAL TRANSFORM
1	3	-2	0.40000000596046	0.6000000834465	0.5	1.6285449266434	0.58154702186584	0.54790300130844	-0.60133630037308
manual TRS TRANSFORM
1	3	-2	0.40000009536743	0.6000000834465	0.50000005960464	1.9699511528015	0.56074678897858	0.58925795555115	-0.58166825771332
```

src/modules/data/modelData_gltf.c
```diff
@@ -888,7 +890,7 @@ ModelData* lovrModelDataInitGltf(ModelData* model, Blob* source, ModelDataIO* io
           scale[2] = NOM_FLOAT(json, token);
         } else if (STR_EQ(key, "name")) {
           gltfString name = NOM_STR(json, token);
-          map_set(&model->nodeMap, hash64(name.data, name.length), model->nodeCount - i);
+          map_set(&model->nodeMap, hash64(name.data, name.length), model->nodeCount - extraNodes - i);
```
bjornbytes, line 893: I wonder if `node - model->nodes` would be a more clear/robust way to write this without needing `extraNodes`.

src/api/l_graphics_model.c
```diff
@@ -45,7 +50,12 @@ uint32_t luax_checknodeindex(lua_State* L, int index, Model* model) {
       lovrCheck(nodeIndex != MAP_NIL, "ModelData has no node named '%s'", name);
       return (uint32_t) nodeIndex;
     }
-    case LUA_TNUMBER: return lua_tointeger(L, index) - 1;
+    case LUA_TNUMBER: {
+      uint32_t node = luax_checku32(L, index) - 1;
+      ModelData* data = lovrModelGetInfo(model)->data;
+      lovrCheck(node < data->nodeCount, "Invalid node index '%d'", node + 1);
+      return node;
+    }
```
bjornbytes, line 58: If the bounds checks are getting added here, it probably makes sense to remove bounds checks from the C implementations (`lovrModelAnimate`, `lovrModelGetNodeDrawCount`, etc.).  It probably makes more sense to do the checks in one place in the Lua bindings instead of having to remember to do it in each C function.

bjornbytes:
Some relevant bits of the glTF spec:

> When matrix is defined, it **MUST** be decomposable to TRS properties.

and

> TRS properties **MUST** be converted to matrices and postmultiplied in the T * R * S order; first the scale is applied to the vertices, then the rotation, and then the translation.

`updateModelTransforms` does this, which should be the correct order:

```
  mat4_init(global, parent);
  mat4_translate(global, T[0], T[1], T[2]);
  mat4_rotateQuat(global, R);
  mat4_scale(global, S[0], S[1], S[2]);
```

Still investigating...

bjornbytes:
Ok, there's indeed an order issue with some variants of `Mat4:set` (TSR instead of TRS).  This was a quick patch to fix:

```diff
diff --git src/api/l_math_vectors.c src/api/l_math_vectors.c
index 5b4fe8c4..8fba62fa 100644
--- src/api/l_math_vectors.c
+++ src/api/l_math_vectors.c
@@ -1762,20 +1762,26 @@ int l_lovrMat4Set(lua_State* L) {
         float az = luax_checkfloat(L, index++);
         mat4_rotate(m, angle, ax, ay, az);
       } else {
+        float sx, sy, sz;
+
         if (vectorType == V_VEC3) {
-          m[0] = v[0];
-          m[5] = v[1];
-          m[10] = v[2];
+          sx = v[0];
+          sy = v[1];
+          sz = v[2];
           index++;
         } else if (lua_type(L, index) == LUA_TNUMBER) {
-          m[0] = luax_checkfloat(L, index++);
-          m[5] = luax_checkfloat(L, index++);
-          m[10] = luax_checkfloat(L, index++);
+          sx = luax_checkfloat(L, index++);
+          sy = luax_checkfloat(L, index++);
+          sz = luax_checkfloat(L, index++);
+        } else {
+          sx = sy = sz = 1.f;
         }
 
         float rotation[4];
         luax_readquat(L, index, rotation, NULL);
         mat4_rotateQuat(m, rotation);
+
+        mat4_scale(m, sx, sy, sz);
       }
     }
   }
```

Fixes scale, but rotation is still off, which is where you're at now

bjornbytes:
I think the issue is that `mat4_getOrientation` doesn't work properly when the matrix has scale.  Normalizing the result quaternion doesn't help either.  I think the matrix columns need to be divided by scale before the rotation can be extracted.  This maybe implies that the decomposition functions should be combined.

`mat4_getAngleAxis` correctly divides  by scale and doesn't have this issue, so doing `mat4:unpack` will properly decompose the rotation.

bjornbytes:
`mat4_getOrientation` is only used for `Model:getNodeTransform`, `Model:setNodeTransform`, `Model:resetNodeTransforms`, so the issue is isolated to there.

src/api/l_graphics_model.c
s-ol: True, I checked a couple of them and there were some that didn't do any bounds checking in the implementation so I thought lifting it out might be good to unify this. I'll do a pass and remove the checks

s-ol: (In particular, all the node transform functions that call lovrModelGet/SetNodeTransform internally don't do bounds checking currently)

src/modules/data/modelData_gltf.c
s-ol: Should we apply the same to `model->textures` and `model->materials` then?

s-ol:
> I think the matrix columns need to be divided by scale before the rotation can be extracted. This maybe implies that the decomposition functions should be combined.

I see; my linalg is not strong so I think it might be better if you deal with this outside of this PR.

bjornbytes:
Sounds good.  Thank you for the PR, I'll let you know when scaled nodes are fixed!

src/modules/data/modelData_gltf.c
bjornbytes: Added

bjornbytes:
Ok, scaling should be fixed now

PR #604, opened 2022-11-16
implement XR_MSFT_controller_model

s-ol:
This appears to be working pretty well already, but I'm testing against my own (very WIP) Monado implementation which doesn't have any rotating elements yet, so I'd wait at least until that's tested to merge.

There are also no safeguards against passing wrong models into love.headset.animate and the XrControllerModelProperties and GLTF node lookup should probably be cached. Both these issues will probably require attaching extra metadata to the Model.

Suggestions welcome :slightly_smiling_face:


bjornbytes:
Wow, I can't believe it worked out of the box!  I guess that's the power of glTF.

src/modules/graphics/graphics.c
```diff
@@ -2925,7 +2925,7 @@ void lovrModelGetNodeTransform(Model* model, uint32_t node, float position[4], f
   }
 }
 
-void lovrModelSetNodeTransform(Model* model, uint32_t node, float position[4], float scale[4], float rotation[4], float alpha) {
+void lovrModelSetNodeTransform(Model* model, uint32_t node, float position[3], float scale[3], float rotation[4], float alpha) {
```
bjornbytes, line 2928: 4 floats is the intent here because lovr currently uses 4 floats for vec3 (helps compiler with vectorization).

src/modules/headset/headset_openxr.c
```diff
@@ -1773,6 +1823,63 @@ static bool openxr_animate(Device device, Model* model) {
...
+  if ((tracker = getHandTracker(device))) {
+    return openxr_animateFB(tracker, model);
+  }
+
+  XrControllerModelKeyMSFT modelKey;
+  if ((modelKey = getControllerModelKey(device))) {
+    return openxr_animateMSFT(modelKey, model);
+  }
+
+  return NULL;
```
bjornbytes, line 1880: This should be `false`

src/modules/headset/headset_openxr.c
```diff
@@ -1773,6 +1823,63 @@ static bool openxr_animate(Device device, Model* model) {
   return true;
 }
 
+static bool openxr_animateMSFT(XrControllerModelKeyMSFT modelKey, Model* model) {
+  XrControllerModelNodePropertiesMSFT nodeProperties[16];
```
bjornbytes, line 1827: OpenXR might make us fill in `type` for each element of `nodeProperties` and `nodeStats`, sort of like `extensionProperties` is doing.  Not 100% sure on it though.

src/modules/headset/headset_openxr.c
```diff
@@ -1773,6 +1823,63 @@ static bool openxr_animate(Device device, Model* model) {
...
+    const char* name = nodeProperties[i].nodeName;
+    ModelData* data = lovrModelGetInfo(model)->data;
+    uint64_t nodeIndex = map_get(&data->nodeMap, hash64(name, strlen(name)));
+    lovrCheck(nodeIndex != MAP_NIL, "ModelData has no node named '%s'", name);
+
+    float position[4], scale[4], rotation[4];
+    lovrModelGetNodeTransform(model, nodeIndex, position, scale, rotation, ORIGIN_ROOT);
+
+    vec3_add(position, (vec3)&nodeStates[i].nodePose.position);
+    quat_mul(rotation, rotation, (quat)&nodeStates[i].nodePose.orientation);
```
bjornbytes, line 1861: Based on the wording of the extension

> [the pose is] the XrPosef of the node in it’s parent node space

I would have expected the pose in the nodeState to overwrite the node's local pose, not get added as an offset.  i.e. `lovrModelSetNodeTransform(model, nodeIndex, &nodeStates[i].nodePose.position.x, ...)`).

src/modules/headset/headset_openxr.c
```diff
@@ -1685,13 +1706,42 @@ static ModelData* openxr_newModelData(Device device, bool animated) {
...
+  unsigned char* modelData = malloc(size);
+  if (!modelData) return NULL;
+
+  if (XR_FAILED(xrLoadControllerModelMSFT(state.session, modelKey, size, &size, modelData))) {
+    free(modelData);
+    return NULL;
   }
 
+  Blob* blob = lovrBlobCreate(modelData, size, "Controller Model Data");
+  ModelData* model = lovrModelDataCreate(blob, NULL); // @TODO: null not allowed here
```
bjornbytes, line 1724: Hm, maybe `NULL` could be allowed and the model loader would just throw if there were any references to external files.  This can be left for the future though.

bjornbytes:
Regarding node lookups: since the controller keys are cached and per-device, we would be able to globally cache the node lookups for each device (something like 16 `uint32_t` node indices or something).  It could be populated the first time a model is created.  This would make the `animate` function much more strict about the inputs it accepts though.  Alternatively, node name hashes could be stored to make `animate` more flexible (only node names need to match, not node indices), but I'm not sure how much overhead that would end up saving.

src/modules/graphics/graphics.c
s-ol: hm, I figured there was a reason. Like that it complains when passing a 3-component xr vector, but I guess I can cast through `float*` to get it to shut up?

src/modules/headset/headset_openxr.c
s-ol: yeah, I was unsure about this but did to me the spec reads like initialization is up to the runtime here?

> If propertyCapacityInput is not 0, properties must be a pointer to an array of propertyCapacityInput [XrExtensionProperties](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XrExtensionProperties) structures

vs

> nodeProperties is a pointer to an application-allocated array that will be filled with the [XrControllerModelNodePropertiesMSFT](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#XrControllerModelNodePropertiesMSFT) values. It can be NULL if nodeCapacityInput is 0.

s-ol:
> Regarding node lookups: since the controller keys are cached and per-device, we would be able to globally cache the node lookups for each device (something like 16 `uint32_t` node indices or something).

Maybe it's overly complex, but I was thinking something like an optional pointer with an enum for the model type (handtracking or msft) and associated data (tracker/modelKey, cached node indices). When `animate=false` in newModel, it would be null.
The animate function would then only accept the model argument.

src/modules/headset/headset_openxr.c
s-ol: I just carefully read the section on [buffer size parameters](https://registry.khronos.org/OpenXR/specs/1.0/html/xrspec.html#buffer-size-parameters) and you're right actually:

> In case that elements is a struct with type and next fields, the application must set the type to the correct value as well as next either to NULL or a struct with extension related data in which type and next also need to be well defined.

src/modules/graphics/graphics.c
s-ol: oh nevermind, not doing direct copies anymore anyway so this change was completely unnecessasy.

src/modules/headset/headset_openxr.c
s-ol: yes, that's true. I think I got a bit confused when I first looked at the way the transforms are stored and thought these were absolute, so this was supposed to apply them locally bet that was all wrong.

This is where the float[3]/float[4] thing comes in, passing the position straight in gives this warning:

```
/home/s-ol/Documents/other/lovr/src/modules/headset/headset_openxr.c: In function ‘openxr_animateMSFT’:
/home/s-ol/Documents/other/lovr/src/modules/headset/headset_openxr.c:1867:5: warning: ‘lovrModelSetNodeTransform’ accessing 16 bytes in a region of size 4 [-Wstringop-overflow=]
 1867 |     lovrModelSetNodeTransform(model, nodeIndex, &nodeStates[i].nodePose.position.x, NULL, rotation, 1);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/s-ol/Documents/other/lovr/src/modules/headset/headset_openxr.c:1867:5: note: referencing argument 3 of type ‘float[4]’
/home/s-ol/Documents/other/lovr/src/modules/headset/headset_openxr.c:1867:5: note: referencing argument 4 of type ‘float[4]’
/home/s-ol/Documents/other/lovr/src/modules/headset/headset_openxr.c:1867:5: note: referencing argument 5 of type ‘float[4]’
```

It's not pretty, but making a copy with `vec3_init` works.

bjornbytes:
Originally `.animate` was kept permissive in what it accepts because a few people were authoring custom controller models and wanted to easily animate them based on controller inputs.  I don't really think the idea panned out though, because it ended up being too tricky to set up models exactly like the built-in ones.  So it's probably fine to make `.animate` more strict.

There is a `metadata` field in ModelData right now, which could be a good place to store the pointer/enum you described.  But the glTF loader currently stores the glTF JSON in there, and I'm not sure how to easily circumvent that.  The "name" field of the Blob could be another place to store a bit of arbitrary data?  It's maybe a bit weird.

Anyway, aside from a few small style nits this looks good to me!  Let me know when you're comfortable merging.

s-ol:
> There is a `metadata` field in ModelData right now, which could be a good place to store the pointer/enum you described. But the glTF loader currently stores the glTF JSON in there, and I'm not sure how to easily circumvent that. The "name" field of the Blob could be another place to store a bit of arbitrary data? It's maybe a bit weird.

I saw that field but didn't realize that it was unused aside from lua api access. I could throw out the json data after loading and add a `metadata_type` enum so that the lua api knows not to pass out random data perhaps?

> Anyway, aside from a few small style nits this looks good to me! Let me know when you're comfortable merging.

I'd give this a shot still I think, but feel free to point out any of these details and I'll get them cleaned up :)

s-ol:
> Originally `.animate` was kept permissive in what it accepts because a few people were authoring custom controller models and wanted to easily animate them based on controller inputs. I don't really think the idea panned out though, because it ended up being too tricky to set up models exactly like the built-in ones. So it's probably fine to make `.animate` more strict.

I see, the opposite could also make sense at times: animating a stock model with non-user inputs (e.g. for tutorialization). The controller and hand models are both black boxes from the application perspective, so I'd simplify, but maybe the cross-vendor API will improve on this later.

bjornbytes:
I'm doing some work on dev so I'll merge this to avoid any possible bitrot.  Thanks again for the PR, feel free to send any additional changes if needed!  In a few weeks I should be able to test on Windows + WMR as well.

s-ol:
> I'm doing some work on dev so I'll merge this to avoid any possible bitrot. Thanks again for the PR, feel free to send any additional changes if needed! In a few weeks I should be able to test on Windows + WMR as well.

I think this is actually good as it is, I'd just recommend that you take a look at where I put the struct and enum in the last change and think a moment about the names I gave things there, I jotted those down pretty quickly to get things working but don't think they're very clear as it stands.

PR #605, opened 2022-11-22
Pico android flavor

brainrom:
Tested on Pico Neo 3 with PUI v4.9.3
About "Imposter": I want to explicitly show, that this binding doesn't work and leads to engine crash during startup, even it's listed in the Pico's example.
Also included generic (Monado-Android) Android flavor.


src/modules/headset/headset_openxr.c
```diff
@@ -330,6 +330,11 @@ static bool openxr_init(HeadsetConfig* config) {
 #ifdef LOVR_VK
       { "XR_KHR_vulkan_enable2", NULL, true },
 #endif
+
+#ifdef __ANDROID__
+      { "XR_KHR_android_create_instance", NULL, true },
+      { "XR_PICO_android_controller_function_ext_enable", NULL, true},
```
bjornbytes, line 336: Because the 'feature' here is `NULL`, the extension will always get added.  Does this mean the instance will fail to create on other OpenXR runtimes that don't support the extension?

src/modules/headset/headset_openxr.c
```diff
@@ -778,15 +835,18 @@ static bool openxr_init(HeadsetConfig* config) {
...
-        XR_INIT(xrSuggestInteractionProfileBindings(state.instance, &(XrInteractionProfileSuggestedBinding) {
+        int res = (xrSuggestInteractionProfileBindings(state.instance, &(XrInteractionProfileSuggestedBinding) {
           .type = XR_TYPE_INTERACTION_PROFILE_SUGGESTED_BINDING,
           .interactionProfile = path,
           .countSuggestedBindings = count,
           .suggestedBindings = suggestedBindings
         }));
+        if (XR_SUCCEEDED(res)) successProfiles++;
       }
+      XR_INIT(successProfiles);
```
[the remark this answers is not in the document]
brainrom: Monado breaks on `/interaction_profiles/pico/neo3_controller` interaction profile, so I wanted to test, if at least one profile is loaded. But `int successProfiles = 0;` should be before profiles iteration.

src/modules/headset/headset_openxr.c
brainrom: `XR_PICO_android_controller_function_ext_enable` needs to be removed, because it's unused. `XR_KHR_android_create_instance` are added for android, and Android OpenXR environments (Pico, Oculus, Monado-Android) supports it.

src/modules/headset/headset_openxr.c
```diff
@@ -362,7 +376,10 @@ static bool openxr_init(HeadsetConfig* config) {
       .applicationInfo.applicationVersion = 0,
       .applicationInfo.apiVersion = XR_CURRENT_API_VERSION,
       .enabledExtensionCount = enabledExtensionCount,
-      .enabledExtensionNames = enabledExtensionNames
+      .enabledExtensionNames = enabledExtensionNames,
+#ifdef __ANDROID__
+      .next = &infoAndroid
```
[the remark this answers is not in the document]
brainrom: OK

etc/AndroidManifest_generic.xml
```diff
@@ -0,0 +1,36 @@
...
+
+	  <meta-data android:name="android.app.lib_name" android:value="lovr" />
+		  <intent-filter>
+			<action android:name="android.intent.action.MAIN"/>
+			<category android:name="org.khronos.openxr.intent.category.IMMERSIVE_HMD" />
+			<category android:name="android.intent.category.LAUNCHER"/>
+		  </intent-filter>
+	  </activity>
+  </application>
+</manifest>
```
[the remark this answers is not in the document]
brainrom: OK

CMakeLists.txt
```diff
@@ -201,8 +202,18 @@ if(LOVR_ENABLE_HEADSET AND LOVR_USE_OPENXR)
...
+      set(LOVR_OPENXR_OCULUS_PATH "${CMAKE_CURRENT_SOURCE_DIR}/deps/oculus-openxr" CACHE STRING "The path to the Oculus OpenXR loader")
+      set_target_properties(openxr_loader PROPERTIES IMPORTED_LOCATION "${LOVR_OPENXR_OCULUS_PATH}/Libs/Android/${ANDROID_ABI}/Release/libopenxr_loader.so")
+    elseif(ANDROID_FLAVOR STREQUAL "pico")
+      set(LOVR_OPENXR_PICO_PATH "${CMAKE_CURRENT_SOURCE_DIR}/deps/pico-openxr" CACHE STRING "The path to the Pico OpenXR loader")
+      set_target_properties(openxr_loader PROPERTIES IMPORTED_LOCATION "${LOVR_OPENXR_PICO_PATH}/Libs/Android/${ANDROID_ABI}/libopenxr_loader.so")
+    elseif(ANDROID_FLAVOR STREQUAL "generic")
+      set(LOVR_OPENXR_GENERIC_PATH "${CMAKE_CURRENT_SOURCE_DIR}/deps/generic-openxr" CACHE STRING "The path to the Generic OpenXR loader")
+      set_target_properties(openxr_loader PROPERTIES IMPORTED_LOCATION "${LOVR_OPENXR_GENERIC_PATH}/Libs/Android/${ANDROID_ABI}/libopenxr_loader.so")
+    else()
+      message(FATAL_ERROR "Unsupported Android flavor (expected: oculus, pico or generis)")
```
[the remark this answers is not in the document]
brainrom: OK

bjornbytes:
Sorry to leave this hanging.  Pico SDK v2.1.0 was released 12/06/22 and says

> - Support OpenXR standard API 1.0.24
> - Migrate the original non-standard extensions of PICO to standard extensions
> - To use the new version of the OpenXR SDK, ROM version should be 5.2.x and later (For Neo3, a supported version is expected to be released after 2022/12/15).
> - Some PICO non-standard extensions have been removed. See the migration instructions in the development documentation for details.

Were you running on this new version, or the previous one?  I'm curious if this means Pico will work with the generic OpenXR loader now.  That would be awesome.

brainrom:
I have only Neo 3, firmware which supports new SDK still doesn't released, so I use previous version.
Anyway, I can't found any clues, that Pico now works with the generic loader, SDK docs still says about custom Pico loader. There is still a long way to come to universal VR apk. Maybe for now it's better to write universal manifest, place all loaders inside APK and determine the platform on Java side somehow?

bjornbytes:
Stable libc++ file path for all NDK versions (including NDK 25+) seems to be

```
./toolchains/llvm/prebuilt/linux-x86_64/sysroot/usr/lib/aarch64-linux-android/libc++_shared.so
```

This should fix the build.  The path might need to account for the host tag (`ANDROID_HOST_TAG`).  Let me know if you're up for it, otherwise I can merge and make the change.

I am also adding a TODO to switch other libraries (ode, msdfgen) to use shared C++ library in CMake and tup, which is the more correct behavior (Android requires this if your project uses multiple shared libraries due to global conflicts in the C++ standard library).  It might reduce file size a bit.

brainrom:
I think, use `ANDROID_HOST_TAG` is a good idea (even if almost all Android HMDs are arm64). Before merge, I think we should also make a Git mirror of Pico OpenXR loader (on lovr-org  preferably) and add it as dependency, and I should test latest Pico XR SDK on recently released Neo3 firmware.

brainrom:
I tested the 2.1.0 SDK with the new firmware, it works flawless. Also I added the ifdef, around `XR_MND_headless`, because it still breaks Pico (and useless on Android at all). Please make a GitHub mirror of [Pico OpenXR SDK](https://developer-global.pico-interactive.com/sdk?deviceId=1&platformId=3&itemId=11), add it as dependency, and merge to master.

src/modules/headset/headset_openxr.c
```diff
@@ -757,6 +777,44 @@ static bool openxr_init(HeadsetConfig* config) {
...
+        { ACTION_B_DOWN, "/user/hand/right/input/b/click" },
+        { ACTION_B_TOUCH, "/user/hand/right/input/b/touch" },
+        { ACTION_X_DOWN, "/user/hand/left/input/x/click" },
+        { ACTION_X_TOUCH, "/user/hand/left/input/x/touch" },
+        { ACTION_Y_DOWN, "/user/hand/left/input/y/click" },
+        { ACTION_Y_TOUCH, "/user/hand/left/input/y/touch" },
+        { ACTION_THUMBREST_TOUCH, "/user/hand/left/input/thumbrest/touch" },
+        { ACTION_THUMBREST_TOUCH, "/user/hand/right/input/thumbrest/touch" },
+        { ACTION_VIBRATE, "/user/hand/left/output/haptic" },
+        { ACTION_VIBRATE, "/user/hand/right/output/haptic" },
```
bjornbytes, line 817: Don't forget `{ 0, NULL }` here, this is currently crashing on Windows.

src/modules/headset/headset_openxr.c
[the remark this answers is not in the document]
bjornbytes: Ok, I understand.  Currently we abort initialization if binding suggestion fails for *any* of the interaction profiles, but due to runtime instability the proposal is to change it to abort if *all* of them fail.

What if we made binding suggestion failure a warning instead of an abort?

```c
XrResult result = xrSuggestInteractionProfileBindings(...);

if (XR_FAILED(result)) {
  lovrLog(LOG_WARN, "XR", "Failed to suggest input bindings for %s", interactionProfilePaths[i]);
}
```

Although suggesting the bindings is usually supposed to succeed, it seems okay to proceed without input if they fail, as long as we log it.

etc/AndroidManifest.xml
```diff
@@ -14,16 +14,26 @@
...
+
+  <!-- Khronos/Generic -->
+  <uses-permission android:name="org.khronos.openxr.permission.OPENXR" />
+  <uses-permission android:name="org.khronos.openxr.permission.OPENXR_SYSTEM" />
+  <queries>
+    <provider android:name="org.khronos.openxr.runtime_broker" android:authorities="org.khronos.openxr.runtime_broker;org.khronos.openxr.system_runtime_broker" />
+  </queries>
+
   <application android:allowBackup="false" android:label="LÖVR" android:extractNativeLibs="false" android:debuggable="true">
+	<meta-data android:name="pvr.app.type" android:value="vr"/> <!-- Pico -->
```
bjornbytes, line 26: This should use spaces instead of tabs, for consistency.

bjornbytes:
Tested on Quest with NDK 25.1.8937393, works great.

I added the Pico OpenXR SDK submodule, let me know if you have any issues with it.

Left a few more comments as well.

src/modules/headset/headset_openxr.c
brainrom: Fixed

etc/AndroidManifest.xml
brainrom: Fixed

bjornbytes:
Thanks!  🎉

brainrom:
FINALLY!!!!

PR #616, opened 2022-12-13
Debian

fendrin:
Files for building debian packages.


bjornbytes:
Thank you for adding this!  I'm a little bit apprehensive about merging this pull request for a few reasons:

- I don't know anything about creating and maintaining distro packages!  (this is solvable)
- There seems to be a lot of complexity here, both in terms of files and maintenance burden (changelog, scripts, licenses, worrying about external packages)
- Introducing debian builds is strong precedent for adding and maintaining builds for other distros

For these reasons, I don't know that I'm comfortable maintaining this right now.

PR #619, opened 2022-12-22
Check if torus shape is cached

jmiskovic:
Without it drawing the second torus will crash the app


bjornbytes:
Double checked that this was the only one missing.  Thanks.

PR #627, opened 2022-12-26
A getter for World list of tags

jmiskovic:
I was missing this when serializing/re-creating the physics scene.
Amazingly enough just about everything else about the world state and geometry was already retrievable by the API 🙌


jmiskovic:
BTW world state snapshots were added to [phywire](https://github.com/jmiskovic/lovr-phywire) lib.

PR #629, opened 2023-01-06
Fix constrained rendering in windowed mode on macOS #620

### LINKED ISSUE #620: Desktop rendering on macOS in windowed mode is constrained to 25% of the view.
In macOS, when rendering in a window (not fullscreen, no headset connected) the view is limited to 25% of the total window area.

This issue was verified in version v0.16.0 using a locally built (but unmodified) version of LOVR. 

See discussion & example image here: https://matrix.to/#/!XVAslexgYDYQnYnZBP:matrix.org/$1671371144169315aOiDS:matrix.org

undefdev:
Fixes #620

Desktop rendering on macOS in windowed mode was previously constrained to 25% of the view. This was caused by using the initial window size rather than the framebuffer size when setting the viewport.

This issue has been fixed by using `os_window_get_fbsize` instead of `os_window_get_size` when setting the viewport. This ensures that the viewport is correctly set based on the framebuffer size, which takes DPI into account.

Tested on macOS 13.1 ARM64.

### DIFF
--- a/src/modules/graphics/graphics.c
+++ b/src/modules/graphics/graphics.c
@@ -649,7 +649,7 @@ bool lovrGraphicsInit(GraphicsConfig* config) {
       .srgb = true
     };
 
-    os_window_get_size(&state.window->info.width, &state.window->info.height);
+    os_window_get_fbsize(&state.window->info.width, &state.window->info.height);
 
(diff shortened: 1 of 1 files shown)


bjornbytes:
Thanks for the test

PR #644, opened 2023-03-01
Fix signs for light probe coefficients

jmiskovic:
Light probes are neat!

While testing I noticed the light often came from wrong direction, and traced it down to few wrong signs when converting cubemap UV coordinates to 3D direction vector, and signs for SH coefficients.

I used https://www.ppsloan.org/publications/StupidSH36.pdf as a reference. Testing was done with [this script](https://gist.github.com/jmiskovic/cf2713b6b476f2c49f10e701ae3fe545).

I've also compared results with [three.js implementation](https://threejs.org/examples/?q=probe#webgl_lightprobe_cubecamera), it comes out completely identical to my eye.

image

### DIFF
--- a/src/api/l_math.c
+++ b/src/api/l_math.c
@@ -132,15 +132,30 @@ static int l_lovrMathNewCurve(lua_State* L) {
 
 static void getPixelEquirect(void* image, uint32_t x, uint32_t y, uint32_t z, float color[4]) {
   lovrImageGetPixel(image, x, y, z, color);
+  if (lovrImageIsSRGB(image)) {
+    color[0] = lovrMathGammaToLinear(color[0]);
+    color[1] = lovrMathGammaToLinear(color[1]);
+    color[2] = lovrMathGammaToLinear(color[2]);
+  }
 }
 
 static void getPixelCubemap(void* image, uint32_t x, uint32_t y, uint32_t z, float color[4]) {
   lovrImageGetPixel(image, x, y, z, color);
+  if (lovrImageIsSRGB(image)) {
+    color[0] = lovrMathGammaToLinear(color[0]);
+    color[1] = lovrMathGammaToLinear(color[1]);
+    color[2] = lovrMathGammaToLinear(color[2]);
+  }
 }
 
 static void getPixelCubemapLayers(void* context, uint32_t x, uint32_t y, uint32_t z, float color[4]) {
   Image** images = context;
   lovrImageGetPixel(images[z], x, y, 0, color);
+  if (lovrImageIsSRGB(images[z])) {
+    color[0] = lovrMathGammaToLinear(color[0]);
+    color[1] = lovrMathGammaToLinear(color[1]);
+    color[2] = lovrMathGammaToLinear(color[2]);
+  }
 }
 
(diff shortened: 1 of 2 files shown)


jmiskovic:
Quick rundown of changes:

* The SH basis coefficients alternate between + and - sign, as per appendix A2 of StupidSH36
* When iterating through texels, offset by 0.5 to get to center of a texel
* Converting UV to 3D direction - first image is -X, second is +X... so case 0 should have -1.f as X coord
* V image coord is 0 for top and has to be inverted when converting to 3D space
* introduced GammaToLinear for better contrast, and to match existing examples

shakesoda:
The color being linearized is suspect for anything other than sRGB input
colors - does this only run on sRGB images?

bjornbytes:
No, it's possible to create a light probe from e.g. a cubemap in a KTX file with a floating point format.

I think the `getPixel` callback should look at the `lovrImageIsSRGB` flag to see if the conversion is needed.

jmiskovic:
Also changed equirect. Not sure if it is correct because it's not as easy to generate in runtime as it is with skybox, but it looks better.

I couldn't find any KTX skybox file that lovr would cooperate with. Some were "supercompressed" and others reported "unsupported image format (glType = 0, glFormat = 0, glInternalFormat = 37492)". If you have any skybox in that format I can add the conversion check and test it.

bjornbytes:
If you've got filament's `cmgen` tool, it can convert cubemap/equirect images to KTX cubemaps in `rg11b10f` format.

bjornbytes:
I just learned that the convention where the sign alternates like this is called [Condon-Shortley phase factor](https://en.wikipedia.org/wiki/Spherical_harmonics#Condon%E2%80%93Shortley_phase).  Some scientific fields use it, while others leave it out.  I think as long as you are consistent with the negation during both A) basis construction B) evaluation, everything should turn out the same.  The Sloan paper lists them with this phase factor, whereas the [Stanford paper](https://graphics.stanford.edu/papers/envmap) (which I used) doesn't use it, along with THREE/filament/[Wikipedia](https://en.wikipedia.org/wiki/Table_of_spherical_harmonics).  Even though both are valid, I lean towards going back to all-positive coefficients since it seems to be a little more common in graphics.

Regarding cubemaps, I definitely got these wrong and forgot to flip on X/Y -- I've mostly been testing equirect so far.  (Note that the face order is still +X/-X, but there is an X flip needed during sampling for...reasons I don't remember.  But it's in cubemap.frag).

jmiskovic:
Reverted the Condon-Shortley alternating signs. Testing script in above gist is also updated.

Kept the +X/-X change to accommodate the cubemap flip, the results are obviously wrong without it. Reverted some code alignment whitespace.

The linearization now checks for sRGB image. It is done in another commit so it can be rejected or done in a different way. (I hated both options of adding the check for each and every pixel, and branching to 3 more functions)

bjornbytes:
Thanks for the fixes, looking good now.  Yeah, I'm worried about the performance of all the per-pixel checks, but didn't want to pass an Image/pointer into the math module.  I want to do some tests to see how bad it really is though.

PR #645, opened 2023-03-01
Address various MSVC warnings (casts, comparisons, function pointers)

mcclure:
Address size_t narrowing warnings in MSVC by adding casts. Also adds some asserts to make sure the casts are safe.

I have not tested this patch beyond compiling it because I'm not sure how.

### DIFF
--- a/src/api/l_graphics_pass.c
+++ b/src/api/l_graphics_pass.c
@@ -891,7 +891,12 @@ static int l_lovrPassCopy(lua_State* L) {
     uint32_t srcOffset = luax_optu32(L, 4, 0);
     uint32_t dstOffset = luax_optu32(L, 5, 0);
     const BufferInfo* info = lovrBufferGetInfo(buffer);
-    uint32_t limit = MIN(blob->size - srcOffset, info->length * info->stride - dstOffset);
+    uint32_t bufferSize = info->length * info->stride;
+    lovrCheck(bufferSize==0 || bufferSize/info->stride == info->length, "Buffer size exceeds supported limits. This is a bug in Lovr.");    
+    lovrCheck(srcOffset <= blob->size, "Source index is %d but blob is only %d bytes", srcOffset, (uint32_t)blob->size);
+    lovrCheck(dstOffset <= bufferSize, "Destination index is %d but buffer is only %d bytes", dstOffset, bufferSize);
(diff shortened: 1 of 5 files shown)


mcclure:
These changes fix all warnings on a dev-0.16 compile, except the one I mentioned in Matrix

> graphics.c? Line 4530:
>   `for (uint32_t side = 0; side <= thicc; side++, z *= -1.f, nz *= -1.f, angle = 0.f) {`
> The side <= thicc is making MSVC angry because it thinks this is a signed/unsigned mismatch, but this isn't even the real problem, the real problem is that thicc is a boolean, so the for loop makes no sense to start with. Is this a typo for the variable "c", set several lines above based on thicc?

mcclure:
Bjorn explained the mystery line so I fixed that plus a line in the Phonon code.

src/api/l_graphics_pass.c
```diff
@@ -891,7 +891,11 @@ static int l_lovrPassCopy(lua_State* L) {
     uint32_t srcOffset = luax_optu32(L, 4, 0);
     uint32_t dstOffset = luax_optu32(L, 5, 0);
     const BufferInfo* info = lovrBufferGetInfo(buffer);
-    uint32_t limit = MIN(blob->size - srcOffset, info->length * info->stride - dstOffset);
+    uint32_t bufferSize = info->length * info->stride;
+    lovrCheck(srcOffset <= blob->size, "Source index is %d but blob is only %d bytes", srcOffset, (uint32_t) blob->size);
+    lovrCheck(dstOffset <= bufferSize, "Destination index is %d but buffer is only %d bytes", dstOffset, bufferSize);
```
bjornbytes, line 896:
```suggestion
    lovrCheck(srcOffset <= blob->size, "Source byte offset is %d but Blob is only %d bytes", srcOffset, (uint32_t) blob->size);
    lovrCheck(dstOffset <= bufferSize, "Destination byte offset is %d but Buffer is only %d bytes", dstOffset, bufferSize);
```

Small suggestion to differentiate between index/offset terminology (index is 1-indexed and means "multiplied by byte stride" in context of buffers).

bjornbytes: Oh, I can commit this

PR #659, opened 2023-03-23
Fix acos domain error in angle between two Vec3

jmiskovic:
A reproduction:

`print(vec3(3.0253329277039, -2.000001, -0.79650980234146):angle(vec3(3.0253329277039, -2, -0.79650980234146)))`


bjornbytes:
I guess it is just floating point imprecision?  Thanks for the fix

jmiskovic:
Yes, numerical errors when computing near singularities.

PR #674, opened 2023-06-17
Add mousewheel callback

jmiskovic:
Added `lovr.wheelmoved(x, y)` callback. Scrolling right gives positive  x, scrolling up gives positive y values. This is same callback name and sign convention as LOVE2D.

GLFW implementation is tested on linux & windows. Emscripten implementation is best effort; it compiles and tries to set correct signs based on docs. xcb is not implemented.

### DIFF
--- a/src/core/os_android.c
+++ b/src/core/os_android.c
@@ -334,6 +334,10 @@ void os_on_mouse_move(fn_mouse_move* callback) {
   //
 }
 
+void os_on_mousewheel_move(fn_mousewheel_move* callback) {
+  //
+}
+
 void os_on_permission(fn_permission* callback) {
   state.onPermissionEvent = callback;
 }
--- a/src/api/l_event.c
+++ b/src/api/l_event.c
@@ -17,6 +17,7 @@ StringEntry lovrEventType[] = {
   [EVENT_MOUSEPRESSED] = ENTRY("mousepressed"),
   [EVENT_MOUSERELEASED] = ENTRY("mousereleased"),
   [EVENT_MOUSEMOVED] = ENTRY("mousemoved"),
+  [EVENT_MOUSEWHEELMOVED] = ENTRY("wheelmoved"),
 #ifndef LOVR_DISABLE_THREAD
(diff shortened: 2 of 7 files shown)


jmiskovic:
The MacOS build fails with unrelated error. The commit doesn't touch MacOS and I don't have access to try if the wheel callback works.

```
/Users/runner/work/lovr/lovr/plugins/lua-https/src/apple/NSURLClient.mm:8:2: error: "ARC is off"
#error "ARC is off"
```

bjornbytes:
Yup, don't worry about macOS build, it's broken right now.  lua-https only works with the XCode generator, not the Makefile generator.  The `-fobj-arc` flag needs to be applied manually, but this requires me to fork and switch the submodule URL, which I've been procrastinating on.

src/core/os_android.c
```diff
@@ -334,6 +334,10 @@ void os_on_mouse_move(fn_mouse_move* callback) {
   //
 }
 
+void os_on_mousewheel_move(fn_mousewheel_move* callback) {
+  //
```
bjornbytes, line 338: Note for future: It would be possible to add checks for `AMOTION_EVENT_ACTION_SCROLL` and `AMOTION_EVENT_AXIS_HSCROLL`/`AMOTION_EVENT_AXIS_VSCROLL` to support this on Android.  I don't have any bluetooth mice to test w/ though.

bjornbytes:
Thank you!

src/core/os_android.c
jmiskovic: I'm interested in extending the supporting to Android.

BTW what's your opinion in funneling the phone touch events to mouse events like Love2D does?

bjornbytes: I think it would be totally fine to add touch emulation for the mouse inputs, just as a secret/unofficial "nice to have" feature.

Mouse can still be useful for VR (virtual desktop scenario), but I can't think of a VR use case for touch.  Even if you were using a phone as an input device for a VR experience, that would probably better be left as a separate phone app that communicates with the VR app.  So LÖVR probably isn't ready to add a full touch API.

PR #677, opened 2023-07-01
Implement mousearm simulator controls;

bjornbytes:
Updates simulator controls:

- Use [mousearm](https://github.com/jmiskovic/lovr-mousearm)-style hand pose, where the controller is positioned based on the mouse
- Improved smoothing behavior, fixes slight stuttering
- Adds sprint (hold shift to move 5x faster)

~~Still need to add back velocity and make scroll change hand distance.~~

Not yet sure how to handle the discontinuity in the cursor position when releasing the mouse button.  It causes a jump in the hand position which is a bit jarring.

- Current approach is to cause the controller to lose tracking while mouse button is held.  It's slightly better, but might make testing more difficult and is still a little distracting for experiences that draw hands.
- I also tried some kind of solution where the controller stays in a fixed spot in the world while the mouse button is held down, and then interpolates towards the new position when the mouse is released.  It felt weird to have the hand zooming all over the place as you're looking around the scene.
- One alternative is to just use a normal mouse cursor for the mouselook instead of using GLFW's hidden cursor.  I think this is too annoying, because you tend to click outside the window after moving the mouse too far.

### DIFF
--- a/src/core/maf.h
+++ b/src/core/maf.h
@@ -391,6 +391,12 @@ MAF mat4 mat4_fromQuat(mat4 m, quat q) {
   return m;
 }
 
+MAF mat4 mat4_fromPose(mat4 m, vec3 v, quat q) {
+  mat4_fromQuat(m, q);
+  memcpy(m + 12, v, 3 * sizeof(float));
+  return m;
+}
+
 MAF mat4 mat4_identity(mat4 m) {
   m[0] = 1.f;
   m[1] = 0.f;
--- a/src/core/os_glfw.h
+++ b/src/core/os_glfw.h
@@ -283,6 +283,8 @@ static int convertKey(os_key key) {
     case KEY_DOWN: return GLFW_KEY_DOWN;
     case KEY_LEFT: return GLFW_KEY_LEFT;
     case KEY_RIGHT: return GLFW_KEY_RIGHT;
+    case KEY_LEFT_SHIFT: return GLFW_KEY_LEFT_SHIFT;
+    case KEY_RIGHT_SHIFT: return GLFW_KEY_RIGHT_SHIFT;
     case KEY_ESCAPE: return GLFW_KEY_ESCAPE;
     case KEY_F5: return GLFW_KEY_F5;
     default: return GLFW_KEY_UNKNOWN;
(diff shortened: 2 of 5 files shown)


jmiskovic:
Excited to see this built-in!

I've just noticed some erratic hand orientations when I turn backward:

mousearm_orientation

This is also present in my implementation :(

```lua
function lovr.draw(pass)
  for i, hand in ipairs(lovr.headset.getHands()) do
    local pose = mat4(lovr.headset.getPose(hand))
    pass:box(pose:scale(0.1, 0.01, 0.01))
  end
  for a = 0, math.pi * 2, math.pi / 8 do
    local pos = vec3(quat(a, 0, 1 ,0)):mul(10)
    pass:text(math.floor(math.deg(a)) .. '°', mat4():target(pos, vec3()):scale(-1, 1, 1))
  end
end
```

jmiskovic:
The cause is computing the orientation quat from direction vector, where vec3(0, 0, 1) is opposite from forward direction so the up-vector changes rapidly with small movements. 

The best way I can think off is two-step rotation, first along the horizon and then vertical rotation. Then the erratic rotation happens only when looking all the way up or down.

```c
    float forward[4] = { 0.f, 0.f, -1.f, 0.f };
    float horizon_dir[4] = { hand[0], 0.f, hand[2], 0.f };
    float vertical_quat[4];
    vec3_normalize(horizon_dir);
    quat_between(vertical_quat, horizon_dir, hand);
    quat_between(state.handOrientation, forward, horizon_dir);
    quat_mul(state.handOrientation, state.handOrientation, vertical_quat);
```

bjornbytes:
You're right, we lost the up vector, I think I finally found a good solution by computing a full rotation basis:

- Use the forward vector from the mouse.
- Cross product forward vector with head's up vector to get the "right" vector.
- Cross product the "right" vector and the forward vector to get an orthogonal up vector.
- These form a rotation matrix, which can be converted to a quaternion

In the code I used `mat4_target` because it was a bit more clean.

jmiskovic:
BTW I actually like that it loses tracking while rotating camera. It forces users to deal with tracking issues while still on desktop where it's easier to trigger and fix them.

Issue #86, opened 2018-12-12
Frustum Culling Tools


bjornbytes:
LÖVR should give you the tools you need to do frustum culling (without actually trying to do it for you).

This is an interesting problem for VR because there are two cameras and each one has a frustum.  Some magic math will need to be used to combine the two frusta into a single one.  That combined frustum can be exposed by lovr.headset.  I'm not sure what format is best to represent a frustum: mat4?  6 planes?  How are planes represented?

From there, there's probably something in lovr.math like `mat4:contains` or `lovr.math.testFrustum` that is able to test things against frusta.  You could use your own or the one returned from lovr.headset.

When doing the actual tests, need to think about:

- What sorts of shapes can you test?  Points, spheres, AABBs, OBBs?
- Does it return In/Out or In/Out/Intersects?  The latter can be used to optimize certain spatial hashes.

shakesoda:
i've got more around somewhere, but here's the aabb-frustum test i use in my games https://github.com/excessive/ludum-dare-43/blob/master/src/math/Intersect.hx#L41

bjornbytes:
`lovr.graphics.frustum` to render a frustum?  Mostly for debugging probably.

bjornbytes:
`lovr.graphics` can just frustum cull primitives internally because it knows the camera and knows the AABBs of its primitives (can also use AABBs of Models/Meshes when they are provided).  It needs to be configurable in case you have a weird vertex shader.  Not sure if culling each primitive individually as it's about to be drawn isn't too slow/awful compared to something more fancy.

But also the new physics stuff will also have some kind of Frustum object in case you want to do custom tests/culling/clustering.  Headset still ought to expose getFrustum that returns data for one of these objects.

PR #678, opened 2023-07-05
Frustum culling;

bjornbytes:
- Adds Pass:setViewCull to enable/disable frustum culling.
- Renames Pass:setCullMode to Pass:setFaceCull (with backcompat).
- Adds `drawsCulled` pass stat.

Some stuff currently missing:

- Text is not culled, but should be.
- VR view frusta are not merged yet.

Optimizations I tried that weren't worth it:

- Detecting infiniteZ and reducing plane count to 5 made it slower (branching?)
- Putting transforms and bounds in a separate array from the draws (to avoid ever touching memory for culled draws, and reduce size of Draw struct) did have a measurable impact of 5% in the "hot" loop that actually does draws.  But draw calls are not actually a bottleneck of render pass recording right now, and so this change doesn't seem worth the extra code complexity.
- Pre-transforming bounding boxes by the draw's transform (in `lovrPassDraw`) makes culling twice as fast, but makes overall overhead slightly higher when draws are only submitted once.  This would probably still be a worthwhile optimization, because you get faster pass submission in the case where you're saving a pass and submitting it many times (~20% assuming lots of its draws are culled).

### DIFF
--- a/etc/monkey.h
+++ b/etc/monkey.h
@@ -1,4 +1,4 @@
-float monkey_size[3] = { 1.367188f, 0.984375f, 0.851563f };
+float monkey_bounds[6] = { 0.000000f, 0.000000f, 0.000000f, 0.683594f, 0.492188f, 0.425781f };
 float monkey_offset[3] = { -0.683594f, -0.492188f, -0.425781f };
 
 uint8_t monkey_vertices[] = {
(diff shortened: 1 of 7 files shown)


bjornbytes:
Closes #86

PR #679, opened 2023-07-06
LightProbe

bjornbytes:
Adds a new `LightProbe` object to `lovr.math` that can be used to easily do soft diffuse lighting.

Here are some thoughts about why this design might not be optimal:

- It's CPU-only.
  - Generating light probes from images on the CPU is slower than doing it on the GPU.
  - A reasonable use case is generating dynamic light probes by rendering the scene to a cubemap and creating a LightProbe from that.  With this design, dynamic light probes would require the skybox texture to be read back to the CPU, which either adds latency or a blocking wait.
- Each instance is just a single probe.
  - Fancy 3D engines will allow you to pre-compute hundreds of light probes at various locations in an environment, storing them in an optimized format.  At runtime, to light an object with the probes, the engine will quickly compute the closest 4 or so probes to the object and use a blend of them for the indirect diffuse lighting.  Should LÖVR try to offer something like this, or is it best left up to the application?  If it's up to the app, is it even worth it for LÖVR to expose the basic math for probes?  e.g. apps might instead just store big lists of spherical harmonic coefficients in tables and put them directly in `Buffer`s rather than create a bunch of tiny little `LightProbe` instances that don't provide much value.

Things still missing on this PR:

- Need to add shader helper `evaluateLightProbe`
- I think some of the basis evaluation can be simplified with a helper function.


bjornbytes:
There should be like `Pass:setLightProbe`, and shaders should:

- Incorporate light probe lighting by default
- Have a shader flag to enable/disable it
- Expose a helper function to evaluate the light probe

Plus ways to compute a light probe using the GPU from a Texture e.g. Texture:computeLightProbe, Texture:newLightProbe, or lovr.graphics.newLightProbe.  These would be able to write the result to a Buffer or read it back and return it as a table/LightProbe.

bjornbytes:
I think `Pass:setLightProbe` is too opinionated for LÖVR.  It's hard for LÖVR to make a good decision about how to incorporate the diffuse lighting from spherical harmonics, and for all but the simplest cases it will likely end up disabled and add overhead.

However, I do think that having some way to compute spherical harmonics using the GPU is really valuable.  I landed on `Pass:computeSphericalHarmonics` for the API.  `Texture:computeSphericalHarmonics` could work, but it has some drawbacks:

- It's not thread safe, unlike a Pass method.
- The synchronization for it is suboptimal compared to a Pass method.  If you're trying to compute 100 light probes and put them all in 1 Buffer, LÖVR's current hazard tracking would do a barrier between each one and serialize the compute dispatches.  If you put them all in a Pass instead, they will all run at the same time and you can choose to `Pass:barrier` if you need any waits.
- You get more fine grained control over ordering when using Passes.  If I want to render the scene to a cubemap, compute spherical harmonics from it, and then render my main pass, I can throw a `Pass:computeSphericalHarmonics(cubemap)` on my main pass and keep everything in a single submit.  If it's on the internal transfer stream, I would need to render the cubemap, submit that pass, then call `Texture:computeSphericalHarmonics` and submit my main pass.

Also, I've decided to switch the name from `LightProbe` to `SphericalHarmonics`.  This name is more accurate because the term "light probe" just means "a thing that records light from every direction".  You could represent a light probe with anything -- spherical harmonics, cubemaps, function, etc.  Spherical harmonics is one possible representation of a light probe, and is the one that we're actually implementing here.  It can also be used for more than light probes -- LÖVR's already using spherical harmonics in the audio module for ambisonics!

bjornbytes:
OK, so I realized this is all just math with Lua tables and compute shaders, so I'm going to turn it into a library instead!  (Hopefully with fancy tetrahedral interpolation and machinery to bake probes).  Closing this!

PR #687, opened 2023-07-20
World query early exit

xiejiangzhi:
```lua
world:queryBox(vec3(), vec3(1), function(shape)
  if job_is_finished then
     return false -- early exit
  end
end)
```

### DIFF
--- a/src/api/l_physics_world.c
+++ b/src/api/l_physics_world.c
@@ -25,7 +25,7 @@ static int nextOverlap(lua_State* L) {
   }
 }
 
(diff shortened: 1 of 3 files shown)


bjornbytes:
Hey, thanks for this!  One change I'd like to make is to use `shouldStop` naming instead of `should_stop`, since variables in the code use camelCase convention.  I don't mind cleaning this up after merge, or you can take care of it, just let me know.

xiejiangzhi:
I also made a little change: set `shouldStop = true` if no callback.

PR #688, opened 2023-07-20
Mesh

bjornbytes:
Adds back Mesh!  Originally it was removed in v0.16.0 because A) I (foolishly) thought `Pass:mesh` with buffers would be sufficient and B) updating meshes would have required a transfer pass, but that's been improved now.

The API is:

```lua
lovr.graphics.newMesh([format], vertices|vertexCount, [indices], [storage=cpu])
Mesh:getVertexCount
Mesh:getVertexStride
Mesh:getVertexFormat
Mesh:getVertexBuffer --> gpu mesh only
Mesh:getIndexBuffer --> gpu mesh only
Mesh:setIndexBuffer --> gpu mesh only
Mesh:getVertices
Mesh:setVertices
Mesh:getIndices
Mesh:setIndices
Mesh:getBoundingBox
Mesh:setBoundingBox
Mesh:computeBoundingBox --> cpu mesh only
Mesh:get/setDrawMode
Mesh:get/setDrawRange
Mesh:get/setMaterial
```

Format is a regular buffer format with some extra restrictions (no nested structs/arrays, no matrix types).

Format can be omitted, in which case you get a default format with positions, normals, and UVs.

Vertices can be provided as a count, table, Blob, or Buffer.

Indices can optionally be provided as a table, Blob, or Buffer (no more `:setVertexMap` immediately after creation!)

The storage for the mesh can be handled in two different ways:

- `cpu` meshes will store a copy of their vertices and indices on the CPU, flushing to GPU buffers before rendering.  Having the data on the CPU takes extra memory, but means:
  - Allows for fast `:getVertices` / `:getIndices` methods
  - If you call `:setVertices` / `:setIndices` multiple times per frame, those will get batched into a single upload
  - Allows for `:computeBoundingBox`
  - Allows you to create `MeshShape`s with the Mesh, or use it with `lovr.audio.setGeometry`.
  - The vertex/index buffer objects are completely opaque, and can not be retrieved to Lua
    - This is because otherwise the CPU/GPU data could get out of sync with each other.
- `gpu` meshes only store the data on the GPU.
  - `:setVertices` / `:setIndices` will write directly to GPU buffers without an extra copy to CPU arrays.
  - `:getVertices` / `:getIndices` will do a very very slow readback
  - You can get the underlying Buffer objects, allowing you to generate mesh data with compute shaders
  - No `:computeBoundingBox` is possible, but you can still do `:setBoundingBox`.
  - Creating a Mesh from a Buffer implies GPU storage mode.


bjornbytes:
Merging because I want to make some more improvements to buffer formats that will impact Mesh.  I think this is in a pretty good state, but bugs are to be expected for a while as more testing/refactoring is done while it cooks on `dev`.

PR #689, opened 2023-07-21
Add world:queryTriangle(v1, v2, v3, function(shape) end)

xiejiangzhi:


bjornbytes:
Thanks for the PR.  I'm kind of on the fence about this one.  It seems like it would be more flexible to have the `queryShape` function so you can pass a MeshShape with any set of triangles.  As it stands this requires a new MeshShape for every triangle you want to query.  That is also true of querySphere/queryBox, but it's common to build a mesh out of multiple triangles, and I get the sense that mesh shapes in ODE have a bit more overhead than other simpler shapes.

Additionally, whereas box/sphere are 2 of the more "basic" 3D shapes that seem common to query, triangles are slightly more esoteric and open the door for adding things like queryPoint, queryPlane, queryCircle, etc.  Maybe it's good to strive to have a full set of collision detection functions like that eventually, but I sense it would be good to design that system holistically instead of incrementally.

Closing for now, but always open to more discussion about it.

PR #690, opened 2023-07-26
Make window resizable by default

jmiskovic:
I prefer flexible windows during development, and the `conf` can still set it to false on rare occasions it's needed. To me resizable by default makes most sense.

But when we change the default value we kind of break the API, which is not nice.

### DIFF
--- a/etc/boot.lua
+++ b/etc/boot.lua
@@ -45,7 +45,7 @@ function lovr.boot()
       width = 720,
       height = 800,
       fullscreen = false,
-      resizable = false,
+      resizable = true,
       title = 'LÖVR',
       icon = nil
     }


bjornbytes:
One minor reason to keep it false: on tiling window managers on Linux (i3, sway, etc.), non-resizable windows will spawn as "floating" windows with their true dimensions, whereas resizable windows will open as another tile with an arbitrary size.  It's possible to configure certain windows to always be floating by matching the window class, however LÖVR currently isn't setting an X11 window class (it could, and probably should).  Still, it would be nice if tiling windows kept the LÖVR window as floating by default without people having to add an override in their window manager or conf.lua.

Issue #274, opened 2020-07-05
Task System


bjornbytes:
Currently LÖVR is still pretty much single threaded.  On multi-core machines the framework is only able to take advantage of a tiny fraction of the total power of the machine.  Moving to modern graphics APIs also especially benefits from being able to do rendering work on multiple threads.

The `lovr.thread` module lets you create threads and communicate between them, but it's pretty difficult to use.  Also, giving people control over threads directly has a lot of pitfalls.  Often times in this setup, people will create too few or too many threads, and getting an optimal setup that minimizes stalls, starvation, context switching, bugs, etc. is difficult or impossible.

Instead of a thread-oriented approach, a higher level task-oriented approach has some advantages.  In this setup, LÖVR would internally create a global pool of worker threads (usually one for every core on the machine), and tasks could be scheduled to run on the threads.  In Lua, a task would just be a function plus argument(s), which would get dumped to bytecode and `load`ed on whichever threads it runs on.  A scheduling system would ensure that tasks are distributed among the threads.  Some very simple pseudocode:

```lua
function lovr.load()
  lovr.thread.schedule(function()
    while true do
      -- some task that runs once per second, like
      -- checking file modification times for reloading

      -- coroutines could let tasks sleep or wait
      -- for something else to finish
      coroutine.yield(1)
    end
  end)
end

function lovr.update(dt)
  -- the update function could be broken up into
  -- lots of different tasks that get run on threads
  -- these tasks could even spawn other tasks

  -- simple fork/join to update objects in parallel chunks of 100
  local handle = lovr.thread.schedule(function(objects)
    for i = 1, #objects, 100 do
      lovr.thread.schedule(updateObjects, { i, i + 99 })
    end
  end)

  -- do some other work while objects are updating

  lovr.thread.wait(handle)
end
```

The advantages of this are:

- Lua code wouldn't need to create and manage threads.  Instead regular functions are scheduled, which is easier to use and understand.
- The number of threads is abstracted, you could use a conf.lua flag to switch between single-threaded tasks and saturating all the cores.  People don't have to worry about creating the correct number of threads.
- A higher level task system would allow for declaratively defining task dependencies.  You could schedule something and say "by the way, wait for the frustum culling task to finish before you do this", instead of needing to explicitly set up channels or signals.

Restrictions/Challenges

[...]

PR #691, opened 2023-07-28
HTTP(S)

bjornbytes:
Early draft of an HTTP module (or function/plugin, TBD).  Definitely copying LÖVE's homework a bit for this, (lua-https was too difficult to integrate, especially on Android).

API:

```lua
local status, data, headers = lovr.http.request(url, {
  method = nil, --> nil for auto GET/POST based on whether data is provided
  data = nil, --> string, Blob, or table (table gets url encoded as application/x-www-form-urlencoded)
  headers = {},
  timeout = false --> can be integer for seconds
})
```

Returns 0 on failure.

Request is synchronous, asynchronous requests would require machinery from #274.

Does not support multipart/form-data request bodies.

Does not support folding multiline headers.

Backends:

- [x] linux (curl)
- [x] windows (wininet)
- [x] macos (nsurl)
- [ ] android (java)
- [ ] emscripten (fetch)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -517,6 +517,10 @@ if(LOVR_ENABLE_HTTP)
     src/api/l_http.c
     src/modules/http/http.c
   )
+
+  if(WIN32)
+    target_link_libraries(lovr wininet)
+  endif()
 else()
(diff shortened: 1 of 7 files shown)


bjornbytes:
This was moved to the [lovr-http](https://github.com/bjornbytes/lovr-http) plugin.

PR #692, opened 2023-07-29
Split glsl function: initSurface = setupSurfaceData + calcSurfaceData

xiejiangzhi:


etc/shaders/lovr.glsl
```diff
@@ -273,8 +265,26 @@ void initSurface(out Surface surface) {
   if (flag_clearcoatTexture) surface.clearcoat *= getPixel(ClearcoatTexture, UV).r;
 
   surface.clearcoatRoughness = Material.clearcoatRoughness;
+}
+
+void calcSurfaceData(inout Surface surface) {
+  surface.geometricNormal = surface.normal;
```
bjornbytes, line 271: Note: This sets `geometricNormal` to the bent normal, it's supposed to only use the raw normal from the mesh.

bjornbytes:
Thanks, anything that makes `Surface` more flexible is a step in the right direction I think.  As I'm thinking about this it would be helpful to see how your shader intends on using this.

bjornbytes:
Here's a proposal (pseudocode)

```glsl
float getMetalness() {
  float metalness = Material.metalness;
  if (flag_metalnessTexture) surface.metalness *= getPixel(MetalnessTexture, UV).b;
  return metalness;
}

// getRoughness, getBaseColor, getOcclusion, etc.

Surface getEmptySurface() {
  surface.metalness = 1;
  surface.roughness = 1;
  surface.baseColor = vec4(1);
  surface.glowColor = vec3(0);
  surface.occlusion = 1;
  return surface;
}

Surface getSurface() {
  surface.metalness = getMetalness();
  surface.roughness = getRoughness();
  surface.baseColor = getBaseColor();
  surface.glowColor = getGlowColor();
  surface.occlusion = getOcclusion();
  // etc.
  return surface;
}

void initSurface(inout surface) {
  surface.position = 
  surface.geometricNormal =
  surface.view =
  surface.reflection =

  surface.fo = mix();
  surface.diffuse = mix();
  surface.roughness = clamp();
  surface.roughness2 = surface.roughness * surface.roughness;
}
```

It's pretty similar to yours, but:

- There's a function to get an empty "blank" surface if you don't want to pay the cost of loading all the material data.
- There are functions to load individual properties from the material.  It can be useful if you're doing your own lighting without Surface/getLighting, or if you want to use a blank surface but keep some of the material's properties.
- You can do `initSurface(getSurface())` instead of 3 lines (I think this works in GLSL).

Maybe `getDefaultSurface` is a better name than `getSurface`, not sure.

xiejiangzhi:
Do we need to modify such as variable name, surface field deletion. Because I don't understand the specification of glTF, I will only try to make minor changes to the existing code. If this requires a lot of changes, it may be difficult for me to change what you expect without the description of the surface data structure. result.

If you provide the structure of your expected Surface, I can modify it. Of course, you can also close this PR and modify it yourself (I feel that you are almost finished writing the code).

bjornbytes:
Ok, I can implement it.  End result will look something like this, pretty sure it will work for your original need:

```glsl
vec4 lovrmain() {
  Surface surface = getSurface();
  surface.metalness = mymetalness;
  surface.roughness = myroughness;
  initSurface(surface);

  // getLighting, etc.

  return ...;
}
```

Thanks again for the suggestion and PR!

bjornbytes:
Ok done.

```glsl
Surface surface = getDefaultSurface(); // replaces initSurface

Surface surface = newSurface(); // get a default/empty surface, allowing you to set your own properties
applyMaterial(surface); // loads all the data/textures from the material into the surface (optional)
surface.metalness = mymetalness; // you can set custom properties
finalizeSurface(surface); // computes extra variables and clamps things

// getLighting(surface, ...)
```

PR #696, opened 2023-08-18
feat: overlay layers

SpookySkeletons:
This switches overlay from a bool to an int everywhere so you can utilize layer height composition.

Untested and produces one compiler warning due to the int casting implicitly to a bool.

Submitting sooner rather than later, take a look if you will.


SpookySkeletons:
Tested, functional, just needs an intelligent way to cast int.

bjornbytes:
Thanks for the patch!

PR #702, opened 2023-09-13
Fix getter for distance joint anchors

jmiskovic:


bjornbytes:
:smiling_face_with_tear: thanks

bjornbytes:
https://github.com/bjornbytes/lovr-docs/blob/master/tests/joints.lua

so it doesn't happen again somehow

PR #704, opened 2023-09-18
Surface Improvements

bjornbytes:
Restores ability to open window after initializing graphics module.

Surface is created lazily instead of being required upfront.

Use native platorm handles instead of GLFW callbacks.

Some minor reorganization around core/gpu present API and xr transitions.

Linux links against libxcb/libX11/libX11-xcb for XGetXCBConnection.

Still filling in and testing esp macOS.


bjornbytes:
Here goes nothing

PR #713, opened 2023-10-20
Layers

### LINKED ISSUE #438: Compositor Layers
Several people have expressed interest in the idea of supporting compositor layers.

Background (no pun intended): Compositor layers are super cool.  Basically instead of sending 1 texture to the XR runtime that contains everything in the scene (skybox, scene geometry, UI elements), you can send things in several different pieces and make the poor runtime deal with compositing all of them into the final eye textures.  There are many advantages:

- The compositor is way better at positioning and reprojecting stuff than you are.  So if you sent a cubemap layer with your skybox, or a 2D quad layer with an image, in theory it will feel much more "stable" than normal.
- Quad layers make better use of pixels than the warped eye textures, so you can get better clarity for UI/fonts.
- Might save rendering time by not rerendering static parts of the scene (but I guess the runtime still has to do it, so not sure if there are any real savings).

Also, submitting depth buffers to the runtime (for better reprojection and occlusion) is achieved via submitting an extra layer.

Here are some ideas for supporting these in LÖVR.

- `lovr.headset.newLayer` creates a compositor layer.  A Layer has a type (2d, 3d, cubemap, equirect, cylinder).  Not all types are supported by all runtimes, so their should probably be a way to check which types are supported.  Also, not all headset drivers will even support layers.  In that case, just one 3d layer will be supported.
- `Layer:getCanvas`.  Layers are backed by swapchains and manage a Canvas object.  You can retrieve the Canvas to render to it if you want to update the layer's contents.
- `lovr.headset.renderTo` will probably go away, instead there will be `lovr.headset.submit` that submits layers to the runtime.
- `lovr.headset.getLayers` and `lovr.headset.setLayers`.  These will control which layers get submitted to the compositor.  Originally I wanted a stateless API like `lovr.headset.submit(...layers)`, but since this will probably get called in `boot.lua` there needs to be some way to get the list of layers to it.
- Various metadata accessors on Layers.  Quad layers can have a pose in the world, an origin (head/floor), dimensions in meters, and an eye affinity (which eye it renders to, in theory you could make a stereo quad layer by rendering 2 quad layers in the same place with different eye affinities?).  Cylinders can have angle/radius.  All layers can have a tint for doing color fades (without rerendering their contents). etc.
- You can change the environment blend mode that says how layers should blend with layers behind them.
- There will be a default layer.  For stereo setups this will be a single 3d layer.  lovr.draw draws to the default layer (or the window if headset module isn't active).
- A note on cameras: Before rendering to a layer, its Canvas camera will magically get set up to use something sensible (orthographic for 2D quad, 6 directional views for cubemap, or viewer-synchronized eye views for a 3d layer).  You're allowed to change these though.  In particular, for 3D layers you're encouraged to update their camera matrices as late as possible, to get better prediction (on the gpu branch I think you'll be able to set canvas cameras after recording a canvas batch and it'll still apply to the batch).
- If a Layer is only going to get recorded once (i.e. a static image), you can set a special flag that enables some optimizations.  It would be convenient to create layers from images/textures directly.

bjornbytes:
Adds a new object called `Layer` that puts a textured quad in 3D space.

The difference between this and `Pass:plane` is that it's higher quality because the VR runtime can sample the pixels later in the composition process (better tracking, higher pixel density).  The main use case is higher quality for 2D UI/text in VR.

API is:

- `lovr.headset.newLayer(width, height)`
- `lovr.headset.getLayers()`
- `lovr.headset.setLayers({layers})` (max of 10)
- `Layer:get/setPosition`
- `Layer:get/setOrientation`
- `Layer:get/setPose`
- `Layer:get/setSize` (in meters)
- `Layer:get/setViewMask` (set which eyes layer shows up in, for stereo quads)
- `Layer:get/setViewport` (only show a sub-rectangle of the layer texture in the quad)
- `Layer:get/setSupersample` (Quest-only)
- `Layer:set/setSharpen` (Quest-only)
- `Layer:getTexture`
- `Layer:getPass`

Layers will be depth tested so they sort properly with the main view, but this is only supported on Quest.  For this reason, layers should maybe be considered experimental for cross-platform applications.

Other assumptions/constraints/YAGNI:

- Texture formats / usage flags / mipmaps aren't configurable
- Layer color/alpha tint (this would be nice, but not enough runtime support)
- Layer opaque/premultiplied flag (always transparent + unpremultiplied)
- Layer types other than quad (there are cube/equirect layers, but not much runtime support)
- Layer coordinate space origin (can just :setPose with the pose you want)
- No static swapchains
- No control over default layer

It would also be nice if the simulator composited layers.

Closes #438


bjornbytes:
They work!

- Runtime gets grumpy if you submit a layer without rendering to it at least once (should clear or at least acquire a texture on create)
- Layer pass should set an identity view matrix and orthographic projection matrix.
- openxr_setLayers should check layer count too
- should clamp/assert swapchain dimensions
- Simulator needs layer implementation
- lovr.mirror could composite layers onto window (viewport is kinda hard though boo)

src/modules/headset/headset_openxr.c
```diff
@@ -405,6 +425,89 @@ static XrControllerModelKeyMSFT getControllerModelKey(Device device) {
...
+
+  XR(xrEnumerateSwapchainImages(swapchain->handle, MAX_IMAGES, &swapchain->textureCount, (XrSwapchainImageBaseHeader*) images), "Failed to query swapchain images");
+
+  for (uint32_t i = 0; i < swapchain->textureCount; i++) {
+    swapchain->textures[i] = lovrTextureCreate(&(TextureInfo) {
+      .type = stereo ? TEXTURE_ARRAY : TEXTURE_2D,
+      .format = depth ? state.depthFormat : FORMAT_RGBA8,
+      .srgb = !depth,
+      .width = state.width,
+      .height = state.height,
```
bjornbytes: Should be width/height

src/api/l_headset_layer.c
```diff
@@ -0,0 +1,185 @@
...
+  return 4;
+}
+
+static int l_lovrLayerSetViewport(lua_State* L) {
+  Layer* layer = luax_checktype(L, 1, Layer);
+  uint32_t viewport[4];
+  viewport[0] = luax_optu32(L, 2, 0);
+  viewport[1] = luax_optu32(L, 3, 0);
+  viewport[2] = luax_optu32(L, 4, ~0u);
+  viewport[3] = luax_optu32(L, 5, ~0u);
```
bjornbytes: `~0u` isn't being handled properly yet

src/modules/headset/headset.h
```diff
@@ -167,6 +181,22 @@ typedef struct HeadsetInterface {
...
+  Layer* (*newLayer)(uint32_t width, uint32_t height);
+  void (*destroyLayer)(void* ref);
+  void (*getLayerPose)(Layer* layer, float* position, float* orientation);
+  void (*setLayerPose)(Layer* layer, float* position, float* orientation);
+  void (*getLayerSize)(Layer* layer, float* width, float* height);
+  void (*setLayerSize)(Layer* layer, float width, float height);
+  ViewMask (*getLayerViewMask)(Layer* layer);
+  void (*setLayerViewMask)(Layer* layer, ViewMask mask);
+  void (*getLayerViewport)(Layer* layer, uint32_t* viewport);
+  void (*setLayerViewport)(Layer* layer, uint32_t* viewport);
```
bjornbytes: Should switch to signed ints for viewport

src/modules/headset/headset_openxr.c
```diff
@@ -2544,36 +2713,32 @@ static void openxr_submit(void) {
...
+      state.layer.layerFlags = 0;
     }
-  }
-
-  if (state.frameState.shouldRender) {
-    XR(xrReleaseSwapchainImage(state.swapchain[COLOR], NULL), "Failed to release color swapchain image");
 
-    if (state.features.depth) {
-      XR(xrReleaseSwapchainImage(state.swapchain[DEPTH], NULL), "Failed to release depth swapchain image");
+    if (state.features.layerDepthTest && state.features.depth) {
```
bjornbytes: Maybe only chain this if `state.layerCount > 0`

bjornbytes:
This is good enough for an initial version!  They don't work on the simulator or show up in the mirror window yet, but all the basic functionality is there.  These will be kind of an experimental feature for a while since depth-tested layers aren't widespread, but they provide a lot of value on Quest 2+ so it's cool to be able to start playing with them.

PR #719, opened 2023-11-05
Boot rework

bjornbytes:
This is a complete overhaul of the startup process for LÖVR!  Some of the build system stuff isn't finished yet, but in the meantime I'm gonna write about it here since the design is somewhat stable.

The key points are:

- The nogame screen is distributed as a fused zip instead of being embedded in the C code.
- All CLI behavior is moved into a new file, `arg.lua`.  This is also in the fused zip.  It handles all argument parsing, including finding the project passed in on the command line, --help/--version, etc.

The impact of these changes are:

- The nogame screen and argument handling can be removed entirely when creating a fused project.  This makes projects a little bit smaller and allows the nogame screen and CLI to be as big as they want without bloating every shipped app.
- Fused projects will no longer respond to lovr's development CLI arguments.  It wasn't a *huge* deal, but it always kinda bothered me that you could do `game.exe --help` and see a weird message about lovr.
- Because LÖVR will always be a zip, and that zip will always be mounted, you can do some interesting things:
  - You can edit the nogame screen.
  - You can add your own CLI behavior, and it will work for every app.
  - You can add libraries to the zip and use them in any project (!)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -779,11 +779,7 @@ elseif(ANDROID)
   if(LOVR_BUILD_EXE)
     set(ANDROID_JAR "${ANDROID_SDK}/platforms/${ANDROID_PLATFORM}/android.jar")
     set(ANDROID_TOOLS "${ANDROID_SDK}/build-tools/${ANDROID_BUILD_TOOLS_VERSION}")
-
-    # If assets are included in the apk then add '-A assets' to aapt, otherwise don't add any flags
-    if(ANDROID_ASSETS)
-      set(ANDROID_ASSETS -A ${ANDROID_ASSETS})
-    endif()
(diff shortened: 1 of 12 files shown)


bjornbytes:
This is confirmed working on Windows and Linux.  macOS will probably need special handling for `LOVR_BUILD_BUNDLED=ON` builds, Android will definitely need special handling but it might be as simple as defaulting `ANDROID_ASSETS` to the `etc/nogame` folder and skipping the regular nogame exe packaging.

bjornbytes:
Ok, macOS and Android work as well now!

There might be one issue where the nogame's conf.lua affects projects that are missing a conf.lua of their own, need to test that but otherwise this is probably ready to merge?

PR #720, opened 2023-11-18
Auto wake up collider when calling API to apply force or change force.

xiejiangzhi:


bjornbytes:
Nice, that was simple.  Thanks 💜

PR #722, opened 2023-11-22
Add quat:ToEuler and quat:fromEuler

xiejiangzhi:


bjornbytes:
I don't know much about euler angles.  Is ZXY the most common order to use for the different angles?

xiejiangzhi:
> I don't know much about euler angles. Is ZXY the most common order to use for the different angles?

Me too. I’ve read some codes and the most commonly used ones are zyx
But I just used it and found that in the game, the values of pitch and yaw are reversed.

xiejiangzhi:
After some testing, it was discovered that this conversion had a bug. . .
Because I don’t understand this conversion, the code is modified after copying.

xiejiangzhi:
OK, I have fixed the bug, 
Now, it's result is the same as (three.js math) https://www.andre-gaschler.com/rotationconverter/  
Rotation order is YXZ

bjornbytes:
I'd like to merge this, but can we rename the functions to get/set prefix instead of from/to, to match the naming of other methods?

- get/setEuler
- get/setAngles
- get/setEulerAngles

Also, the math should be done with floats instead of doubles, to avoid conversions (sinf/cosf/atan2f instead of sin/cos/atan2, and .5f instead of .5).  I can do this after merging though.

xiejiangzhi:
It has been modified as you said

bjornbytes:
Thanks!  And sorry about the delay

src/api/l_math_vectors.c
```diff
@@ -1574,6 +1594,8 @@ const luaL_Reg lovrQuat[] = {
   { "direction", l_lovrQuatDirection },
   { "conjugate", l_lovrQuatConjugate },
   { "slerp", l_lovrQuatSlerp },
+  { "getEuler", l_lovrQuatGetEuler },
+  { "SetEuler", l_lovrQuatSetEuler },
```
jmiskovic, line 1598: Should be `setEuler`

PR #724, opened 2023-11-26
Use virtual memory in vector pools

jmiskovic:
The issue when growing the pool is gone, and the rest of pool vector operations work as well as before (from my limited testing).


bjornbytes:
Awesome, I'll try to do more testing on this as well!

PR #728, opened 2024-01-02
Shape Vertex Compression

bjornbytes:
Compresses the vertex format used for shapes down from 32 bytes to 16 bytes, a 50% savings!

Positions were changed from `f32x3` (12) to `sn16x4` (8).  This requires shape vertex positions to be -1 to 1, which is true for almost every shape except capsule and torus which need to be scaled.

Normals were changed from `f32x3` (12) to a new `sn10x3` (4) format.  This format is not supported by Quest 1!

UVs were changed from `f32x2` (8) to `un16x2` (4).  All UVs were 0-1 already.

So far I've tested on Intel HD and haven't noticed more than a few % difference in CPU vertex generation and GPU vertex shading performance.  Still, using half as much temporary buffer memory is really nice.  I've also heard that mobile GPUs really like smaller vertex formats, so it will be interesting to test on there.


bjornbytes:
Dropping this for now.  After more testing the CPU overhead is higher than I was originally seeing, and the GPU speedup is smaller than I was hoping for.  This was still only testing on discrete/integrated desktop GPUs.  It will probably make more sense for Model, which can do the conversion once at load time and tends to have more vertices.

PR #732, opened 2024-01-21
Model Vertex Compression

bjornbytes:
Changes models to use packed normals/tangents.  This reduces the size of model vertices from 52 to 32 bytes (~40%).

There's a new `sn10x3` DataType that can be used in buffers (similar to the existing `un10x3` type).  Note that this format is not supported on the Quest 1!

Shaders have new `packSnorm10x3` and `unpackSnorm10x3` helpers, and the animation compute shaders unpack/repack the normals and tangents.

Still gotta profile to see if runtime cost is significant, especially for animation.


src/modules/data/modelData.c
```diff
@@ -243,6 +244,20 @@ void lovrModelDataCopyAttribute(ModelData* data, ModelAttribute* attribute, char
...
+        float x = ((float*) src)[0];
+        float y = ((float*) src)[1];
+        float z = ((float*) src)[2];
+        float w = attribute->components == 4 ? ((float*) src)[3] : 0.f;
+        *(uint32_t*) dst =
+          ((((uint32_t) (int32_t) (x * 511.f)) & 0x3ff) <<  0) |
+          ((((uint32_t) (int32_t) (y * 511.f)) & 0x3ff) << 10) |
+          ((((uint32_t) (int32_t) (z * 511.f)) & 0x3ff) << 20) |
+          ((((uint32_t) (int32_t) (w * 2.f)) & 0x003) << 30);
+      }
```
bjornbytes, line 259: Should add `else { lovrUnreachable(); }`.  Note that glTF requires normal vectors to be f32x3 (and tangents to be f32x4).

bjornbytes:
Rendering models on mobile GPUs seems to be significantly faster now (I was testing an animated model with ~65k vertices and saw 25% faster GPU timestamps for the render pass, which seems pretty crazy).  No perf change was measured on desktop.

PR #735, opened 2024-01-30
Jolt physics engine integration

jmiskovic:
Almost done :sweat_smile: 

Some loose ends:
* callbacks or way to fetch colliding shapes - requires API changes I think
* inertia matrix getter/setter - is anyone actually using these?
* compound shapes
* increasing the number of tags to 32 might have broken the implementation

At current state it's already useful, so I think it's ready for review and further discussions.


src/modules/physics/physics_jolt.c
```diff
@@ -0,0 +1,1474 @@
+#include <stdlib.h>
+#include "core/maf.h"
+#include "util.h"
+#include "physics.h"
+#include "joltc.h"
+
+#define MAX_BODIES 2048
```
xiejiangzhi: This value is too small and only sufficient for testing.
In the official hello world, the recommended value for the production environment is around 65535.
MaxBodyPairs, MaxContactConstraints also need to be larger
https://github.com/jrouwe/JoltPhysics/blob/cc49e07d7c4eb854a60cb21ea0abdfcc93511410/HelloWorld/HelloWorld.cpp#L239

CMakeLists.txt
```diff
@@ -158,27 +162,37 @@ endif()
...
+      include_directories(deps/ode/include "${CMAKE_CURRENT_BINARY_DIR}/ode/include")
+      set(LOVR_PHYSICS_LIB ode)
     endif()
-    include_directories(deps/ode/include "${CMAKE_CURRENT_BINARY_DIR}/ode/include")
-    set(LOVR_ODE ode)
+  elseif(LOVR_PHYSICS_LIBRARY STREQUAL "JOLT")
+    set(BUILD_SHARED_LIBS OFF)
+    include_directories(deps/jolt-physics-sharp/src/joltc)
+    add_subdirectory(deps/jolt-physics-sharp/src/joltc jolt)
+    target_compile_options(joltc PRIVATE -Wno-comment)
```
bjornbytes: Current Windows build failure would be fixed by wrapping this in `if(NOT MSVC)`

src/modules/physics/physics_jolt.c
```diff
@@ -0,0 +1,1474 @@
...
+  const JPH_NarrowPhaseQuery* query = JPC_PhysicsSystem_GetNarrowPhaseQueryNoLock(world->physics_system);
+  float transformArray[16] = {
+    scale[0], 0.f, 0.f, 0.f,
+    0.f, scale[1], 0.f, 0.f,
+    0.f, 0.f, scale[2], 0.f,
+    position[0], position[1], position[2], 1.f
+  };
+  array_to_rmatrix_struct(transformArray, &transform);
+  JPH_AllHit_CastShapeCollector_Reset(cast_shape_collector);
+  JPH_NarrowPhaseQuery_CastShape(query, queryBox, &transform, &direction, &base_offset, cast_shape_collector);
```
bjornbytes, line 255: I'm curious: would it be better to use `CollideShape` for these shape queries, since the direction is unused?

src/modules/physics/physics_jolt.c
```diff
@@ -0,0 +1,1474 @@
...
+void lovrWorldSetLinearDamping(World* world, float damping, float threshold) {
+  lovrLog(LOG_WARN, "PHY", "Jolt doesn't support global LinearDamping option");
+}
+
+void lovrWorldGetAngularDamping(World* world, float* damping, float* threshold) {
+  lovrLog(LOG_WARN, "PHY", "Jolt doesn't support global AngularDamping option");
+}
+
+void lovrWorldSetAngularDamping(World* world, float damping, float threshold) {
+  lovrLog(LOG_WARN, "PHY", "Jolt doesn't support global AngularDamping option");
```
bjornbytes: I think the World damping accessors in ODE just set the default damping for new bodies, so maybe these could be supported (but we could also think about removing them).

bjornbytes:
I don't think I have any big suggestions.  We could merge this right away and keep improving it on `dev` if the CMake was changed to use ODE by default for now.

src/modules/physics/physics_jolt.c
jmiskovic: For now I've added both dampings, as well as sleeping allowed.

Small note - the docs could be clarified. From wording its easy to assume that setting the world damping at any time will change the damping of previously created colliders.

src/modules/physics/physics_jolt.c
jmiskovic: In future I would like to expand the current API to support casting, so that shape query is just a special case of shape casting with zero direction.

CMakeLists.txt
jmiskovic: I made the proposed change but at the moment I don't have access to the Windows machine to test.

src/modules/physics/physics_jolt.c
jmiskovic: Thanks for catching this, I've set the values to the ones in the linked document.

src/modules/physics/physics_jolt.c
bjornbytes: Clarified the docs for this

amerkoleci:
Hi 
I'm the maintainer of the C layer that I use for the C# bindings, feel free to request any missing bindings/or submit PR with stuff you need.

Cheers

bjornbytes:
This was a massive effort, thank you so much @jmiskovic!  Once I finish up some other stuff I'll work on polishing it some more and work on adding some of the ideas in the physics issue.  Can't wait to switch over to Jolt as the default!

PR #743, opened 2024-02-14
Headset simulator supersampling

jmiskovic:
The idea is to be able to draw at half resolution without effort of going through render texture.

The calculation of virtual hand position needed to be corrected for resolution change, but the mouse coordinates are unchanged. They still work in window coordinate system, as if supersampling is disabled.

### DIFF
--- a/src/modules/headset/headset_simulator.c
+++ b/src/modules/headset/headset_simulator.c
@@ -136,6 +136,8 @@ static bool simulator_isSeated(void) {
 static void simulator_getDisplayDimensions(uint32_t* width, uint32_t* height) {
   float density = os_window_get_pixel_density();
   os_window_get_size(width, height);
+  *width *= state.config.supersample;
+  *height *= state.config.supersample;
   *width *= density;
   *height *= density;
 }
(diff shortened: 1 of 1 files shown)


src/modules/headset/headset_simulator.c
```diff
@@ -368,6 +368,8 @@ static Pass* simulator_getPass(void) {
 
   uint32_t width, height;
   simulator_getDisplayDimensions(&width, &height);
+  width *= state.config.supersample;
+  height *= state.config.supersample;
```
bjornbytes: I think `simulator_getDisplayDimensions` should factor in the supersample instead of here, so that `lovr.headset.getDisplayDimensions` returns the supersampled size (like OpenXR does).

jmiskovic: I considered that one, but it's also used by `simulator_getViewAngles` which would return wrong values if the change was made there.

jmiskovic: Wait that's not true at all :smile: 

I've pushed the improvement.

bjornbytes:
Looks good!  Merging after build

jmiskovic:
Last push was just removing the extra newline.

PR #751, opened 2024-02-29
Fixed nogame.zip creatrion

brainrom:
`logo.spv` wasn't packed in `nogame.zip`, therefore nogame is failed to start. Now all files (maybe, other files will be added in the future) in nogame folder are packed.
This affects only PC version.


bjornbytes:
Oops, I only tested on tup, thank you for the fix!

PR #754, opened 2024-03-03
Make lovrAssert act like a function

brainrom:
`do {} while(0)` is the perfect way to make function-like macro act like a real function (require ;, encapsulate local variables)


bjornbytes:
Personally I don't have any preference on this.

PR #757, opened 2024-03-11
Writable Archives

bjornbytes:
Adds the ability to mount an archive for writing.  This means LÖVR can write anywhere on the filesystem, instead of only the save directory.

How it works: `lovr.filesystem.mount` takes a `MountMode` after the mountpoint.  The mount mode can be either `read` (the default) or `readwrite`.

Filesystem functions that previously targeted the save directory (`write`, `createDirectory`, `remove`) will now use the virtual filesystem: they search through the mounted archives, find the first one in `readwrite` mode with a mountpoint that contains the input path, and write to that archive.

The save directory isn't special anymore -- its archive is just `readwrite` instead of `read` now.

I haven't tested this code very well yet.

### DIFF
--- a/src/api/l_filesystem.c
+++ b/src/api/l_filesystem.c
@@ -5,6 +5,12 @@
 #include <stdlib.h>
 #include <string.h>
 
+StringEntry lovrMountMode[] = {
+  [MOUNT_READ] = ENTRY("read"),
+  [MOUNT_READWRITE] = ENTRY("readwrite"),
+  { 0 }
+};
+
 StringEntry lovrOpenMode[] = {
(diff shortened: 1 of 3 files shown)


bjornbytes:
I found a bug in `getSourceDirectory`.  If you do `lovr .`, getSourceDirectory returns `.` which is not a valid path to mount.  It should be converted to an absolute path.

(Actually I'm not sure why that fails to mount, need to look into it).

Ah, it fails because `.` is already mounted...probably need to relax the logic for mounting duplicate archives now.

bjornbytes:
Archiving this

PR #759, opened 2024-03-12
Fix boot.lua merge

brainrom:
boot.lua was broken during merging with stable branch.


bjornbytes:
Gah!  Thanks

I added a pre-commit hook to catch this in the future.

PR #763, opened 2024-04-11
Fix setMass; correct terrain sampling

jmiskovic:
For terrain sampling fix, this is the before and after.

image

image

```Lua
world = lovr.physics.newWorld()
collider = world:newTerrainCollider(4, function(x, z) return 1 end, 128)

function lovr.draw(pass)
  pass:setProjection(1, mat4():orthographic(-3, 3, 3, -3, 0.01, 100))
  pass:plane(0, 0, 0,  4, 4, -math.pi/2, 1,0,0, 'line')
  for x = -5, 5, 0.2 do
    for z = -5, 5, 0.2 do
      o = vec3(x,  100, z)
      t = vec3(x, -100, z)
      world:raycast(o, t,
        function(shape, x, y, z, nx, ny, nz)
         pass:sphere(x, y, z, 0.05)
         pass:line(x, y, z, x+nx, y+ny, z+nz)
        end)
    end
  end
end
```

### DIFF
--- a/src/api/l_physics_shapes.c
+++ b/src/api/l_physics_shapes.c
@@ -114,8 +114,8 @@ Shape* luax_newterrainshape(lua_State* L, int index) {
     uint32_t n = luax_optu32(L, index + 1, 100);
     float* vertices = lovrMalloc(sizeof(float) * n * n);
     for (uint32_t i = 0; i < n * n; i++) {
-      float x = scaleXZ * (-.5f + ((float) (i % n)) / n);
-      float z = scaleXZ * (-.5f + ((float) (i / n)) / n);
+      float x = scaleXZ * (-.5f + ((float) (i % n)) / (n - 1));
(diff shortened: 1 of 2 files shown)


bjornbytes:
Couldn't this also be `/ (n - 1)`  instead of `/ (n + 1)`?  That way the callback gives you this when samples = 2

```
-2.00, -2.00,
2.00, -2.00,
-2.00, 2.00
2.00, 2.00
```

instead of this

```
-2.00, -2.00
-.667, -2.00
-2.00, -.667
-.667, -.667
```

Also for some reason Jolt crashes when samples == 2, but that's a separate thing.

jmiskovic:
Yup, fixed. The screenshot was with `n-1` but I submitted wrong version.

bjornbytes:
Cool, looks good!

PR #772, opened 2024-05-09
Assert the input range in joint limits setters

jmiskovic:
The Jolt already asserts these values, which results in segfault instead of helpful error message.

For cone joint the value out of range will now throw error instead of silent clamp.

### DIFF
--- a/src/modules/physics/physics_jolt.c
+++ b/src/modules/physics/physics_jolt.c
@@ -1938,7 +1938,8 @@ float lovrConeJointGetLimit(ConeJoint* joint) {
 }
 
 void lovrConeJointSetLimit(ConeJoint* joint, float angle) {
-  JPH_ConeConstraint_SetHalfConeAngle((JPH_ConeConstraint*) joint->constraint, CLAMP(angle, 0.f, (float) M_PI));
+  lovrCheck(angle >= 0 && angle <= (float) M_PI, "Cone joint angle limit must be between 0 and PI");
+  JPH_ConeConstraint_SetHalfConeAngle((JPH_ConeConstraint*) joint->constraint, angle);
 }
 
 // DistanceJoint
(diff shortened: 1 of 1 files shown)


src/modules/physics/physics_jolt.c
```diff
@@ -1975,6 +1976,8 @@ void lovrDistanceJointGetLimits(DistanceJoint* joint, float* min, float* max) {
 }
 
 void lovrDistanceJointSetLimits(DistanceJoint* joint, float min, float max) {
+  lovrCheck(min <= 0.f, "Distance joint lower distance limit must be negative");
```
bjornbytes: min can be positive, it just needs to be <= max (e.g. min distance of 1 will prevent colliders from getting closer than 1m, min == max maintains constant distance).

src/modules/physics/physics_jolt.c
```diff
@@ -1975,6 +1976,8 @@ void lovrDistanceJointGetLimits(DistanceJoint* joint, float* min, float* max) {
...
+  lovrCheck(max >= 0.f, "Distance joint upper distance limit must be positive");
```
bjornbytes:
```suggestion
  lovrCheck(max >= 0.f, "Distance joint upper distance limit can not be negative");
```

src/modules/physics/physics_jolt.c
```diff
@@ -2196,6 +2201,8 @@ void lovrSliderJointGetLimits(SliderJoint* joint, float* min, float* max) {
 }
 
 void lovrSliderJointSetLimits(SliderJoint* joint, float min, float max) {
+  lovrCheck(min <= 0.f, "Slider joint lower distance limit must be negative");
+  lovrCheck(max >= 0.f, "Slider joint upper distance limit must be positive");
```
bjornbytes:
```suggestion
  lovrCheck(max >= 0.f, "Slider joint upper distance limit can not be negative");
```

src/modules/physics/physics_jolt.c
```diff
@@ -2196,6 +2201,8 @@ void lovrSliderJointGetLimits(SliderJoint* joint, float* min, float* max) {
 }
 
 void lovrSliderJointSetLimits(SliderJoint* joint, float min, float max) {
+  lovrCheck(min <= 0.f, "Slider joint lower distance limit must be negative");
```
bjornbytes:
```suggestion
  lovrCheck(min <= 0.f, "Slider joint lower distance limit can not be positive");
```

PR #780, opened 2024-06-06
BMFont

bjornbytes:
Adds support for BMFont fonts.  They are useful for rendering fonts in 2D contexts (potentially with the new `Layer` object) or for stylized image fonts or icons/emoji.  Both text and binary BMFont files are supported.  Non-unicode charsets are not supported.

The usage is similar to TTF:

```lua
function lovr.load()
  font = lovr.graphics.newFont('font.fnt')
end

function lovr.draw(pass)
  pass:setFont(font)
  pass:text('hi!', x, y, z)
end
```


bjornbytes:
YEAAAAAAAAAAAAAAAAA

PR #789, opened 2024-06-25
Table Vectors

bjornbytes:
This is a branch that makes vectors use tables instead of userdata.  It is an evolution of what was proposed in #737.  For now, the Lua API tries to match the existing one as closely as possible.  Methods that take vector objects like vec3 can instead take any table, e.g. `pass:sphere({ x, y, z }, radius)`.

Also, there are only permanent vectors now, temporary vectors are removed.  LuaJIT's garbage collector seems to be able to handle the increased GC load for most workloads, and it's possible to pool table vectors to avoid garbage explosion in tight loops.

Why do this?

- Customization.  Since any array-like table can be passed to lovr functions that take vectors, people can use alternative vector libraries or monkeypatch the builtin one.  It isn't as easy to customize the existing userdata vectors, and this has been annoying.
- JIT compilation.  Performance is a bit of a tossup between tables and userdata, but in some cases tables win out since LuaJIT can JIT compile code that uses tables.  The existing userdata vectors prevent JIT compilation and calling methods on them is a lot slower than calling pure Lua functions.
- Simplicity.  It's simpler to have 1 type of vector instead of 2.  The temporary vectors overrode the global lightuserdata metatable, which is hacky.  Implementing all the little vector methods in C required a lot more carefully-written and verbose C code compared to the amount of Lua code required.

It has the following known incompatibilities:

- `__len` is not supported on tables in Lua 5.1 and LuaJIT (without 5.2 compat), so you can't do `#v` and need to do `v:length()` now instead.
- ~~Vectors can't be pushed to Channels anymore, because Channels don't support tables yet (this is still a todo).   Mat4 works~~  Channels support tables now.
- The `rgba` and `stpq` swizzles were dropped.
- `lovr.math.drain` was removed, because there aren't temporary vectors anymore (TODO it could be added back as a deprecated function that does nothing, to avoid breakage).

One thing that's a little bit annoying is that, for quaternions, you can't do `{ angle, ax, ay, az }` like you can for vec3 and instead need to do `quat(angle, ax, ay, az)` to convert it to quaternion components.  This is a small gotcha but I think it's ok.


bjornbytes:
Performance test on IK benchmark:

- LuaJIT (JIT: on)
  - `dev`: 94 FPS
  - `tvec`: 113 FPS
- LuaJIT (JIT: off)
  - `dev`: 94 FPS
  - `tvec`: 81 FPS
- Lua
  - `dev`: 112 FPS
  - `tvec`: 52 FPS
  - `tvec` (no __index metamethod): 60 FPS

bjornbytes:
Archiving this for now, may revisit later.  The performance numbers for vanilla Lua (50% slowdown) were bad enough to hold off on merging this.

PR #791, opened 2024-07-09
Adds in Magic Leap 2 support

SfernandezML:
Just a few minimal changes to get this working on the ML2. 
* Reduces the minimum api level to 29
* Adds a separate build step in the workflow to build an x86_64 apk as well
* Adds in the ML2 hand tracking permission

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -105,6 +105,49 @@ jobs:
         with:
           name: lovr.apk
           path: build/lovr.apk
+  magicleap:
+    name: Magicleap
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Update Packages
+        run: sudo apt update
+      - name: Install Packages
+        run: sudo apt install -y glslang-tools
+      - name: Install Android 29
+        run: ${ANDROID_HOME}/cmdline-tools/latest/bin/sdkmanager --install "platforms;android-29"
+      - name: Checkout
+        uses: actions/checkout@v3
+        with:
+          submodules: 'recursive'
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -105,6 +105,49 @@ jobs:
...
+          -keypass hunter2
+          -storepass hunter2
+          -alias key
+          -keyalg RSA
+          -keysize 2048
+          -validity 10000 &&
+          cmake ..
+          -D CMAKE_TOOLCHAIN_FILE=$ANDROID_NDK/build/cmake/android.toolchain.cmake
+          -D ANDROID_SDK=$ANDROID_HOME
+          -D ANDROID_ABI=x86_64
```
bjornbytes, line 139: Since this is the same as the main Android job except for the ABI, we could probably use a [matrix](https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs) for it.  Basically there would be a matrix section at the top:

```yaml
android:
  name: Android
  strategy:
    matrix:
      abi: [arm64-v8a, x86_64]
```

And then add in `${{ matrix.abi }}` when setting `ANDROID_ABI` and the APK name.  I can do this as a future improvement if you want though.

SfernandezML: that would be great, I am not super familiar with github workflows

bjornbytes:
This is awesome, thanks!

bjornbytes:
I added a build matrix in e5659de0e6b5896c7f0512bb037f85f12f62b241 and it appears to work!

PR #794, opened 2024-07-13
Add collider:moveKinematic API

xiejiangzhi:
Hand physics can be implemented in a simpler way: `hand:moveKinematic(pos, rot, dt)`

### DIFF
--- a/src/api/l_physics_collider.c
+++ b/src/api/l_physics_collider.c
@@ -382,6 +382,16 @@ static int l_lovrColliderSetPose(lua_State* L) {
   return 0;
 }
 
+static l_lovrColliderMoveKinematic(lua_State* L) {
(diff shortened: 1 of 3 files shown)


bjornbytes:
🤔 Should people pass the dt from `lovr.update` or the fixed timestep from `lovr.physics.newWorld`?  Would it require a "fixedUpdate" callback like some other engines have?

xiejiangzhi:
It should be. Just like the current calculation, if you use Lovr's FixedStep, then whether you use moveKinematic or not, you can't calculate the exact position because you can't get the correct physics delta time with update call.
I implemented fixed step myself in my project, without using Lovr's

bjornbytes:
Yeah, I think we either need to add a fixed update callback or remove the fixed timestep.  Either way, it doesn't affect this PR.  Thanks for the patch!

PR #795, opened 2024-07-15
Sampler support border wrap(clamp to zero), and fix mirror wrap

xiejiangzhi:


xiejiangzhi:
By default, borderColor is not set in gpu sampler, so it is clamped to zero.
If you want to set it, you should add a borderColor parameter in the sampler, so that you can set any color.

bjornbytes:
Note: this is not supported (yet) on WebGPU
Note: it clamps to transparent black

But other than that, seems good, should work for all textures from what I can see.

PR #797, opened 2024-07-24
Output label by tostring for texture & pass

xiejiangzhi:


xiejiangzhi:
I think it might be better to use the initial label when LovrPassCreate, which is simpler.

bjornbytes:
Thanks!

PR #803, opened 2024-08-17
Fix infinite loop during model animation caused by integer overflow.

cjbarre:
Update calculations to avoid overflow per discussion in the discord with Bjorn.


cjbarre:
Realizing that max may not make sense here, but you get the idea 😆

bjornbytes:
Thank you for the great investigation, I wouldn't have been able to track this down without trying a bunch of different machine/GPU combos!

PR #804, opened 2024-08-18
Remove Exceptions

bjornbytes:
TLDR; stop using `lua_error` for throwing exceptions, instead change code to use return values for errors.  No intended changes to Lua API (although filesystem functions return better error messages now).

Currently, when LÖVR needs to throw an error, it uses `lovrThrow` (or `lovrCheck` or `lovrAssert` which are convenience wrappers).  This calls `lua_error`, which immediately does a longjmp into Lua's pcall error handling code.  That has worked fine for a long time, but it's been causing some problems lately!  Can you spot the bug?

```c
Thing* lovrThingCreate(int count) {
  Thing* thing = malloc(sizeof(Thing));
  lovrCheck(count > 0, "Count is zero!");
  thing->count = count;
  return thing;
}
```

If count is zero here, we don't free `thing` and end up leaking memory!  This kind of code is *everywhere*.  Now, a lot of times this is fine.  If we throw an error, we're probably just going to show the error screen and quit soon.  Also most of these error checks are just things people run into while actively writing and testing Lua code.  So who really cares about these memory leaks?  Well, so far, nobody!

Here's another example:

```c
Buffer* lovrBufferCreate(BufferInfo* info) {
  Buffer* buffer = malloc(sizeof(Buffer));
  mtx_lock(&state.lock);
  buffer->handle = getBuffer(info);
  mtx_unlock(&state.lock);
  return buffer;
}
```

This one is way worse!  If `getBuffer` were to do anything that throws an exception, then our mutex doesn't get unlocked.  If we then try to render the error screen, and it happens to do anything that wants to use that lock, there will be a deadlock and LÖVR will just freeze up.  As more parts of LÖVR are getting multithreaded this has become more of a looming concern.

Also, although we don't have a C API yet, people would be pretty grumpy if LÖVR's C API required you to use setjmp/longjmp or just abort the program to handle errors.

To summarize: generally errors shouldn't leave the framework in an inconsistent state!

Solutions
---

In C++, we could use RAII to automatically destroy resources if an exception was thrown, but this is C so we can't use RAII (phew!).

Recently I tried adding a "defer" system where C code can register destructors to run in the event of `lovrThrow`.  It did work, but was too painful to use, because it required dedicated callback functions and a lot of extra noise.

The common way to do error handling in C is to return some kind of integer or enum describing whether the function succeeded or failed.  I tried this, but it didn't pan out.  It required Lua bindings to synthesize error messages from return codes, and there often wasn't enough context to do this very well.  errno-style systems have the same issue.

I settled on returning a succeed/fail boolean result (zero is always failure), with a GetLastError-style function that returns a detailed error message about what went wrong.  This lets C code unwind the stack and clean up after itself when an error happens, and doesn't require Lua bindings to synthesize error messages: they can just get the error string and throw that.

`lovrCheck` and `lovrAssert` still exist, but they set the error message and return zero when their condition is false.


bjornbytes:
I'm gonna merge this, there might be some bug fallout but it's an important step towards making stuff more robust and multithreadable

PR #806, opened 2024-08-29
Shape queries

bjornbytes:
Adds `Shape:containsPoint` and `Shape:raycast`

- `Shape:containsPoint(x, y, z)`
- `x, y, z, nx, ny, nz = Shape:raycast(x1, y1, z1, x2, y2, z2)`

All units are in world space.  The queries take into account the shape's offset and collider pose (if any).

The math is kinda gross because Jolt's shape queries are in the shape's local space, relative to its center of mass.  So we have to do lots of conversions between world space and center-of-mass space.

### DIFF
--- a/src/api/l_physics_shapes.c
+++ b/src/api/l_physics_shapes.c
@@ -338,6 +338,34 @@ static int l_lovrShapeGetAABB(lua_State* L) {
   return 6;
 }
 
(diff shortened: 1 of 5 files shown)


bjornbytes:
Wait... this isn't taking into account the collider poses yet

PR #824, opened 2024-12-19
lovr.headset.setBackground;

bjornbytes:
This refactors the new background layer types (cube/sphere) into a `lovr.headset.setBackground` function that can take an equirect or cubemap texture/images.  Now there is just 1 layer type (quads) and LayerType is removed.

### DIFF
--- a/src/api/l_headset.c
+++ b/src/api/l_headset.c
@@ -92,13 +92,6 @@ StringEntry lovrDeviceAxis[] = {
   { 0 }
 };
 
-StringEntry lovrLayerType[] = {
-  [LAYER_QUAD] = ENTRY("quad"),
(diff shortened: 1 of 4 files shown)


bjornbytes:
- stereo backgrounds aren't supported yet, but could be.
- formats are annoying.  you can't do `lovr.headset.setBackground('sky.hdr')` because that will load an `rgba32f` image, but right now layers must be `rgba8`.  Runtimes do support other texture formats for layers, but `rgba32f` isn't supported everywhere.  However, you CAN do `lovr.headset.setBackground(lovr.graphics.newTexture('sky.hdr'))` because this will do a texture blit, which supports format conversion.  Hopefully we can smooth this out in the future.

PR #826, opened 2024-12-27
Add getVertices to ConvexShape

xiejiangzhi:


bjornbytes:
I feel like the API here is too opinionated/specific.

- It uses a hardcoded data format.  So if someone wants something slightly different (smooth normals, packed normals, extra attributes), then they have to give up and rewrite it in Lua.
- The data it returns is already exposed by `:getPoint` and `:getFace`, so it's already possible to write this in Lua.
- There aren't really places in the existing API where we return a vertex blob unless you count `ModelData:getBlob`.

For now I'd recommend moving this code to Lua or keeping it on a fork.

I do still think the current API for the convex hull data is a little clunky.  Adding `:getPoints` may help, for people that want all the points in a table anyway.  Or it could return multiple returns like LÖVE's `PolygonShape:getPoints` does, but I'm not sure what the max number of points in the hull is, so it might be too many return values.

PR #828, opened 2024-12-31
Window Focus/Visible Events

bjornbytes:
- `lovr.focus` and `lovr.visible` are also called for the desktop window now.  They have a second parameter, `display`, indicating whether the event came from the `headset` or the `window`.
- Add `lovr.system.isWindowVisible` and `lovr.system.isWindowFocused`.

Minor TODO: simulator could emit visible and mounted events.

### DIFF
--- a/src/core/os_glfw.h
+++ b/src/core/os_glfw.h
@@ -131,7 +131,7 @@ uintptr_t os_get_xcb_window(void) {
 static struct {
   GLFWwindow* window;
   fn_quit* onQuitRequest;
-  fn_focus* onWindowVisible;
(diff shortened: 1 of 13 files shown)


jmiskovic:
:+1:  Works well

src/core/os_glfw.h
```diff
@@ -123,6 +131,7 @@ uintptr_t os_get_xcb_window(void) {
 static struct {
   GLFWwindow* window;
   fn_quit* onQuitRequest;
+  fn_focus* onWindowVisible;
```
bjornbytes: Small typo here

PR #838, opened 2025-02-06
Add a check for missing mesh in gltf

jmiskovic:
Provide user an error rather than crashing when trying to draw the model

### DIFF
--- a/src/modules/data/modelData_gltf.c
+++ b/src/modules/data/modelData_gltf.c
@@ -530,6 +530,11 @@ bool lovrModelDataInitGltf(ModelData** result, Blob* source, ModelDataIO* io) {
     }
   }
 
+  if (model->blobCount && !model->blobs) {
+    model->blobCount = 0;
+    ASSERT(false, "Mesh missing in model");
+  }
+
   // We only support a single root node, so if there is more than one root node in the scene then
   // we create a fake "super root" node and add all the scene's root nodes as its children.
   if (info.sceneCount > 0 && scenes[rootScene].nodeCount > 1) {


jmiskovic:
Reported [here](https://matrix.to/#/!wvCsUVzWCWEZBlBrUY:lovr.org/$JFtJz8UA8pAGYx-zxcm4ukm_BT5JOW58is2VGkGavOI)

src/modules/data/modelData_gltf.c
```diff
@@ -530,6 +530,11 @@ bool lovrModelDataInitGltf(ModelData** result, Blob* source, ModelDataIO* io) {
     }
   }
 
+  if (model->blobCount && !model->blobs) {
```
bjornbytes, line 533: This broke all newModel calls for me, since `model->blobs` is always NULL here.  For me, the model is importing fine but there's a crash during rendering.  The vertex buffer was NULL during the vertex animation.  I pushed 49e90aa39811080631878f427d6991b23928165e to just skip vertex animation when there isn't a vertex buffer.

jmiskovic: Ah sorry. I first tried to assert on `model->primitiveCount` but then the cleanup would crash when releasing blobs, because I did it prematurely. Your method is cleaner.

bjornbytes: Ah I see.  I think there is still an issue here with the error/cleanup code.  Since `blobCount`/`imageCount` are set before `blobs`/`images` are assigned, if there is an error then `lovrModelDataDestroy` is going to try to free non-existent blobs/images.  I'll see if I can improve it.

bjornbytes: Fixed up in 949b0bae9e3d0fc9f3631c31ef86b32c0a9614fe, just need to ensure arrays exist before releasing their contents

PR #854, opened 2025-03-09
Fix OBJ texture Y direction inversion (#841)

### LINKED ISSUE #841: OBJ Texture Coordinate Y Direction
Currently when an OBJ is imported, its materials [get a `uvScale` of `(1, -1)`](https://github.com/bjornbytes/lovr/blob/dev/src/modules/data/modelData_obj.c#L57) to account for a difference between LÖVR's texture direction (y down) and OBJ's texture direction (seemingly Y up, but I can't find this specified anywhere).

However, this can be annoying if you import your OBJ models with `{ materials = false }`, since then you lose out on that Y flip and need to flip all your images or create a material with flipped UVs.  The flip could instead be baked into the vertex data, so that the texture coordinates are always correct.

PhilipeSousa:
## **Fix OBJ Texture Y Direction Inversion**  

- Fixes the Y texture coordinate inversion for OBJ models.  
- Closes #841.  

## **Reproducing the Issue**  
The issue was reproduced by importing an OBJ model **without materials** using `{ materials = false }`. This caused the texture to appear **inverted in the Y direction**:  

### Importing without mat
Importing without mat

### Fix Implementation
The fix was applied directly in the `lovrModelDataInitObj(...)` function, ensuring the Y coordinate inversion is handled at the vertex level, rather than relying on material properties.

#### After changes imported without materials again
Screenshot from 2025-03-08 21-37-33

#### After changes imported with materials
Screenshot from 2025-03-08 21-38-36

### DIFF
--- a/src/modules/data/modelData_obj.c
+++ b/src/modules/data/modelData_obj.c
@@ -54,7 +54,7 @@ static bool parseMtl(char* path, char* base, ModelDataIO* io, arr_image_t* image
         .color = { 1.f, 1.f, 1.f, 1.f },
         .glow = { 0.f, 0.f, 0.f, 1.f },
         .uvShift = { 0.f, 1.f },
-        .uvScale = { 1.f, -1.f },
+        .uvScale = { 1.f, 1.f },
         .metalness = 1.f,
         .roughness = 1.f,
         .clearcoat = 0.f,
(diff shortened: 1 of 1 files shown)


bjornbytes:
Looks good.  Thanks for the PR!

PR #856, opened 2025-03-13
Fix building with `-DLOVE_SYSTEM_LUA`

Frityet:
Previously the libdirs were not properly set causing the actual lua libs to be unable to be found at link time


Frityet:
accidental regression as I based this off the 0.18.0 cmakelists.txt, will fix

CMakeLists.txt
```diff
@@ -124,12 +125,14 @@ if(LOVR_USE_LUAJIT AND NOT EMSCRIPTEN)
       target_compile_definitions(buildvm PRIVATE _CRT_SECURE_NO_WARNINGS)
     endif()
     set(LOVR_LUA_INCLUDE deps/luajit/src ${CMAKE_BINARY_DIR}/luajit)
+    set(LOVR_LUA_LIBDIR deps/luajit/src)
```
bjornbytes: I don't understand why this path (and `deps/lua` for vanilla Lua) are added to the library search path, since there aren't any libraries output in those paths.

bjornbytes: I guess the intent is so that you can compile the version in `deps` manually and have that library found?

bjornbytes: It doesn't seem like it would work unless `BEFORE` is added to the `target_link_directories` call though.

Frityet: > I don't understand why this path (and `deps/lua` for vanilla Lua) are added to the library search path, since there aren't any libraries output in those paths.

oh that was just for consistency

Frityet: @bjornbytes fixed

CMakeLists.txt
```diff
@@ -338,6 +341,11 @@ target_include_directories(lovr PRIVATE
   ${CMAKE_CURRENT_SOURCE_DIR}/etc
 )
 
+target_link_directories(lovr PRIVATE
+  ${LOVR_LUA_LIBDIR}
+  ${LOVR_GLFW_LIBDIR}
```
bjornbytes, line 346: Maybe also add OpenXR, for completeness.  It may be more convenient to maintain a single list of all the library dirs as well, like `LOVR_LIBRARY_DIRS`.

CMakeLists.txt
```diff
@@ -744,7 +753,7 @@ elseif(APPLE)
   set_target_properties(lovr PROPERTIES
     MACOSX_RPATH TRUE
     BUILD_WITH_INSTALL_RPATH TRUE
-    INSTALL_RPATH "@executable_path"
+    INSTALL_RPATH "@executable_path;@executable_path/../lib"
```
bjornbytes, line 756: Why was this added to rpath?

Frityet: if `lovr` is installed to `/usr/local/bin` then libraries will be in `/usr/local/lib`, etc

bjornbytes: Installing to a system-wide location like this isn't supported right now

Frityet: it does work though

PR #860, opened 2025-03-22
Add support for second set of UVs;

bjornbytes:
This is an attempt to add support for importing Models with a second set of UVs.

- There is a new vertex attribute available in vertex shaders named `VertexUV2`, and a new fragment shader input named `UV2` which can be used to sample textures.
- All UVs for models are converted from 32 bit floats to unsigned shorts.  This allows us to squeeze in a second set of UVs without increasing the size of model vertices, but it has a few consequences:
  - Precision loss.  Instead of 4 billion unique UV values, there are only 65535.  This is still a lot of precision though, any difference shouldn't be noticeable.
  - UVs can only be between 0 and 1 now, instead of any number.  This could break some models, although the vast majority of models use UVs between 0 and 1.  You can still use the UV offset and UV scale feature to get larger UV ranges.
    - LÖVR could try to warn you if the glTF says it has UVs outside of the range, or even normalize the UVs to a bigger range and apply a UV scale (trading off even more precision to extend the UV range).
- If the second set of UVs is missing, it uses data from the first set of UVs (??? this is kinda weird, maybe take this out).
  - It does mean that shaders could use UV2 and still work when the model only has 1 set of UVs.  This could be cool for `OcclusionTexture`, which is often used for lightmaps with the second set of UVs.  LÖVR could use UV2 when sampling from the occlusion texture in any builtin shader functions and this wouldn't break models that only have 1 set of UVs.
- glTF has this feature where each material texture can indicate which set of UVs it wants to use.  Not many engines support this and this PR doesn't add any support for this.


bjornbytes:
I think dropping support for UVs outside the 0-1 range is too egregious.  Some ideas:

- Just make them 32 bit and go up to 40 bytes.
- Support multiple vertex formats, so a model can either use 32 bit UVs or 16 bit UVs.  Models without UV2 would use 32 bit UVs, so you at least get full precision if you aren't using lightmaps.
- Make UVs `sn16x2` (-1 to 1) and bake in a 2x factor into the UV scale, so we can support UVs from -2 to 2.  Sacrifices even more precision but it should be okay up to 16K textures.

bjornbytes:
Closing this, it got too crusty after the model rework.  Still need to figure out how to set up the attribute format for multiple UVs.

PR #865, opened 2025-04-12
cubic FilterMode

### LINKED ISSUE #864: Hardware bicubic texture filtering
Hardware bicubic filtering is available [on Quest devices](https://developers.meta.com/horizon/documentation/native/android/mobile-openxr-bicubic/) via the [VK_IMG_filter_cubic](https://registry.khronos.org/vulkan/specs/latest/man/html/VK_IMG_filter_cubic.html) extension. It may be available on other devices, but I'm not sure.

Bicubic texture filtering can appear visually superior to bilinear filtering, especially for cases like lightmaps, where the texels can appear quite large. It can be implemented in shaders, but the hardware version should perform better.

This could be exposed to LÖVR by adding a `"cubic"` `FilterMode`. Since not all hardware will support it, if the game attempts to use that mode when unsupported, LÖVR could either silently downgrade to `"linear"`, or throw an error. Additionally, it should be possible to query at runtime whether hardware bicubic filtering is possible (`lovr.graphics.getFeatures`?), in case a game wishes to use a shader-based fallback.

Image

bjornbytes:
Fixes #864

### DIFF
--- a/src/api/l_graphics.c
+++ b/src/api/l_graphics.c
@@ -117,6 +117,7 @@ StringEntry lovrDataType[] = {
 StringEntry lovrFilterMode[] = {
   [FILTER_NEAREST] = ENTRY("nearest"),
   [FILTER_LINEAR] = ENTRY("linear"),
+  [FILTER_CUBIC] = ENTRY("cubic"),
   { 0 }
 };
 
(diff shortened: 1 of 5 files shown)


bjornbytes:
Hm, I'm not liking the results so far, it adds a sharpening effect instead of smoothing everything out.  I wonder if this is meant to be applied to the whole VR texture as described [here](https://developers.meta.com/horizon/documentation/native/android/mobile-openxr-bicubic/)

PR #871, opened 2025-06-15
OpenXR Render Models

bjornbytes:
Adds support for `XR_EXT_render_model` and `XR_EXT_interaction_render_model`.

Changes:

- Instead of creating models from Devices, there are model "keys".
  - This is because it's possible to have multiple models for a single device (hand plus controller), or models that do not correspond to a device at all.
  - Use `lovr.headset.getModelKeys` to return a list of model keys.
  - These are opaque `lightuserdata` identifiers.
  - Use the key to create a model: `lovr.headset.newModel(key)`
  - You can still create models from Devices, for backwards compatibility.
- There is a new `modelschanged` event that is fired whenever models are changed.
  - You should load models in here, potentially diffing the keys to figure out which ones need to be loaded and which can be removed.
- You can query the pose to draw a model at with the following new variants:
  - `lovr.headset.isTracked(model)` --> if false, don't draw the model
  - `lovr.headset.getPosition(model)`
  - `lovr.headset.getOrientation(model)`
  - `lovr.headset.getPose(model)`
  - `lovr.headset.getDirection(model)`
- The `animated` flag in `lovr.headset.newModel` was removed.  All models are animated now.
- Removed support for the old Microsoft extension, `XR_MSFT_controller_model`.


bjornbytes:
Confirmed that Quest hand tracking meshes work on both the old and new APIs.

PR #872, opened 2025-06-19
Headset Rework

### LINKED ISSUE #279: Hard-coded keyboard/mouse input
The keys WASDQE, Arrow keys, mouse movement and mouse click are presently bound hard-coded to what is apparently a headset offset value. 

Input should never be hard-coded. These also don't generate keyboard callbacks, so they can't be overridden by lovr.keyboard. 

Suggested:
Merge lovr.keyboard as a base library and implement these controls as lovr-layer callbacks instead

Repro steps:
1. Utilize a specular light source to illuminate a model (passing camera/view position to the shader)
2. Set mirror to:
lovr.mirror()
  lovr.graphics.clear()
  lovr.draw()
end
3. Start app and use wasdqe/arrows and observe light source moves despite there being "no input code"

Further testing:
Tying wasd or arrows to a transform offset creates all sorts of lighting glitches when changing position and rotation of the viewpoint.

bjornbytes:
Headset rework:

- Headset drivers no longer exist.  You're always going to get OpenXR on desktop (or WebXR on web, but I removed it for now)
- Removed the `HeadsetInterface` struct.  The headset module is like a regular module now, e.g. `lovrHeadsetGetPose`.
- The simulator is moved to Lua:
  - `lovr.simulate(dt)` callback is used to update the simulator state.
  - New functions to update simulated headset data:
    - `lovr.headset.get/setSimulatedPose/Position/Orientation`
    - `lovr.headset.get/setSimulatedButton`

TODO

- [x] Need to add support for the simulated hand pose
- [x] Should change `--simulator` to set `t.headset.start` to false
- [x] Actually need to remove drivers from default conf
- [x] Could make headset more robust when not active (some methods will crash right now)
- [x] Move headset event processing after `xrWaitFrame` to improve event latency
- [x] Fix awkward first frame due to default mouse position being (0,0)


bjornbytes:
AAAAAA

PR #878, opened 2025-07-04
Add support for terrain holes

jmiskovic:
Jolt's heightmap shape supports a special height value FLT_MAX, which signifies the cNoCollisionValue gets handled as a hole in the terrain.

The FLT_MAX is not readily available in Lua. Instead, the heightfield callback can return a `nil` value for x,z coordinates to specify that the heightmap area should be ignored for collisions.

The downside: previously if you unintentionally leave the callback fn empty you'd get a helpful error message. Now it will create a void terrain, which to me makes sense.

### DIFF
--- a/src/api/l_physics_shapes.c
+++ b/src/api/l_physics_shapes.c
@@ -150,8 +150,14 @@ Shape* luax_newterrainshape(lua_State* L, int index) {
       lua_pushnumber(L, x);
       lua_pushnumber(L, z);
       lua_call(L, 2, 1);
-      luax_check(L, lua_type(L, -1) == LUA_TNUMBER, "Expected TerrainShape callback to return a number");
-      vertices[i] = luax_tofloat(L, -1);
+      int height_type = lua_type(L, -1);
+      if (height_type == LUA_TNUMBER) {
+        vertices[i] = luax_tofloat(L, -1);
+      } else if (height_type == LUA_TNIL) {
+        vertices[i] = FLT_MAX;
+      } else {
+        luax_check(L, 0, "Expected TerrainShape callback to return a number or nil");
+      }
       lua_pop(L, 1);
     }
     TerrainShape* shape = lovrTerrainShapeCreate(vertices, n, scaleXZ, 1.f);


jmiskovic:
```Lua
local phywire = require'phywire'

local terrain = world:newTerrainCollider(25, function(x, z)
  if lovr.math.noise(x, z) > 0.8 then
    return nil
  else
    return math.sin(x)
  end
end)


function lovr.draw(pass)
  phywire.xray(pass, world)
end
```
image

bjornbytes:
Would `math.huge` be a better sentinel value than `nil`?

bjornbytes:
It matches Jolt better, but makes less sense, so I think nil is better.

jmiskovic:
That was my thought process exactly :smiley: 

The `-math.huge` would serve well here exept it is not as easy to remember as `nil`.

DonaldHays:
@bjornbytes might actually be onto something with the `math.huge` sentinel.

It depends on the use case, I think. In the provided sample, I think `nil` makes sense and is more idiomatic to Lua. If my holes are procedural or are stored separately from the height map data, `nil` is good.

But if I wanted to store all my height map data, holes inclusive, in a single blob of data, I could use an array of floats, with `INFINITY` for holes, which I could then just feed blindly into the collider API. Otherwise I would have to pick my own sentinel value (which, honestly, might end up being `INFINITY` anyway), detect it as `math.huge` in the collider API when reading the array, then convert it to `nil`.

bjornbytes:
That's a good point.  It doesn't hurt to support both.  Pushed it

jmiskovic:
The remaining input variant is the heightmap image. I don't know of an elegant way to support holes there. The workaround:

```Lua
local NO_COLLIDE = 3.4028234663852886e+38

local img = lovr.data.newImage(200, 200, 'r32f')
for x = 0, img:getWidth() - 1 do
  for z = 0, img:getHeight() - 1 do
    local h
    if lovr.math.noise(x * 0.02, z * 0.02) > 0.8 then
      h = NO_COLLIDE
    else
      h = math.sin(x * 0.2)
    end
    img:setPixel(x, z, h)
  end
end

local terrain = world:newTerrainCollider(TERRAIN_SIZE, img)
```

PR #879, opened 2025-07-05
Fix simulator reload while mouse button is down

jmiskovic:
If the left mouse button is held while lovr `--watch` is reloading it results in an error:

```
Error:

boot.lua:229: attempt to perform arithmetic on local 'lastX' (a nil value)
```

The fix makes sure to init all the `nil` variables.

Tested with:
```bash
# Changes main.lua to have different number of spaces in the last line every second
#!/bin/bash

FILE="main.lua"

while true; do
    SPACES=$(sed -n '$s/ *$//; s/.*//p' "$FILE" | wc -c)
    NEW_SPACES=$(( (SPACES + 1) % 10 ))
    sed -i -E "s/^(.*[^ ])[ ]*$/\1 $(printf '%*s' "$NEW_SPACES")/" "$FILE"
    sleep 1
done
```

### DIFF
--- a/etc/boot.lua
+++ b/etc/boot.lua
@@ -210,6 +210,7 @@ function lovr.simulate(dt)
 
   if not pitch or not yaw then
     pitch, yaw = quat(lovr.headset.getOrientation()):getEuler()
+    mouseX, mouseY = lovr.system.getMousePosition()
   end
 
   local movespeed = 3


jmiskovic:
I've actually set out to fix another issue, where the camera rotated slightly on reload. It seemed to be related to headset angular velocity or leftover mouse delta. But I can no longer reproduce this for some reason, with or without this fix.

bjornbytes:
I was seeing something like that too, briefly.  I thought it was related to some issue with euler angles, but I had trouble reproducing it.

PR #882, opened 2025-07-26
Add lengthFromT and tFromLength to Curve

jmiskovic:


jmiskovic:
```Lua
local curve = lovr.math.newCurve({
  0, 0, -4,
  4, 0, -5,
  -4, 0, -5,
  2, 0, 0,
  0, 0, -7,
})

local function getScreenToWorldTransform(pass)
  local w, h = pass:getDimensions()
  local clip_from_screen = mat4(-1, -1, 0):scale(2 / w, 2 / h, 1)
  local view_pose = mat4(pass:getViewPose(1))
  local view_proj = pass:getProjection(1, mat4())
  -- m.is_orthographic = view_proj[16] == 1
  local world_from_screen = view_pose:mul(view_proj:invert()):mul(clip_from_screen)
  return world_from_screen
end


local dragged_p = nil

local function drawDragControlPoints(pass)
  local pose = mat4():rotate(-math.pi / 2, 1,0,0)
  local scale = 5
  local x, y = lovr.system.getMousePosition()
  local world_from_screen = getScreenToWorldTransform(pass)
  local ray_origin = vec3(world_from_screen:mul(x, y, 1))
  local ray_target = vec3(world_from_screen:mul(x, y, 0.001))
  local ray_direction = (ray_target - ray_origin):normalize()
  -- intersect the ray onto panel plane and see if it lands within panel
  local plane_direction = quat(pose):direction()
  local dot = ray_direction:dot(plane_direction)
  local hit_spot
  if math.abs(dot) > 1e-5 then
    local plane_pos = vec3(pose)
    local ray_length = (plane_pos - ray_origin):dot(plane_direction) / dot
    hit_spot = ray_origin + ray_direction * ray_length
  else
    return
  end
  local dragging = lovr.system.isMouseDown(2)
  local threshold = 0.3
  for i = 1, curve:getPointCount() do
    local p = vec3(curve:getPoint(i))
    pass:setColor(1, 1, 0.5)
    pass:sphere(p, 0.04,  0, 0,1,0,  11,11)
    if lovr.system.wasMousePressed(2) then
      if p:distance(hit_spot) < threshold then
        dragged_p = i
        break
      end
    end
  end
  if lovr.system.wasMouseReleased(2) then
    dragged_p = nil
  end
  if dragged_p then
    curve:setPoint(dragged_p, hit_spot)
  end
end


local function drawCurve(pass, curve, N)
  N = N or 40
  local p1 = vec3(curve:evaluate(0))
  local p2
  pass:setColor(0.8, 0.8, 0.3)
  for i = 1, N do
    p2 = vec3(curve:evaluate(i / N))
    pass:capsule(p1 + vec3(0, 0, 0), p2 + vec3(0, 0, 0), 0.005, 11)
    p1 = p2
  end
end


function lovr.draw(pass)
  drawDragControlPoints(pass, curve)
  drawCurve(pass, curve)
  pass:setColor(0.8, 0.8, 0.8)
  -- show total length
  local l = curve:lengthFromT(1, 64) -- 64 intervals are enough for 5 digits
  pass:text(string.format('length: %.5f', l), 0, 0.5, -2, 0.1)
  pass:setColor(0.2, 0.6, 0.2)
  -- subdivide curve by every 0.1
  for x = 0, l,  0.1 do
    local t = curve:tFromLength(x)  -- visuals don't seem to improve above 16 iterations, the default
    local p = vec3(curve:evaluate(t))
    pass:box(p, 0.03)
  end
end
```

bjornbytes:
Thanks for implementing this!

Thinking about naming, so far I came up with `Curve:getLength(t)` instead of lengthFromT (t can default to 1.0 and work as a way to get the length of the whole curve) and `t = Curve:step(distance)` instead of tFromLength (kinda matches `:evaluate` and `:render`).  Or possibly `Curve:getParameter(distance)` instead of `:step`.

src/modules/math/math.c
```diff
@@ -218,6 +218,74 @@ void lovrCurveRemovePoint(Curve* curve, size_t index) {
...
+    points[i * 4 + 2] = (float)(n - 1) * (P[(i + 1) * 4 + 2] - P[i * 4 + 2]);
+    points[i * 4 + 3] = (float)(n - 1) * (P[(i + 1) * 4 + 3] - P[i * 4 + 3]);
+  }
+  evaluate(points, n - 1, t, dp);
+  lovrFree(points);
+}
+
+float lovrCurveLengthFromT(Curve* curve, float t, int iterations) {
+  lovrCheck(curve->points.length >= 8, "Need at least 2 points to compute curve length");
+  lovrCheck(t >= 0.f && t <= 1.f, "Parameter t must be within [0, 1]");
```
bjornbytes: `lovrCheck` doesn't throw an error anymore but instead returns zero (I did this to make it easier to clean up resources when throwing an error).

It might be more straightforward to clamp t to 0..1 instead of doing error checking (which would require modifying the function signature in an annoying way)

jmiskovic:
I've removed the checks and clamped inputs, tested it and it works with empty curve.

I like the new names. `step` was shorter than the other so I added that.

The `curve:getLength()` variant added, t defaults to t=1.

bjornbytes:
Cool, merging!

PR #888, opened 2025-08-14
Luau

### LINKED ISSUE #870: Luau
The `luau` branch has a working version of LÖVR that uses [Luau](https://luau-lang.org) instead of LuaJIT.  You can run stuff like this:

```luau
function lovr.load()
  local a: number = 1_000
  a += 1
  print(`a is {a}`)
end
```

This issue is meant to track/discuss the status of Luau in LÖVR.  Luau has many differences from both Lua and LuaJIT, and some things are missing and/or broken.  There are also interesting things about Luau that remain to be explored, like experimenting with its vectors and integrating with tooling for typechecking.

Known issues:

- [x] `require` doesn't exist
- [x] `os.getenv` doesn't exist (decided to leave as-is)
- [x] `collectgarbage` doesn't exist (should polyfill, see [this](https://github.com/luau-lang/luau/blob/68cdcc4a3a5f3ed23186c4f7f6b8a5aacf835bee/CLI/src/Repl.cpp#L117), remove hack from `test/conf.lua` once finished)
- [x] Plugins don't work (lovr-http and lua-enet fork could be updated to support luau?)

Pending investigations:

- [x] Vectors (integrating into lovr.math's vector API, performance, 3 vs. 4 component vectors)
- [x] Does it work on WebAssembly?
- [x] Test on Android
- [x] Should LÖVR also support .luau file extensions?  (Added)
- [x] `__namecall` optimization
- [x] Imports/sandboxing
- [x] Native code generation
- [x] Should we be using `lua_newuserdatadtor` and/or tagged userdata?

bjornbytes:
This adds support for [Luau](https://luau.org) as an alternative to Lua and LuaJIT.  It also reworks the vector API.

Luau
---

First: LuaJIT is still going to be the default!  This is just an opportunity to let people experiment with Luau + LÖVR, see how it goes, and provide feedback.

#### Why add support for Luau?

- Luau's vectors seemed particularly compelling, since LÖVR has struggled with exposing an efficient vector API.  Luau's vectors are primitive types built in to the language.  They're about as expensive as numbers and don't create any garbage.
- There are some other interesting features of Luau, like type checking, syntax conveniences, and predictable performance.

#### Compatibility

Luau isn't 100% compatible with Lua 5.1.  However, so far in testing, most simple projects will work without any changes.  The biggest sources of incompatibility are probably the lack of `ffi`, no `io` library, and limited `os`/`debug`/`package` libraries.  See [this page](https://luau.org/compatibility) for more info on compatibility.

Luau is compatible with LÖVR's plugin system, but since Luau isn't compatible with Lua's C API, not all plugins will work with Luau out of the box.  `lua-enet` and `lovr-http` worked without any changes, FWIW.

When using Luau, `require` will first check for files with the `.luau` extension, and afterwards fall back to `.lua`.

#### Fork

This branch uses a fork of Luau with the following changes:

- Adds a 16-bit `quaternion` type, in addition to the `vector` type.
- Modifies the `vector` standard library slightly.
- Adds `require` and `collectgarbage` functions.
- Modifies Luau's C API to more closely match Lua's C API.

#### Performance

The branch hasn't implemented support for some Luau-specific optimizations like `__namecall`, tagged userdata, or some fastcall methods for vector/quaternion.  Also, some optimizations are disabled for now, like "imports" (inlined table access).  There are definitely some opportunities to improve performance.

Vectors
---

Luau's vectors have some constraints that meant LÖVR's existing vector API couldn't be implemented as-is:

- They are immutable value types, like strings.  You can't "change" a string in Lua, you can only assign a variable to a new string.  Vectors are the same way in Luau, you can't do `position:set(1, 2, 3)` or even `position.x = 4`.
- They can only have 3 components (12 bytes).  There is a way to extend the VM to support 4-component vectors but it has performance costs.  (The quaternion type we added uses 16 bit precision, squeezing quaternions into 8 bytes).
- There is only one vector metatable.  It isn't possible to have separate `vec2`/`vec3`/`vec4` types without adding each as a new primitive type, but this requires a lot of effort and complicates the VM.

Separately, there have been long-standing requests for an immutable vector API (#169).  Also, the existing userdata vectors can not get JIT-compiled by LuaJIT, and there is some overhead due to method calls having to go through C.

With all that in mind, we took this as an opportunity to rework LÖVR's vector API.  Here's what it looks like:

```lua
-- vector
vector(x, y, z) --> synonym for vector.pack
vector.pack(x, y, z)
vector.unpack(v)
vector.length(v)
vector.normalize(v)
vector.distance(a, b)
vector.cross(a, b)
vector.dot(a, b)
vector.angle(a, b)
vector.lerp(a, b, t)

-- quaternion
quaternion(angle, ax, ay, az) --> synonym for quaternion.angleaxis
quaternion.pack(x, y, z, w)
quaternion.unpack(q) --> returns raw components
quaternion.conjugate(q)
quaternion.angleaxis(angle, ax, ay, az)
quaternion.toangleaxis(q) --> returns angle, ax, ay, az
quaternion.euler(pitch, yaw, roll)
quaternion.toeuler(q) --> returns pitch, yaw, roll
quaternion.between(v1, v2)
quaternion.lookdir(dir, [up])
quaternion.direction(q)
quaternion.slerp(q1, q2, t)
```

This is implemented using native types on Luau and tables on other Lua implementations.

`Mat4` was turned into a regular LÖVR userdata object (similar to e.g. `Curve`), with the same API.

Details about compatibility:

- `vec2` and `vec4` were removed.
- `:add`, `:sub`, `:mul`, `:div`, and `:set` are removed, because Luau vectors do not support mutation.
- Operations like `:cross`, `:normalize`, and `:lerp` return new vectors.
- There are no temporary vectors anymore.  `lovr.math.drain` still exists for compatibility, but does nothing.
- Swizzles that return multiple components (like `.xyz`) were removed.
- The `#` operator was removed, because in Lua 5.1 tables do not support `__len`.
- Vector constants like `vector.up`, `vector.zero`, `quaternion.identity` still exist.
- Existing constructors (like `vec3`, `Vec3`, `lovr.math.vec3`, `lovr.math.newVec3`) all point to `vector` (or `quaternion`).
- Any method that takes vectors can take any table with x/y/z/w or 1/2/3/4 keys (to support table vectors, but you can give them any table, which is nice).
- Vectors can be passed to buffer fields.  If the field is a `vec4`, its `w` will get set to 1.
- `Mat4:mul` takes an optional `w` component after a vector, which can be used for vec4 multiplication.
- Quaternions are always normalized, so there isn't a `quaternion.length` or `quaternion.normalize`.
- Luau doesn't support adding/subtracting vectors and numbers
- Luau vectors can be sent to channels.

[Here](https://github.com/bjornbytes/lovr-docs/compare/new-vectors) is a branch that ports the docs examples to the new vector API.

In my experience the most painful part of porting is the lack of temporary `Mat4` objects.  Code that makes heavy use of `mat4()` still works, but it doesn't feel great to create tons of new `Mat4` instances.  It's often possible to reduce mat4 use by using the pass transform stack or just using vectors and quaternions, but not always.  Some sort of pooling solution might be needed.


bjornbytes:
Benchmarking [line mesh generation](https://github.com/xiejiangzhi/line_mesh/blob/main/benchmark.lua)

```
lua version     Luau (table, -O1)
p100 x10000, cost: 18165.0605
p100 x10000, cost: 18119.7520
p100 x10000, cost: 19704.2842
---
lua version     Luau (vector, -O1)
p100 x10000, cost: 2718.8388
p100 x10000, cost: 2744.1398
p100 x10000, cost: 2736.9801
---
lua version     Lua 5.1  | LuaJIT 2.1.1759471281        jit on
p100 x10000, cost: 876.9720
p100 x10000, cost: 861.4030
p100 x10000, cost: 884.3280
---
lua version     Lua 5.1  | LuaJIT 2.1.1759471281        jit off
p100 x10000, cost: 9274.8150
p100 x10000, cost: 9158.4040
p100 x10000, cost: 9174.0110
```

Note: LuaJIT tests are using table vectors, not the current lovr.math vectors.

bjornbytes:
Mini vector migration guide:

```lua
v = vec3() --> v = vector()
v = Vec3() --> v = vector()
v:set(x, y, z) --> v = vector(x, y, z)
v:set(u) --> v = u
v:add(u) --> v = v + u, same for sub/mul/div
#v --> v:length()
v = vec3(v):cross(u) --> v = v:cross(u), same for normalize/lerp
v:transform(m) --> m * v
v.x = 4 --> v = vector(4, v.y, v.z)
v.xy = v.yx --> v = vector(v.y, v.x, v.z)
v.r --> v.x

q = quat() --> q = quaternion()
q = Quat() --> q = quaternion()
q:set(angle, ax, ay, az) --> q = quaternion(angle, ax, ay, az)
q:set(r) --> q = r
q:set(v) --> q = quaternion.lookdir(v)
q:set(m) --> q = quaternion(m:getOrientation())
q:set(v1, v2) --> q = quaternion.between(v1, v2)
q:set(x, y, z, w, true) --> quaternion.pack(x, y, z, w)
q:unpack(true) --> q:unpack()
q:unpack() --> q:toangleaxis()
q:getEuler() --> q:toeuler()
q:setEuler(x, y, z) --> quaternion.euler(x, y, z)
q:mul(r) --> q = q * r
q:length() --> 1 (quaternions are always normalized)
q:normalize() --> q
q = quat(r):conjugate() --> q = r:conjugate()
q:slerp(r, t) --> q = q:slerp(r, t)
quat(mat4():target(pos, pos + dir, up)) --> quaternion.lookdir(dir, up)
```

bjornbytes:
Implementation of old vectors, to assist in porting

https://gist.github.com/bjornbytes/62a54c6e200bd111068c5626433a7ad5

PR #898, opened 2025-10-15
Quad Views

bjornbytes:
Adds support for quad views and mono view configurations in OpenXR, instead of assuming stereo.

Quad views render 4 views.  The first 2 are low resolution stereo views.  Views 3 and 4 are inset into the first 2, and have a smaller field of view (so a higher pixel density).  It lets you have a higher PPD (either in the center of the display, or where the user is looking), but keep the periphery low resolution, reducing the total number of pixels shaded.

Adding support for quad views made it trivial to add support for mono views, so that is now supported as well.

I cheated a little bit here and require all the views to have the same resolution, which usually isn't the case in practice.  The swapchain layers use the largest recommended size across all the views, but clamp it to the smallest max view size.  This allows us to continue to use multiview to render all the views in a single pass.  This could be improved in the future by using a per-view viewport, using either `VK_EXT_shader_viewport_index_layer` or `VK_QCOM_multiview_per_view_viewports`, but it's tricky since these aren't guaranteed to be supported (would need to patch SPIR-V bytecode I think).

Aside from the per-view viewport, another improvement would be a better Lua API for picking which view configuration you want.

### DIFF
--- a/src/modules/headset/headset.c
+++ b/src/modules/headset/headset.c
@@ -626,9 +626,9 @@ bool lovrHeadsetConnect(void) {
         break;
       }
     }
-
-    lovrAssertGoto(fail, i + 1 < COUNTOF(supportedViewConfigurations), "No supported view configuration available");
   }
+
+  lovrAssertGoto(fail, state.viewConfiguration, "No supported view configuration available");
 
   XrViewConfigurationView views[4] = {
     [0].type = XR_TYPE_VIEW_CONFIGURATION_VIEW,
--- a/src/modules/headset/headset.c
+++ b/src/modules/headset/headset.c
@@ -154,11 +154,12 @@ enum {
 };
 
 enum {
-  STEREO = (1 << 0),
-  DEPTH = (1 << 1),
-  CUBE = (1 << 2),
-  STATIC = (1 << 3),
-  FOVEATED = (1 << 4)
+  VIEW = (1 << 0),
+  STEREO = (1 << 1),
+  DEPTH = (1 << 2),
+  CUBE = (1 << 3),
+  STATIC = (1 << 4),
+  FOVEATED = (1 << 5)
 };
 
 typedef struct {
@@ -249,6 +250,8 @@ static struct {
   Simulator simulator;
   XrInstance instance;
   XrSystemId system;
+  XrViewConfigurationType viewConfiguration;
+  uint32_t viewCount;
   XrSession session;
   XrSessionState sessionState;
   XrSpace referenceSpace;
@@ -262,8 +265,8 @@ static struct {
   Pass* pass;
   Swapchain swapchains[3];
   XrCompositionLayerProjection layer;
-  XrCompositionLayerProjectionView layerViews[2];
-  XrCompositionLayerDepthInfoKHR depthInfo[2];
+  XrCompositionLayerProjectionView layerViews[4];
+  XrCompositionLayerDepthInfoKHR depthInfo[4];
   XrCompositionLayerPassthroughFB passthroughLayer;
   union {
     XrCompositionLayerBaseHeader header;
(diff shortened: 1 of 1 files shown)


src/modules/headset/headset.c
```diff
@@ -604,32 +609,57 @@ bool lovrHeadsetConnect(void) {
...
+  XrViewConfigurationType viewConfigurations[4];
+  XRG(xrEnumerateViewConfigurations(state.instance, state.system, 4, &viewConfigurationCount, viewConfigurations), "xrEnumerateViewConfigurations", fail);
 
-  if (viewCount != 2) {
-    return lovrSetError("Headset view count must be 2");
+  for (uint32_t i = 0; i < COUNTOF(supportedViewConfigurations); i++) {
+    for (uint32_t j = 0; j < viewConfigurationCount; j++) {
+      if (viewConfigurations[j] == supportedViewConfigurations[i]) {
+        state.viewConfiguration = supportedViewConfigurations[i];
+        break;
```
bjornbytes, line 626: Ah, this probably needs to break out of both loops.  First loop needs `!state.viewConfiguration && i < COUNTOF(supportedViewConfigurations);` as the condition

gareth-morgan-nv: Yeah I think as its written it will always choose the last entry in `supportedViewConfigurations` that exists in `viewConfigurations`.  When I tested this it would always choose `XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO` so you would not get Quad Views.

gareth-morgan-nv (commented):
I think the logic for choosing ViewConfiguration should look something like this, where you have a requested view configuration (e.g. XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO_WITH_FOVEATED_INSET) and then try and find that. Then fallback to whatever is available if its not found in the available configurations:
https://gist.github.com/gareth-morgan-nv/581b0a5fa55dd71290a6ed05b472155e

src/modules/headset/headset.c
bjornbytes: Fixed!

src/modules/headset/headset.c
```diff
@@ -604,32 +609,57 @@ bool lovrHeadsetConnect(void) {
...
-  XrViewConfigurationView views[2] = { [0].type = XR_TYPE_VIEW_CONFIGURATION_VIEW, [1].type = XR_TYPE_VIEW_CONFIGURATION_VIEW };
-  XRG(xrEnumerateViewConfigurationViews(state.instance, state.system, XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO, 0, &viewCount, NULL), "xrEnumerateViewConfigurationViews", fail);
-  XRG(xrEnumerateViewConfigurationViews(state.instance, state.system, XR_VIEW_CONFIGURATION_TYPE_PRIMARY_STEREO, 2, &viewCount, views), "xrEnumerateViewConfigurationViews", fail);
+  uint32_t viewConfigurationCount;
+  XrViewConfigurationType viewConfigurations[4];
+  XRG(xrEnumerateViewConfigurations(state.instance, state.system, 4, &viewConfigurationCount, viewConfigurations), "xrEnumerateViewConfigurations", fail);
 
-  if (viewCount != 2) {
-    return lovrSetError("Headset view count must be 2");
+  for (uint32_t i = 0; !state.viewConfiguration && i < COUNTOF(supportedViewConfigurations); i++) {
```
gareth-morgan-nv, line 622: This seems to work correctly, with this change it treats `supportedViewConfigurations` as a prioritized list of requested view configurations, and chooses the first one that is available.

gareth-morgan-nv (approved):
I tested this with our OpenXR runtime implementation, and it produces the correct quad view result based on the values we provide in XrViewConfigurationView.

bjornbytes:
Thanks so much for testing!  Going to merge this.  In the future I'd like to add a way for Lua code to select a preferred view configuration, but for now the quad -> stereo -> mono priority is adequate.

PR #899, opened 2025-10-15
Blob Views

### LINKED ISSUE #890: Blob Views
Add the ability to create a Blob that references a region of another Blob.

API sketch is like `lovr.data.newBlobView(blob, offset, size)`.  Offset defaults to zero, size defaults to blob->size - offset.

snowkittykira:
Implemented as described in #890, but with offset required, and an optional name parameter.


src/modules/data/blob.c
```diff
@@ -9,12 +9,28 @@ Blob* lovrBlobCreate(void* data, size_t size, const char* name) {
...
   blob->size = size;
   blob->name = lovrStrdup(name);
+  blob->root = blob;
+  return blob;
+}
+
+Blob* lovrBlobCreateView(Blob* parent, size_t offset, size_t size, const char* name) {
+  Blob* blob = lovrCalloc(sizeof(Blob));
+  blob->ref = 1;
+  blob->data = parent->data + offset;
```
bjornbytes: void pointers don't really support pointer arithmetic, should be `(char*) parent->data + offset` to fix the windows build

bjornbytes:
Thanks for the PR!

PR #904, opened 2025-10-23
Fix mat4 mul

xiejiangzhi:


xiejiangzhi:
```
lovr/math.lua:45: expected table: 0x7f216ebfdc88 and table: 0x7f216ebfdfd0 to be equal
				LHS: { ['y'] = 2, ['z'] = 3, ['x'] = 1 }
				RHS: { ['y'] = 4, ['z'] = 3, ['x'] = 1 }
```
When I was debugging, I saw that the v in luax_pushvec3 was (1,4,3), but I don't know why it became 1,2,3 when Lua received it. 
~~I tried removing the two lines around setmetatable and that fixed the issue. This issue will not be fixed in this PR.~~
I see, it's the same bug I left behind. I will fix it in this PR

xiejiangzhi:
I discovered a bug with the copy code.
`lua_getmetatable` doesn't push a value when there's no metatable; it simply returns 0.

src/api/l_math_mat4.c
```diff
@@ -159,12 +159,12 @@ static int l_lovrMat4Mul(lua_State* L) {
...
     mat4_mulVec4(lovrMat4GetData(matrix), v);
     if (lua_istable(L, 2)) {
-      luax_pushvec3(L, v, luax_len(L, 2) > 0);
+      luax_pushvec3(L, v, luax_len(L, 2) > 0, 2);
     } else {
 #ifdef LOVR_USE_LUAU
       lua_pushvector(L, v[0], v[1], v[2]);
 #else
-      luax_pushvec3(L, v, false);
+      luax_pushvec3(L, v, false, -999);
```
bjornbytes: Does this line ever run?  Isn't `lua_type(L, 2)` always going to be a table, when not using Luau?

test/lovr/math.lua
```diff
@@ -16,4 +16,45 @@ group('math', function()
       end
     end)
   end)
+
+  group('mat4', function()
+    test('mul mat4', function()
+      local a = mat4():perspective(math.rad(80), 1440 / 900, 0.01, 0)
+      local b = mat4({ 0, 1.7, 0 }, { 0, 0, 0, 1 }):invert()
+
+      local function to_s(arr)
```
bjornbytes: Are you doing this to handle floating point precision?  `equal` supports an epsilon, so you could do `expect(a * b).to.equal(r, 1e-4)`

src/api/api.c
```diff
@@ -832,7 +832,11 @@ void luax_pushvec3(lua_State* L, float v[3], bool tableArray) {
...
     lua_setfield(L, -2, "z");
-    lua_getmetatable(L, 2);
-    lua_setmetatable(L, -2);
+  }
+
+  if (metatableSrcIndex >= -100 && lua_istable(L, metatableSrcIndex)) {
+    if (lua_getmetatable(L, metatableSrcIndex)) {
+      lua_setmetatable(L, -2);
+    }
   }
```
bjornbytes: Maybe the calling code could do this part?  I don't really like the -999...

```c
luax_pushvec3(L, v, luax_len(L, 2) > 0);
lua_getmetatable(L, 2);
lua_setmetatable(L, -2);
return 1;
```

xiejiangzhi: Maybe we can define a macro called NO_METATABLE_INDEX, what do you think?

src/api/l_math_mat4.c
xiejiangzhi: It should be, I'm not sure, so I added this line

src/api/api.c
bjornbytes: I'd rather do a different approach

- Require that the metatable index is valid (and remove the -999 call, since it shouldn't ever get run)
- Require the caller to set the metatable

bjornbytes:
Thanks for the fix

PR #907, opened 2025-10-25
mat4 new and set compatiblility with pos + scale

xiejiangzhi:


src/api/api.c
```diff
@@ -834,3 +834,27 @@ void luax_pushvec3(lua_State* L, float v[3], bool tableArray) {
...
+  }
+
+#ifdef LOVR_USE_LUAU
+  if (lua_type(L, index) == LUA_TQUATERNION) {
+    return true;
+  }
+#endif
+
+  return false;
+
```
bjornbytes: Can you remove this blank line?  Also, there should be a newline at the end of the file.

src/api/l_math_mat4.c
```diff
@@ -159,7 +159,10 @@ int l_lovrMat4Set(lua_State* L) {
     m[14] = position[2];
 
     // 1 more arg or 4 numbers: rotation, otherwise scale + rotation
-    if (top == index || ((top - index) == 3 && lua_type(L, top) == LUA_TNUMBER)) {
+    if (
+      (top == index && luax_isquat(L, index))
+      || ((top - index) == 3 && lua_type(L, top) == LUA_TNUMBER)
+    ) {
```
bjornbytes: I don't like `if`s that are on multiple lines like this.  Maybe just `if (luax_isquat(L, index) || (top - index == 3 && lua_type(L, top) == LUA_TNUMBER))`?

PR #910, opened 2025-11-16
check for conf.luau if Luau is used

TylkoKuba:
As I was testing the new Luau support I've noticed that we still try to look for `conf.lua` even when Luau is enabled which makes for slightly ugly codebase where just one file has .lua extension, not that the extension makes any difference 😅


bjornbytes:
Thanks, I'm noticing some other related issues in boot.lua regarding `main.lua` -- I don't think `lovr file.luau` works currently, and maybe loading main.luau from a fused zip doesn't work.  These can get cleaned up later though.

PR #912, opened 2025-12-02
Replaced bump VRAM allocator with list allocator

DonaldHays:
# Summary

This pull request replaces Lovr's current bump allocator for VRAM with a simple linear allocator that tracks regions of allocated and free space, allowing freed regions within a block to be reused. Previously, regions within a block could not be reused until the entire block was freed, and even a single living allocation would keep an entire block alive, which could lead to memory starvation.

# Background

A recent discussion in the "internals" channel on the Discord raised the issue that Lovr can exhaust video memory. Out of curiosity, I decided to take a look at the internals and saw that the existing manager was very simple, and thought I would make an attempt at an alternative that could address the issue.

Currently, Lovr allocates video memory from Vulkan in "blocks," which come in a variety of sizes, but often 256 MB. Then, a CPU-side allocator creates allocations within the block using a simple bump allocator: a cursor maintains a position within the block, and an allocation advances the cursor and returns the newly reserved space. The allocator will also increment a reference counter. To free an allocation, the allocator simply decrements the reference counter. If the reference counter hits zero, there are no more allocations within the block, and the allocator will release the entire block back to Vulkan.

This allocator is extremely simple. It is also extremely fast: allocations and frees are both O(1). However, there is a significant limitation. The allocator cannot reuse freed regions within a block. Memory cannot be reused until after the entire block is returned to Vulkan. If there's even a single long-lived allocation within a block, the entire block cannot be freed, and its memory cannot be reused. A mouse cursor, for example, would be a tiny allocation that may live for the life of the entire app. If its block fills up with other data that's then released, the mouse cursor alone will force the block to stay alive, with most of the space unusable. A small number of long-lived allocations across multiple blocks may force Lovr to exhaust VRAM.

# Proposed Solution

The allocator I implemented here is the simplest one I could imagine that could address the issue. It subdivides blocks into 16 KB "pages", and then tracks "regions" of these pages. Each allocator has a fixed-length array of potential region entries, one entry for each page. Since the largest fixed size block is 256 MB, these arrays have 256 MB / 16 KB -> 16,384 entries. A region entry has a flag for whether it's allocated, and a length.

To allocate, the allocator performs a linear scan over the regions, searching for a free region of sufficient size. It then allocates enough space for the allocation, and, if there's leftover space in the region, creates a new unallocated region.

All allocations are aligned to the 16 KB pages. The alignment of the requested allocation is not otherwise taken into account because, unless I misunderstand GPU memory alignment requirements, 16 KB satisfies any GPU alignment requirement.

To free memory, the allocator marks its corresponding region as unallocated, and coalesces the region with the preceding and succeeding regions, if possible.

The allocator continues to use the reference counter from the previous system, and frees blocks when the reference count hits zero. Also, continuing from the previous system, if an allocation overflows a block, requiring a new block, the allocator loses track of fine-grain knowledge of the previous block, and will no longer attempt to reuse memory within it, causing it to again only be able to free after all allocations within are released.

# Downsides, Tradeoffs, and Alternatives

This allocator was the simplest alternative I could think of that addresses the issue with the previous system. However, it has its own limitations.

This allocator adds 32 KB of memory for each allocator to manage the entry lists (2 bytes per entry, 16k entries). Less memory could be used with, say, dynamically-allocated linked list nodes, at the cost of requiring many dynamic allocations. More memory could be required to implement other alternative allocators.

There is simply no beating the O(1) performance of the previous bump allocator, however this allocator is also about the slowest possible alternative, requiring an O(n) scan of regions for both allocations and deallocations. In the worst case, this could require visiting all 16k entries. However, in practice there will likely be far fewer regions. A 32-bit 256x256 texture takes 16 pages, and a 32-bit 4k texture takes 4,096 pages. Plus, even if there are many regions, they're accessed in sequential order in a linear array, which is cache-friendly, and the queried logic is extremely simple. That said, a free-list allocator, while being O(n), would only scan over free regions, not all regions, and a buddy allocator would be a sublinear O(log(n)).

This allocator can only reuse memory within a single block per allocator. If allocations overflow a block, the allocator will only track fine-grained regions on the new block, and the previous block will behave like it did in the previous system, only freeing after all allocations within are freed. An alternative allocator could be designed to manage _all_ blocks.

Pages are 16 KB. Consequently, smaller allocations will waste space. The page size was chosen to try to balance reasonably fine granularity against the size of the per-allocator entry list. 16 KB pages and 256 MB blocks allow entries to be two bytes, finer granularity or larger blocks would require larger list entries.

### DIFF
--- a/src/core/gpu_vk.c
+++ b/src/core/gpu_vk.c
@@ -3409,10 +3409,19 @@ static gpu_memory* allocate(gpu_memory_type type, VkMemoryRequirements info, VkD
         // If there's leftover room, mark the leftover free region, and shrink
         // the current region
         if (requiredPages < region->pageCount) {
-          gpu_alloc_entry* next = &allocator->regions[i + requiredPages];
-          next->allocated = false;
-          next->pageCount = region->pageCount - requiredPages;
+          gpu_alloc_entry* freeRegion = &allocator->regions[i + requiredPages];
+          freeRegion->allocated = false;
+          freeRegion->pageCount = region->pageCount - requiredPages;
           region->pageCount = requiredPages;
+
+          // Coalesce with the next region, if possible
+          uint32_t nextIndex = i + requiredPages + freeRegion->pageCount;
+          if (nextIndex < allocator->pageCount) {
+            gpu_alloc_entry* nextRegion = &allocator->regions[nextIndex];
+            if (!nextRegion->allocated) {
+              freeRegion->pageCount += nextRegion->pageCount;
+            }
+          }
         }
 
         allocator->block->refs++;
@@ -3448,19 +3457,25 @@ static gpu_memory* allocate(gpu_memory_type type, VkMemoryRequirements info, VkD
         memory->pointer = NULL;
       }
 
-      allocator->block = memory;
+      memory->refs = 1;
 
-      // Mark the initial region
-      allocator->pageCount = blockSize / GPU_PAGE_SIZE;
-      allocator->regions[0].allocated = true;
-      allocator->regions[0].pageCount = requiredPages;
-
-      // If there's additional room in the block, mark the free region
-      if (requiredPages < allocator->pageCount) {
-        allocator->regions[requiredPages].allocated = false;
-        allocator->regions[requiredPages].pageCount = allocator->pageCount - requiredPages;
+      // Memory only receives an allocator if it can host multiple allocations
+      // (i.e. it has a fixed block size and this block is not oversized)
+      if(blockSize && info.size < blockSize) {
+        allocator->block = memory;
+  
+        // Mark the initial region
+        allocator->pageCount = blockSize / GPU_PAGE_SIZE;
+        allocator->regions[0].allocated = true;
+        allocator->regions[0].pageCount = requiredPages;
+  
+        // If there's additional room in the block, mark the free region
+        if (requiredPages < allocator->pageCount) {
+          allocator->regions[requiredPages].allocated = false;
+          allocator->regions[requiredPages].pageCount = allocator->pageCount - requiredPages;
+        }
       }
-      allocator->block->refs = 1;
+      
       *offset = 0;
       return memory;
     }
@@ -3472,48 +3487,49 @@ static gpu_memory* allocate(gpu_memory_type type, VkMemoryRequirements info, VkD
 
 static void release(gpu_memory* memory, VkDeviceSize offset) {
   if (memory) {
-    if (--memory->refs == 0) {
-      condemn(memory->handle, VK_OBJECT_TYPE_DEVICE_MEMORY);
-      memory->handle = NULL;
+    gpu_allocator* allocator = NULL;
 
-      for (uint32_t i = 0; i < COUNTOF(state.allocators); i++) {
-        if (state.allocators[i].block == memory) {
-          state.allocators[i].block = NULL;
-        }
+    for (uint32_t i = 0; i < COUNTOF(state.allocators); i++) {
+      if (state.allocators[i].block == memory) {
+        allocator = &state.allocators[i];
+        break;
+      }
+    }
+      
+    if (--memory->refs == 0) {
+      // If the memory has an allocator, reset it, otherwise free the memory
+      if (allocator) {
+        gpu_alloc_entry* region = &allocator->regions[0];
+        region->allocated = false;
+        region->pageCount = allocator->pageCount;
+      } else {
+        condemn(memory->handle, VK_OBJECT_TYPE_DEVICE_MEMORY);
+        memory->handle = NULL;
       }
     } else {
-      for (uint32_t i = 0; i < COUNTOF(state.allocators); i++) {
-        gpu_allocator* allocator = &state.allocators[i];
-        if (allocator->block == memory) {
-          // Mark the region for this allocation as free
-          uint32_t pageOffset = offset / GPU_PAGE_SIZE;
-          gpu_alloc_entry* region = &allocator->regions[pageOffset];
-          region->allocated = false;
+      // Mark the region for this allocation as free
+      uint32_t pageOffset = offset / GPU_PAGE_SIZE;
+      gpu_alloc_entry* region = &allocator->regions[pageOffset];
+      region->allocated = false;
 
-          // See if we can coalesce this region with the next region
-          uint32_t nextOffset = pageOffset + region->pageCount;
-          if (nextOffset < allocator->pageCount) {
-            gpu_alloc_entry* nextRegion = &allocator->regions[nextOffset];
-            if (!nextRegion->allocated) {
-              region->pageCount += nextRegion->pageCount;
-            }
-          }
+      // See if we can coalesce this region with the next region
+      uint32_t nextOffset = pageOffset + region->pageCount;
+      if (nextOffset < allocator->pageCount) {
+        gpu_alloc_entry* nextRegion = &allocator->regions[nextOffset];
+        if (!nextRegion->allocated) {
+          region->pageCount += nextRegion->pageCount;
+        }
+      }
 
-          // See if we can coalesce with the previous region. We have to loop
-          // through regions from the beginning because we don't know where the
-          // last region started
-          for(uint32_t j = 0; j < allocator->pageCount; j += allocator->regions[j].pageCount) {
-            gpu_alloc_entry* previousRegion = &allocator->regions[j];
-            if (pageOffset == j + previousRegion->pageCount) {
-              // We found the previous region, coalesce if it's free
-              if (!previousRegion->allocated) {
-                previousRegion->pageCount += region->pageCount;
-              }
-
-              // Since we found the previous region, we break no matter if we
-              // coalesced or not
-              break;
-            }
+      // See if we can coalesce with the previous region. We have to loop
+      // through regions from the beginning because we don't know where the
+      // last region started
+      for(uint32_t i = 0; i < pageOffset; i += allocator->regions[i].pageCount) {
+        gpu_alloc_entry* previousRegion = &allocator->regions[i];
+        if (i + previousRegion->pageCount == pageOffset) {
+          // We found the previous region, coalesce if it's free
+          if (!previousRegion->allocated) {
+            previousRegion->pageCount += region->pageCount;
           }
         }
       }
--- a/src/core/gpu_vk.c
+++ b/src/core/gpu_vk.c
@@ -3460,8 +3460,8 @@ static gpu_memory* allocate(gpu_memory_type type, VkMemoryRequirements info, VkD
       memory->refs = 1;
 
       // Memory only receives an allocator if it can host multiple allocations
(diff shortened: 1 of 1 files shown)


src/core/gpu_vk.c
```diff
@@ -3381,13 +3395,31 @@ static gpu_memory* allocate(gpu_memory_type type, VkMemoryRequirements info, VkD
...
+        region->allocated = true;
+
+        // If there's leftover room, mark the leftover free region, and shrink
+        // the current region
+        if (requiredPages < region->pageCount) {
+          gpu_alloc_entry* next = &allocator->regions[i + requiredPages];
+          next->allocated = false;
+          next->pageCount = region->pageCount - requiredPages;
+          region->pageCount = requiredPages;
+        }
```
bjornbytes, line 3426: Maybe this could merge with the next region, similar to `release`.

src/core/gpu_vk.c
```diff
@@ -3381,13 +3395,31 @@ static gpu_memory* allocate(gpu_memory_type type, VkMemoryRequirements info, VkD
   };
 
   uint32_t blockSize = blockSizes[type];
-  uint32_t cursor = ALIGN(allocator->cursor, info.alignment);
+  ASSERT(blockSize <= (GPU_ALLOC_PAGES_LENGTH * GPU_PAGE_SIZE), "Block size larger than allocator can handle") return NULL;
+
+  uint32_t requiredPages = ALIGN(info.size, GPU_PAGE_SIZE) / GPU_PAGE_SIZE;
+
+  if (allocator->block) {
```
bjornbytes, line 3403: This could say `&& info.size <= blockSize` so that "oversized blocks" (resources bigger than block size) can avoid needlessly searching through every block, when we know they're  not going to fit.

src/core/gpu_vk.c
```diff
@@ -3417,7 +3449,17 @@ static gpu_memory* allocate(gpu_memory_type type, VkMemoryRequirements info, VkD
       }
 
       allocator->block = memory;
-      allocator->cursor = info.size;
+
+      // Mark the initial region
+      allocator->pageCount = blockSize / GPU_PAGE_SIZE;
```
bjornbytes: Note that this can technically be derived from the memory type / allocator index, so it could be removed if needed.  It may still be useful for convenience or for oversized blocks, and `gpu_allocator` is much bigger now so we don't really need to save the 4 bytes.

src/core/gpu_vk.c
bjornbytes, line 3442: Instead of `info.size` here, should it be allocating `ALIGN(info.size, GPU_PAGE_SIZE)` aka `requiredPages * GPU_PAGE_SIZE` for oversized blocks?

src/core/gpu_vk.c
bjornbytes: Also, this page count isn't correct for oversized blocks.  I think it's innocuous since they don't share space with other resources, but it could be confusing or cause bugs later.

src/core/gpu_vk.c
```diff
@@ -3428,15 +3470,52 @@ static gpu_memory* allocate(gpu_memory_type type, VkMemoryRequirements info, VkD
...
+
+      for (uint32_t i = 0; i < COUNTOF(state.allocators); i++) {
+        if (state.allocators[i].block == memory) {
+          state.allocators[i].block = NULL;
+        }
+      }
+    } else {
+      for (uint32_t i = 0; i < COUNTOF(state.allocators); i++) {
+        gpu_allocator* allocator = &state.allocators[i];
+        if (allocator->block == memory) {
```
bjornbytes: We have 4 padding bytes in `gpu_memory`, we could use them to store its memory type to avoid having to search for the allocator.

It could also be used for the block size, but we don't really need it for anything atm.

src/core/gpu_vk.c
```diff
@@ -3428,15 +3470,52 @@ static gpu_memory* allocate(gpu_memory_type type, VkMemoryRequirements info, VkD
...
 }
 
-static void release(gpu_memory* memory) {
-  if (memory && --memory->refs == 0) {
-    condemn(memory->handle, VK_OBJECT_TYPE_DEVICE_MEMORY);
-    memory->handle = NULL;
+static void release(gpu_memory* memory, VkDeviceSize offset) {
+  if (memory) {
+    if (--memory->refs == 0) {
+      condemn(memory->handle, VK_OBJECT_TYPE_DEVICE_MEMORY);
```
bjornbytes: Rather than unconditionally freeing the block, we could reset the block if it's the current block of its allocator.  This would theoretically avoid thrashing if the first allocation in a block is repeatedly allocated/freed.

bjornbytes: But maybe oversized blocks should always get freed instead of reset, so they don't hog up space.

src/core/gpu_vk.c
```diff
@@ -3428,15 +3470,52 @@ static gpu_memory* allocate(gpu_memory_type type, VkMemoryRequirements info, VkD
...
+          }
 
-    for (uint32_t i = 0; i < COUNTOF(state.allocators); i++) {
-      if (state.allocators[i].block == memory) {
-        state.allocators[i].block = NULL;
-        state.allocators[i].cursor = 0;
+          // See if we can coalesce with the previous region. We have to loop
+          // through regions from the beginning because we don't know where the
+          // last region started
+          for(uint32_t j = 0; j < allocator->pageCount; j += allocator->regions[j].pageCount) {
```
bjornbytes: Could consider `j < pageOffset` here instead of the `break`.  It might be more robust but also might be too clever.

bjornbytes:
Oh and if you don't feel like making further changes, I'm happy to take it over, just let me know.

DonaldHays:
@bjornbytes  Good feedback! I'll make some revisions tomorrow.

src/core/gpu_vk.c
DonaldHays: It definitely should, good catch!

DonaldHays:
@bjornbytes Made changes based on feedback. I'll reply to individual comments, but one of the bigger changes is that I now only assign an allocator to a block if the block can receive multiple allocations (so, for example, an allocator won't point to an oversized block). This obviates some concerns, like redundantly stepping over regions in an oversized block or incorrect allocator page counts (because allocators won't ever point at such blocks), and it enables resetting some blocks instead of freeing them when possible. As an added bonus, if an allocator is managing a block, and the block still has room in it, but then an oversized request comes in (say, a 400 MB buffer), a new block will be created for it, but the allocator will continue to manage the previous block.

src/core/gpu_vk.c
DonaldHays: This should be obviated by the fact I now only let allocators point to blocks that can have multiple allocations. So, it won't scan over an oversized block, because `allocator->block` will be `NULL`.

src/core/gpu_vk.c
DonaldHays: This was a really good catch, fixed.

src/core/gpu_vk.c
DonaldHays: Since I made allocators not point at oversized blocks, the `pageCount` shouldn't even apply to such blocks, obviating the situation of the variable being incorrect.

As for deriving an allocator's page count from memory type and allocator index, I store it for the sake of the later `release` function. The static `blockSize` array could be moved to file scope easily enough, but the memory type was a bit fuzzier to me. Is that just the same as a `gpu_allocator`'s `memoryType` variable?

src/core/gpu_vk.c
DonaldHays: I'm not sure which four bytes in `gpu_memory` you're referring to as padding?

src/core/gpu_vk.c
DonaldHays: I made this requested change. If a memory block is managed by an allocator, it'll now be reset, otherwise it'll be freed.

src/core/gpu_vk.c
DonaldHays: I liked this suggestion and implemented it. Rather than being too clever, I thought it was just more straightforward than a conditional break.

src/core/gpu_vk.c
DonaldHays: Since allocators will no longer manage oversized blocks, I don't think this is necessary.

DonaldHays:
Don't merge yet. Just realized one edge case that could result in a memory leak. Will fix soon.

src/core/gpu_vk.c
bjornbytes: The struct is 24 bytes right now, there are 4 unused bytes after `refs`, because it has to be aligned to 8 bytes.  So we could add `gpu_memory_type` afterwards to use it for the memory -> allocator lookup in `release` without increasing the size of the struct.

```c
struct gpu_memory {
  VkDeviceMemory handle;
  void* pointer;
  uint32_t refs;
  gpu_memory_type type; // still 24 bytes
};

// `allocate` assigns block->type during allocation

// ...later, in release

static void release(gpu_memory* memory) {
  // ...

  gpu_allocator* allocator = &state.allocators[state.allocatorLookup[memory->type]];

  // no need to loop to search for allocator
}
```

It's not crucial, just saves a few lines

DonaldHays:
False alarm on the memory leak concern :sweat_smile:.

I thought a leak was possible because of the new feature that resets allocator-managed blocks when their references hit zero, instead of releasing them. I thought there might have been an edge case that would allocate a new block, move the allocator to that block, and _not_ free an empty-but-not-released allocator-managed block. That doesn't happen, though, because any allocation that could fit in the empty managed block, up to and including an allocation that fills it entirely, will simply fill the required space, without triggering allocation of a new block. And any oversized allocation would create a new block, but not move the allocator over to it.

This scrutiny did expose an opportunity to tighten things up slightly, however, which I did in 8eb6d63. I now only have an allocator start managing a new block if there's at least one full page available to be managed.

src/core/gpu_vk.c
DonaldHays: I get what you mean now! Good idea, and I implemented it. Saves a linear lookup of the allocator.

src/core/gpu_vk.c
DonaldHays: I started going down the path of removing the `pageCount` field on `gpu_allocator` like you suggested, but I need that information at numerous sites in two different functions, and ended up feeling the stored variable was simpler than recalculating it in at least two places. If you feel strongly that you would rather have the stored field removed, I'll go ahead and do it, but otherwise I think I like where it's at now.

DonaldHays:
@bjornbytes okay, I think I've incorporated all the feedback, and am happy where it's at right now. Let me know how you feel!

bjornbytes:
This is awesome, thanks for the PR!

bjornbytes:
> The alignment of the requested allocation is not otherwise taken into account because, unless I misunderstand GPU memory alignment requirements, 16 KB satisfies any GPU alignment requirement.

Surprisingly, I ran into a GPU that requires 64KB aligned textures (got a fun GPU crash on there).  I updated the allocator to use the maximum of the page size and the required alignment.

PR #913, opened 2025-12-04
Add OpenXR loader init properties support

hougantc-nvda:
- Support custom loader properties via lovr.conf initproperties table
- Fix null terminator buffer overflow in extension string allocation
- LOVR_BUILD_WITH_SYMBOLS now exports all symbols on Windows
- Add NVIDIA to contributors


src/util.h
```diff
@@ -60,7 +66,7 @@ int lovrSetError(const char* format, ...);
 typedef void fn_log(void*, int, const char*, const char*, va_list);
 enum { LOG_DEBUG, LOG_INFO, LOG_WARN, LOG_ERROR };
 void lovrSetLogCallback(fn_log* callback, void* userdata);
-void lovrLog(int level, const char* tag, const char* format, ...);
+void LOVR_EXPORT lovrLog(int level, const char* tag, const char* format, ...);
```
bjornbytes: Could you build LÖVR with the `LOVR_BUILD_WITH_SYMBOLS` CMake variable enabled?  That will export all the symbols, which might be easier than changing the code to export individual functions.

src/modules/headset/headset.h
```diff
@@ -252,3 +264,6 @@ bool lovrHeadsetIsSupported(void);
 void lovrHeadsetGetVulkanPhysicalDevice(void* instance, uintptr_t physicalDevice);
 uint32_t lovrHeadsetCreateVulkanInstance(void* instanceCreateInfo, void* allocator, uintptr_t instance, void* getInstanceProcAddr);
 uint32_t lovrHeadsetCreateVulkanDevice(void* instance, void* deviceCreateInfo, void* allocatoor, uintptr_t device, void* getInstanceProcAddr);
+LOVR_EXPORT uintptr_t lovrHeadsetGetInstanceHandle(void);
+LOVR_EXPORT uintptr_t lovrHeadsetGetSessionHandle(void);
+LOVR_EXPORT uint64_t lovrHeadsetGetSystemId(void);
```
bjornbytes: I recently added some methods to expose the OpenXR handles: `xr_get_instance`, `xr_get_system`, and `xr_get_session`.  We should use those instead.  It looks like I forgot to remove the implementations for `lovrHeadsetGetInstanceHandle` / `lovrHeadsetGetSessionHandle` when I did that, sorry if it was confusing!

Also, I'm planning on adding `xr_is_extension_supported` / `xr_is_extension_enabled`.

src/api/l_system.c
```diff
@@ -384,6 +390,7 @@ static const luaL_Reg lovrSystem[] = {
   { "getClipboardText", l_lovrSystemGetClipboardText },
   { "setClipboardText", l_lovrSystemSetClipboardText },
   { "_getScrollDelta", l_lovrSystemGetScrollDelta },
+  { "setEnvironmentVariable", l_lovrSetEnvironmentVariable },
```
bjornbytes: Could you do something like this?

```lua
function setenv(var, val)
  local command = lovr.system.getOS() == 'Windows' and 'set' or 'export'
  os.execute(('%s %s="%s"'):format(command, var, val))
end
```

bjornbytes: Maybe this wouldn't set it for the current process?

src/modules/headset/headset.h
hougantc-nvda: Thanks, I didn't realize those existed. Switching them over.

src/api/l_system.c
hougantc-nvda: Now that I see you've upgraded OpenXR, we can use xrInitializeLoaderKHR for what we need. Namely to set XR_RUNTIME_JSON at runtime.

No longer need to setEnvironmentVariable. Removed that functionality.

src/modules/headset/headset.h
hougantc-nvda: I needed to export in the header as well since on Windows, it would incorrectly export return type as int.

src/util.h
hougantc-nvda: That works out better. I also had to add WINDOWS_EXPORT_ALL_SYMBOLS to LOVR_BUILD_WITH_SYMBOLS on Windows. Windows doesn't respect C_VISIBILITY_PRESET "default".

src/modules/headset/headset.c
```diff
@@ -437,6 +451,37 @@ bool lovrHeadsetConnect(void) {
   }
 #endif
 
+  // Initialize loader with properties (for custom runtime paths, etc.)
+#if defined(__linux__) || defined(__APPLE__) || defined(_WIN32)
```
hougantc-nvda, line 455: It seems you have other code for Android so I targeted xrInitializeLoaderKHR to desktop platforms.

bjornbytes:
Given the `xrInitializeLoaderKHR` doesn't really depend on any of lovr's internal OpenXR stuff, is there a reason the loader properties need to be plumbed through lovr?  Could you do something like `require('myplugin').initLoader({ properties = {} })` in conf.lua and set up the loader before booting the rest of lovr?  These initialization properties feel a bit too low-level / esoteric compared to the usual configuration lovr exposes.

bjornbytes:
I think this has been resolved now that the CloudXR plugin has been released, but feel free to open other issues/PRs if you need other changes

PR #916, opened 2025-12-16
Task

bjornbytes:
This is an experiment to add a coroutine-based task system to LÖVR, with the main goal of making it easier for projects to scale to multiple CPU cores and run expensive work in the background.

Design
---

The basic idea is to create lots of tasks (coroutines) that are each running independent work.  Lots of tasks can all run more or less at the same time, cooperatively.

Normally, Lua coroutines are still single threaded, and don't allow for parallelism.  However, LÖVR can achieve parallelism with them by running **asynchronous functions**.

When a coroutine calls an asynchronous function, it will start work on a thread pool, yield itself, and wake up automatically once the work is complete.  While the coroutine is yielded, other coroutines get a chance to run. 

In this way, many coroutines can all run work "at the same time", with all of them doing work on a thread pool, making use of all of the CPU cores.

API
---

Everything is in a new module, `lovr.task`:

```lua
task = lovr.task.start(fn, ...) --> creates a coroutine and starts it with args
success, results = lovr.task.resume(co, ...) --> starts/resumes a task
lovr.task.poll() --> iterate over ready tasks (for co in lovr.task.poll() do ... end)
lovr.task.wait(...tasks) --> wait for tasks to finish.  returns results/error
lovr.task.getStatus(task) --> get task status (waiting, ready, complete, failed, running)

-- New callback: called when a task is ready to run after yielding
-- Default behavior is to just resume the task and propagate errors,
-- but you can override it to implement custom scheduling/error behavior
-- boot.lua polls tasks before lovr.update automatically
function lovr.taskready(task)
  assert(lovr.task.resume(task))
end
```

Notes:

- Create a task with `coroutine.create` or `lovr.task.start`.  Note that task functions share the global scope with the caller and other tasks.  You don't have to use `Channel` to communicate with the task.
- `success, ...results = lovr.task.wait(...tasks)`.  Blocks until all the tasks are complete, returning their results or the first error that was encountered.
  - Handling multiple return values: Results are collapsed similar to Lua's existing convention: All tasks return their first return value, except for the last task, which returns all of its results.  So if you have 3 tasks that return 1, 2, 3, `lovr.task.wait(a, b, c)` returns `1, 1, 1, 2, 3`.
  - Note that this is an asynchronous function!  This means:
    - If you call it from within a task, it will yield, and become ready once the tasks are complete.
    - If you call from outside a task, this function will block, running tasks and jobs until all the tasks finish.
  - Note that if you wait on a task, it will not show up in `lovr.taskready` anymore, because it's already finished.  You can choose to "fire and forget" a task and let `lovr.taskready` take care of running it on a future frame, or explicitly wait on the task if you need its results sooner.
- `success, ...results = lovr.task.resume(task, ...args)`.  Resumes a task, if possible, otherwise returns false and an error.
  - You'll get an error if the task is already finished, has an error, or isn't ready to run yet.
  - If the task is resuming after a call to an asynchronous function, `args` are ignored and the results from the async operation are given to the task instead.
  - If the task ran successfully and yielded/finished, success will be `true` and you'll get any yield/return values.
  - Note that if you resume a task manually, it won't show up in `lovr.taskready`, since you already ran it.

Asynchronous Functions
---

Asynchronous functions are where the magic happens.  Certain functions in LÖVR have been converted to "asynchronous functions".  When an async function is called in a task, it yields the task and starts an operation in the background.  Once the work is done, the task gets rescheduled again, and will show up in `lovr.taskready` on the next frame (unless it gets resumed or waited on sooner).

For example, take `lovr.data.newImage`.  It still works exactly the same way, but it doesn't block the main thread if you call it in a task:

```lua
-- Creates a coroutine, starts loading the image, and yields immediately
task = lovr.task.start(function()
  return lovr.data.newImage('file.png')
end)

print(lovr.task.getStatus(task)) --> 'waiting', still loading image

-- wait until image is loaded and get the result
local image = lovr.task.wait(task)
```

By itself this isn't interesting.  But there are interesting things you can do with asynchronicity and parallelism.

Here's an example that loads an image in the background:

```lua
function lovr.load()
  lovr.task.start(function()
    local image = lovr.data.newImage('file.png')
    texture = lovr.graphics.newTexture(image)
    loaded = true
  end)
end

function lovr.draw(pass)
  if loaded then
    pass:draw(texture, 0, 2, -3)
  else
    pass:text('Loading...', 0, 2, -3)
  end
end
```

Here's an example that loads 100 images on multiple threads in parallel:

```lua
local tasks = {}
for i = 1, 100 do
  local task = lovr.task.start(function(file)
    return lovr.data.newImage(file)
  end, 'file' .. i .. '.png')
  table.insert(tasks, task)
end
local images = lovr.task.wait(tasks)
```

The initial list of async functions is:

- `lovr.headset.newModel`
- `lovr.data.newImage`
- `lovr.data.newModelData`
- `lovr.data.newSound`
- `lovr.graphics.newTexture`
- `lovr.graphics.newModel`
- `lovr.graphics.compileShader`
- `lovr.graphics.newShader`
- `Buffer:getData`
- `Texture:getPixels`
- `lovr.audio.newSource`
- `lovr.task.wait`
- `lovr.thread.call` (calls any Lua function on a thread asynchronously)
- `lovr.timer.sleep`

Discussion
---

- Why isn't `taskready` just a regular event?
  - Had to remove tasks from queue if they are resumed/waited before being consumed.  Felt weird to remove events like this.
- Will Thread continue to exist?
  - Yes, but it will be much more niche.

TODO
---

- [x] Make .wait work with tables too
- [x] Add more async functions
- [x] Consider adding a way to sleep a task, like `lovr.task.sleep` (async, wakes up task later, could cause problems if you want the wait time to be based on `dt`?)  Maybe just `lovr.timer.sleep` is async
- [x] Add an async function that runs Lua code on the thread pool
- [x] ~Add a parallel-for helper (potentially both task-based and thread-based)~
- [x] Add a way of getting the underlying coroutine for a task?  Could be useful for using the `debug` library with the task to inspect its state.


shakesoda:
for my use cases i tend to want a parallel for (often 2d/3d, too), fwiw

bjornbytes:
There will be a (async) function to run Lua code on a worker thread, and likely a parallel-for helper on top of this.  I haven't figured out the specific API for the parallel-for function yet though.

bjornbytes:
Decided to change the API back to plain coroutines instead of Task objects.  I think it's a little easier to understand (it's more clear that tasks are just coroutines), interops better with other libraries, and has slightly less overhead.  Semantics are pretty much exactly the same.

Also, added a way to wait on a polled condition instead of a job.  This is good for IO bound stuff like sleeps, readbacks, async IO, condition variables, etc.  As a test, I made `lovr.timer.sleep` async.

bjornbytes:
Let's try it!!

PR #918, opened 2025-12-20
Steam Frame Controller Bindings

bjornbytes:
Adds support for the Steam Frame controller bindings documented [here]( https://partner.steamgames.com/doc/steamframe/engines/custom).

Adds 5 new DeviceButtons:

- `dpup`
- `dpdown`
- `dpleft`
- `dpright`
- `bumper`


bjornbytes:
It would be great to add support for [`XR_EXT_dpad_binding`](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_EXT_dpad_binding) as well, so controllers with thumbsticks can support the new dpad buttons.

src/modules/headset/headset.c
```diff
@@ -1248,6 +1272,62 @@ bool lovrHeadsetConnect(void) {
...
+      { ACTION_GRIP_POSE, "/user/hand/left/input/grip/pose" },
+      { ACTION_GRIP_POSE, "/user/hand/right/input/grip/pose" },
+      { ACTION_POINTER_POSE, "/user/hand/left/input/aim/pose" },
+      { ACTION_POINTER_POSE, "/user/hand/right/input/aim/pose" },
+      { ACTION_PINCH_POSE, "/user/hand/left/input/pinch_ext/pose" },
+      { ACTION_PINCH_POSE, "/user/hand/right/input/pinch_ext/pose" },
+      { ACTION_POKE_POSE, "/user/hand/left/input/poke_ext/pose" },
+      { ACTION_POKE_POSE, "/user/hand/right/input/poke_ext/pose" },
+      { ACTION_PALM_POSE, "/user/hand/left/input/palm_ext/pose" },
+      { ACTION_PALM_POSE, "/user/hand/right/input/palm_ext/pose" },
```
DonaldHays, line 1286: I'm not familiar with Lovr's binding system, but I'm noticing bindings for pinch, poke, and palm, and from what I can see the Steam Frame doesn't support those. Is this okay?

src/modules/headset/headset.c
```diff
@@ -1248,6 +1272,62 @@ bool lovrHeadsetConnect(void) {
...
+      { ACTION_TRIGGER_DOWN, "/user/hand/left/input/trigger/value" },
+      { ACTION_TRIGGER_DOWN, "/user/hand/right/input/trigger/value" },
```
DonaldHays: I notice that you're binding `ACTION_TRIGGER_DOWN` to `trigger/value`, when `trigger/click` is available. Is there a reason `trigger/value` is better than `trigger/click` here?

src/modules/headset/headset.c
bjornbytes: Yeah, the `XR_EXT_hand_interaction` extension lets you apply these bindings to any hand/controller device, spec language is:

> The runtime must support all of the following action subpaths on all interaction profiles that are valid for the user paths of /user/hand/left and /user/hand/right, including those interaction profiles enabled through extensions.
> …/input/aim/pose
> …/input/grip/pose
> …/input/pinch_ext/pose
> …/input/poke_ext/pose

src/modules/headset/headset.c
bjornbytes: No, in fact I trust `trigger/click` a little more here (the float will be converted to a boolean, but there's a chance it could use a different threshold that's worse, whereas `click` is more likely to use the threshold intended by the runtime).  I'll change it!

src/modules/headset/headset.c
bjornbytes: (And same for palm pose)

PR #919, opened 2025-12-21
Hand Tracking Microgestures

bjornbytes:
Map `XR_META_hand_tracking_microgestures` extension bindings to the dpad buttons.  You can swipe your thumb along your index finger to get dpad input.  Feels great!

image

Doesn't expose the "center" tap yet.  Should there be a `dpcenter` button, or maybe this could get mapped to `thumbrest`?

### DIFF
--- a/src/modules/headset/headset.c
+++ b/src/modules/headset/headset.c
@@ -339,6 +339,7 @@ static struct {
     bool layerEquirect2;
     bool layerSettings;
     bool localFloor;
+    bool microgestures;
     bool ml2Controller;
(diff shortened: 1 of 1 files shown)


bjornbytes:
Decided to hold off on the trackpad/joystick virtual dpad buttons.

I think these should be added as `thumbstick/left`, `thumbstick/right`, `thumbstick/up`, and `thumbstick/down` buttons (same for touchpad).  That way, there is no conflict between real dpads.

Mapping the hand microgestures to dpad makes sense though.

PR #920, opened 2025-12-22
Account for alignment in required region size

DonaldHays:
Per [this discussion](https://github.com/bjornbytes/lovr/commit/5b02ffb38732fa97358396bfe1c364790b433672#r173297954), this PR accounts for alignment offsets when considering the total number of pages required to create an allocation within a region.


bjornbytes:
Makes sense, thanks for the fix!

PR #922, opened 2026-01-04
Fix Shader:clone() rejecting option flags with the 'flag_' prefix

SCMcLaughlin:
Shader:clone() rejects valid option flags if they are given as string names with the "flag_" prefix.

Ex:
```lua
function lovr.load()
	local shader = lovr.graphics.newShader("unlit", "unlit")
	-- works
	local shader2 = shader:clone{ tonemap = true }
	-- Error: "Unknown shader flag flag_tonemap"
	local shader3 = shader:clone{ flag_tonemap = true }
end
```

Fix this by skipping over the "flag_" prefix in `l_lovrShaderClone()` prior to hashing, as is done in `lovrShaderCreate()`.

### DIFF
--- a/src/api/l_graphics_shader.c
+++ b/src/api/l_graphics_shader.c
@@ -15,7 +15,12 @@ static int l_lovrShaderClone(lua_State* L) {
     ShaderFlag flag = { 0 };
     flag.value = lua_isboolean(L, -1) ? (double) lua_toboolean(L, -1) : lua_tonumber(L, -1);
     switch (lua_type(L, -2)) {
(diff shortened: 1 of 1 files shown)


bjornbytes:
The intent was for both `flag_tonemap` and `tonemap` in the shader to use the `tonemap` flag in Lua.  Is there a reason you want to use the `flag_tonemap` name to set the flag?

SCMcLaughlin:
Never mind, I was doubly confused about `newShader()` and `Shader:clone()`.

bjornbytes:
Yeah, we definitely need better documentation about the flags.  But it looks like newShader/clone are at least consistent (flag_ prefix consistently rejected).

Issue #868, opened 2025-05-31
Issue with Joint initialization


bjornbytes:
Background: In Jolt, when you create a slider, you have to specify a frame of reference for both colliders.  This consists of a point (can be in world space or relative to the collider), a slider axis, and a normal axis.

We currently aren't specifying the 2 points at all, they're both at 0,0,0, which causes sliders to not behave as expected.  There's an option called `AutoDetectPoint`, but we don't use it.  I tried using it, but it picks a single anchor point between the two colliders, weighted by mass, which maybe isn't very useful.  Instead, LÖVR should probably use the positions of the 2 colliders as the 2 points.

Additionally there's the issue of the slider axis and normal axis.  We currently use a helper function `SetSliderAxis` for this, which computes the normal axis as any vector perpendicular to the main slider axis.  This is probably okay.

bjornbytes:
Similarly, we don't set the hinge normal axes.  We should set both of those to be perpendicular to the hinge axis.

PR #923, opened 2026-01-21
Fix slider joint point initialization

jmiskovic:
The slider constraint needs the two anchor points either initialized or auto-detected.

This becomes very apparent if collider is created at least few meters away from the origin point, without the fix the collider is not constrained well to the axis.

### DIFF
--- a/src/modules/physics/physics.c
+++ b/src/modules/physics/physics.c
@@ -2955,6 +2955,7 @@ SliderJoint* lovrSliderJointCreate(Collider* a, Collider* b, float axis[3]) {
   JPH_SliderConstraintSettings settings;
   JPH_SliderConstraintSettings_Init(&settings);
   JPH_SliderConstraintSettings_SetSliderAxis(&settings, vec3_toJolt(axis));
+  settings.autoDetectPoint = true;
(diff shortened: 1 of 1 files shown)


bjornbytes:
When I last looked at this in #868 I wrote

> We currently aren't specifying the 2 points at all, they're both at 0,0,0, which causes sliders to not behave as expected. There's an option called AutoDetectPoint, but we don't use it. I tried using it, but it picks a single anchor point between the two colliders, weighted by mass, which maybe isn't very useful. Instead, LÖVR should probably use the positions of the 2 colliders as the 2 points.

Any thoughts on using the weighted midpoint via AutoDetectPoint vs. using the collider positions as the anchors?

jmiskovic:
Ah, I missed that ticket and discovered this omission independently by investigating the Jolt samples.

I tested it for a bit, making a heavy object 100m away from a light object and applying impulse force to the lighter object. This should be a very unfavorable setup but it still works quite well and stable.

I also tried to manually set the two positions but I didn't have much luck with it. I placed each at the colliders' centers. From what I observed the simulation tried to bring the two anchor points to the same world position, so colliders ran into each other.

IMO autodetection works great.

bjornbytes:
Sounds good to me, thanks!!

PR #925, opened 2026-01-28
Add body tracking based on XR_BD_body_tracking

### LINKED ISSUE #840: Body Tracking
A new OpenXR body tracking extension was released yesterday, bringing the total up to 3 (4 if you include vive trackers).  There's probably going to be a standardized body tracking skeleton at some point, so LÖVR should start to think about it.

Facebook's extension includes hand bones, but (famously) doesn't include legs.  [Pico's extension](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#_conventions_of_body_joints) has a cool picture and has 24 joints for the whole body.  [HTC](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#_conventions_of_body_joints_3) also has a cool picture and is very close to Pico.  But both of these stop at the wrist and don't include hand tracking.  Vive trackers only have a few joints defined (elbow, shoulder, chest, waist, knee, foot, wrist, ankle) with no hierarchy, and there might only be a couple of them active at once, depending on the user's full body tracking setup.

Currently the Lua API for skeletons is kind of fragmented.  There is `lovr.headset.getSkeleton`, which returns joints for a particular device, and you can pass in `left` or `right`.  It's effectively just used for hand tracking.

Separately, there are some extra devices for full body tracking: `elbow/*`, `shoulder/*`, `chest`, `waist`, `knee/*`, and `foot/*`.  These mostly exist because they match the roles used by Vive trackers.  Ultraleap hand tracking also returns elbow positions, and we map those to the `elbow/*` devices as well.

So LÖVR is kind of in a weird place where it seems like `lovr.headset.getSkeleton` should return a full body skeleton, but there are *also* all these "devices" that contain bits of the skeleton pose as well.  We should probably just pick one going forward: either remove devices from getSkeleton and have `.getSkeleton()` just return as much full-body skeletal info as possible in a giant table (including hand tracking), or lean into semantic path device poses for the entire body, possibly even converting the hand tracking joints into Devices.

Semantic paths cause some problems for the API design since they are also used for all the button/axis input, velocity, controller models, and haptics.  Are we ever going to have button inputs on our knees?  `lovr.headset.isDown('knee/left', 'bend')`???  Ask for a 3D model of someone's metacarpal bone?  On the other hand, I could see joint velocity and joint haptics potentially becoming a thing, and the Device methods do provide a nice way to wrap all that up.  There is maybe an opportunity to separate "pose devices" (hand/point, hand/grip, hand/palm, hand/ring/tip, elbow/left, etc.) from "input devices" (left/right controllers, hand tracking gestures, stylus, headset), I need to think about it more.

gareth-morgan-nv:
Adds body tracking to the LOVR LUA interface.  It follows the handTracking pattern and adds a `body` device that can be passed to getSkeleton:

```
    local skeleton = lovr.headset.getSkeleton('body')
    if skeleton then
      local pelvis = skeleton[1]   -- pelvis
      local head = skeleton[16]    -- head

```
Internally uses the [XR_BD_body_tracking](https://registry.khronos.org/OpenXR/specs/1.1/html/xrspec.html#XR_BD_body_tracking) extension, and exposes the skeleton via its joint conventions, but could use any of the other OXR body tracking extensions:
image

An example LUA script using it is here:
 https://gist.github.com/gareth-morgan-nv/5218bdb91b47ac956edea82b54fe063a


bjornbytes:
Thanks for the PR, this is exciting!

- I'm curious about the reason you chose `XR_BD_body_tracking` over the other body tracking extensions.
- Have you tried this out on a specific device/runtime combination?  I don't think I have access to any hardware that supports this extension atm, so it would be useful to know where it's been confirmed to work.
- I will leave some more feedback on the code, but upfront I will say that there are way, way too many comments, relative to the typical style used in the rest of the codebase.  (Some comments are fine!  But we try to reserve them for the most important things to call out, or to break up very long functions/files into sections).
- Minor note: I do think it's pretty awkward to have "devices" for body parts, e.g. `elbow/left`, `chest`, and now even `body`.  I think in the future `lovr.headset.getSkeleton` will be changed to take a `SkeletonType` (either `hand/left`, `hand/right`, or `body`), and then all the body-related devices could be removed and folded into getSkeleton.  But this is orthogonal to this immediate PR, and can happen in the future.  There's some more discussion in #840.

src/modules/headset/headset.h
```diff
@@ -171,6 +174,33 @@ typedef enum {
...
+  BODY_JOINT_HEAD,
+  BODY_JOINT_LEFT_SHOULDER,
+  BODY_JOINT_RIGHT_SHOULDER,
+  BODY_JOINT_LEFT_ELBOW,
+  BODY_JOINT_RIGHT_ELBOW,
+  BODY_JOINT_LEFT_WRIST,
+  BODY_JOINT_RIGHT_WRIST,
+  BODY_JOINT_LEFT_HAND,
+  BODY_JOINT_RIGHT_HAND
+} BodyJoint;
```
bjornbytes: This is currently unused, right?  It could maybe be removed (same goes for `HandJoint` too, actually, I don't remember why that's there).

src/modules/headset/headset.c
```diff
@@ -4170,6 +4290,27 @@ static XrHandTrackerEXT getHandTracker(Device device) {
   return *tracker;
 }
 
+// Body tracker is created lazily because on some implementations xrCreateBodyTrackerBD
+// will return XR_ERROR_FEATURE_UNSUPPORTED if called too early.
```
bjornbytes: Is this actually true?  It seems like this was just copied from `getHandTracker`.  It would be nice to create the body tracker upfront (e.g. in `lovrHeadsetStart`), if possible.

src/modules/headset/headset.c
```diff
@@ -2690,9 +2810,9 @@ bool lovrHeadsetGetSkeleton(Device device, float* poses, SkeletonSource* source)
     pose += 8;
   }
 
-  if (state.extensions.handTrackingDataSource) {
+  if (state.extensions.handTrackingDataSource && source) {
     *source = sourceState.dataSource == XR_HAND_TRACKING_DATA_SOURCE_CONTROLLER_EXT ? SOURCE_CONTROLLER : SOURCE_HAND;
-  } else {
+  } else if (source) {
```
bjornbytes: Minor, bhere isn't really a need to check `source` for nullness, we control all callers and it isn't currently convenient for them to omit the `source` pointer.

src/modules/headset/headset.c
```diff
@@ -2638,6 +2655,109 @@ bool lovrHeadsetGetAxis(Device device, DeviceAxis axis, float* value) {
...
+    for (uint32_t i = 0; i < BODY_JOINT_COUNT; i++) {
+      if (joints[i].locationFlags & (XR_SPACE_LOCATION_POSITION_VALID_BIT | XR_SPACE_LOCATION_ORIENTATION_VALID_BIT)) {
+        anyTracked = true;
+        break;
+      }
+    }
+
+    if (!anyTracked) {
+      return false;
+    }
```
bjornbytes: Would slightly prefer to have a single loop that unconditionally writes joint poses and computes `anyTracked`, like this:

```c
bool anyTracked = false;
float* pose = poses;
for (uint32_t i = 0; i < BODY_JOINT_COUNT; i++) {
  memcpy(pose + 0, position)
  memcpy(pose + 4, orientation)
  anyTracked |= !!(joint[i].locationFlags & VALID);
}

return anyTracked;
```

It just simplifies things a bit.

src/modules/headset/headset.c
```diff
@@ -2638,6 +2655,109 @@ bool lovrHeadsetGetAxis(Device device, DeviceAxis axis, float* value) {
...
+  //  16    RIGHT_ARM_LOWER           RIGHT_ARM_UPPER
+  //  17    RIGHT_HAND_WRIST_TWIST    RIGHT_ARM_LOWER
+  //  18    LEFT_HIP                  HIPS
+  //  19    LEFT_LEG_UPPER            LEFT_HIP
+  //  20    LEFT_LEG_LOWER            LEFT_LEG_UPPER
+  //  21    LEFT_FOOT_ANKLE_TWIST     LEFT_LEG_LOWER
+  //  22    RIGHT_HIP                 HIPS
+  //  23    RIGHT_LEG_UPPER           RIGHT_HIP
+  //  24    RIGHT_LEG_LOWER           RIGHT_LEG_UPPER
+  //  25    RIGHT_FOOT_ANKLE_TWIST    RIGHT_LEG_LOWER
```
bjornbytes: I don't really think it's necessary to include a table here, a link to the spec is sufficient (the LÖVR docs can include a nice table/picture too, for people using the Lua API).

src/modules/headset/headset.h
gareth-morgan-nv: Removed, I kept the hand tracking one.

gareth-morgan-nv:
>     * I'm curious about the reason you chose `XR_BD_body_tracking` over the other body tracking extensions.
It's the hardware I had, also its a simpler extension that the other ones (no confidence, base pose, etc.)

>     * Have you tried this out on a specific device/runtime combination?  I don't think I have access to any hardware that supports this extension atm, so it would be useful to know where it's been confirmed to work.
We have confirmed this works with an experimental version of CloudXR

>     * I will leave some more feedback on the code, but upfront I will say that there are way, way too many comments, relative to the typical style used in the rest of the codebase.  (Some comments are fine!  But we try to reserve them for the most important things to call out, or to break up very long functions/files into sections).
I've done a pass removing them to match the rest of the code, does that look OK?

>     * Minor note: I do think it's pretty awkward to have "devices" for body parts, e.g. `elbow/left`, `chest`, and now even `body`.  I think in the future `lovr.headset.getSkeleton` will be changed to take a `SkeletonType` (either `hand/left`, `hand/right`, or `body`), and then all the body-related devices could be removed and folded into getSkeleton.  But this is orthogonal to this immediate PR, and can happen in the future.  There's some more discussion in [Body Tracking #840](https://github.com/bjornbytes/lovr/issues/840).
That makes sense.

bjornbytes:
Thanks, merging!

PR #926, opened 2026-02-02
Fix world to collider local in shape containsPoint

jmiskovic:
I noticed that shape attached to collider does incorrect query on :containsPoint().

containsPoint

I think the order of translation/rotation was wrong, but since inverseTransformPoint is used twice it was easier to switch to lovrColliderGetLocalPoint. The added benefit is that it will work better with motion interpolation.

Minimal reproduction:
```Lua

local c = world:newBoxCollider(1, 0, 0, 1, 4, 1)
c:setOrientation(math.pi/2, 0,0,1)

print(c:getShape():containsPoint(2, 0, 0)) -- should be true
```

I also checked `Shape:raycast()`, that one gives correct results.

### DIFF
--- a/src/modules/physics/physics.c
+++ b/src/modules/physics/physics.c
@@ -2104,9 +2104,9 @@ static void inverseTransformRay(float* origin, float* direction, float* position
 
 bool lovrShapeContainsPoint(Shape* shape, float point[3]) {
   if (shape->collider) {
-    float position[3], orientation[4];
-    lovrColliderGetPose(shape->collider, position, orientation);
-    inverseTransformPoint(point, position, orientation);
+    float local[3];
+    lovrColliderGetLocalPoint(shape->collider, point, local);
+    vec3_init(point, local);
   }
 
   inverseTransformPoint(point, shape->translation, shape->rotation);


bjornbytes:
I think `inverseTransformPoint` was just broken, it should be this to match `inverseTransformRay`, right?  (translate, then rotate)

```diff
diff --git src/modules/physics/physics.c src/modules/physics/physics.c
index b17b08a4..74a46368 100644
--- src/modules/physics/physics.c
+++ src/modules/physics/physics.c
@@ -2090,8 +2090,8 @@ void lovrShapeGetAABB(Shape* shape, float aabb[6]) {
 static void inverseTransformPoint(float* point, float* position, float* orientation) {
   float inverse[4];
   quat_conjugate(quat_init(inverse, orientation));
-  quat_rotate(inverse, point);
   vec3_sub(point, position);
+  quat_rotate(inverse, point);
 }
 
 static void inverseTransformRay(float* origin, float* direction, float* position, float* orientation) {
```

Also, it looks like all the `getLocalPoint/WorldPoint/LocalVector/WorldVector` don't work with interpolation, whereas `getPose` does.

jmiskovic:
You're right. I've added to my test the shape offset with rotation and that broke down my fix.

containsPoint
Your version works as expected, aside from inertia wonkiness when shape is off center.  

Interpolation is a separate issue I guess.

bjornbytes:
Ok, I applied the diff from the comment.  Thanks!

Issue #892, opened 2025-09-23
Atomic Audio Functions


bjornbytes:
It's useful to make some changes to sources and know that they'll all be executed atomically.  A common example might be playing a bunch of sources and having them start at the same time, but you also might want to change the volume of a collection of sources, or adjust some effects.

Currently this isn't possible because each individual Source method takes out a lock.  It's possible for the audio thread to start playing a source in between any of these calls, so you can't guarantee that the source is in the state you want when it starts playing.

LÖVR could add `lovr.audio.play`, `lovr.audio.pause`, and `lovr.audio.stop` which take a collection of sources, ensuring they are all changed atomically.  There are a few issues with it:

- `stop` currently has a naming conflict with the existing `lovr.audio.stop` function, which stops an audio device.  The variants could just overlap, but it's probably too confusing.
- It only lets you adjust playback state, it doesn't let you make other changes to the sources.

There's another idea I've had called like "audio transactions".  Basically, *all* of the audio state would be double buffered.  Lua code could modify one copy while the audio thread is using the other copy, and at explicit points the states could be swapped, allowing the audio thread to see new changes.

You could imagine a `lovr.audio.commit` function that swaps the buffers and lets the audio thread see any changes to sources.  boot.lua could call this after `lovr.update`, so that you don't have to think about flushing changes manually.  You could also call this explicitly after making some changes to sources, for slightly better responsiveness.

jeveloper:
Double buffering seems like a good approach. :)

Issue #866, opened 2025-04-24
Add dynamic audio geometry support


Flameborn:
Hello,

As the title says, it would be great to have dynamic geometry support for game audio, so that moving objects can be properly spatialised within a created scene. See the guide at https://valvesoftware.github.io/steam-audio/doc/capi/guide.html#dynamic-geometry

bjornbytes:
Agreed, I'll try to add this when I upgrade phonon to v4 (hopefully soon)

Flameborn:
Very much appreciated, this is the only missing feature that prevents me from using Lövr exclusively for immersive audio game development.

bjornbytes:
There's an `AudioMesh` object now that supports `:setTransform` and `:setEnabled`.

Flameborn:
This is fantastic, thank you very much!

Issue #474, opened 2021-08-16
Upgrade phonon to v4


bjornbytes:
Happening on `phonon-v4` branch.

bjornbytes:
phonon significantly restructured the way sources are processed during spatialization.  Basically they all get processed in a batch now instead of individually.

- Sources are created and can be added/removed from a "simulator".
- Every audio frame, you set the inputs for each source (transform, effects, settings, etc.).
- Then you call `iplSimulatorRunDirect` or `iplSimulatorRunReflections` which processes all sources attached to the simulator.
- Then you retrieve outputs for each source, which has all the attenuation values, and reverb info.
- These outputs get fed to "effects" which do the actual mixing.

It's a bit different than LÖVR's current spatializer structure, which processes sources one at a time (plus the "tail" for reverb).  There may be an opportunity to change the API to process things in a batch, which would match phonon better and might allow for reduced overhead.  Not sure what it would look like yet.

Issue #409, opened 2021-04-12
Per-triangle AudioMaterials


bjornbytes:
Phonon lets you have a material for each triangle of a mesh, but right now we only let you set one material for the whole mesh.

Similarly, when using Oculus' box model, you can set a material for each face of the box.

The AudioMaterial passed to `lovr.audio.setGeometry` should probably optionally accept a table of AudioMaterials.

PR #928, opened 2026-02-05
Super Audio Upgrade!

bjornbytes:
Summary of changes:

- Enables Steam Audio by default in all LÖVR builds!
- Upgrades Steam Audio from version 2 to version 4.8 (#474)
- Add a new `AudioMesh` object, with dynamic transform (#866, #406)
- Audio meshes can have per-triangle audio materials (#409).
- Adds `lovr.audio.setHRTF` to enable a custom HRTF.  Binaural spatialization will only work if an HRTF has been set, otherwise the spatializer will fall back to simple panning.
- Most changes to sources are queued up and committed when `lovr.audio.update` is called (`lovr.run` does it for you after `lovr.update`) (#892).
- Add `t.audio.debug` to get log messages from Steam Audio and miniaudio.
- Audio thread is lock-free, less risk of stalling/underruns
- Lots of extra conf.lua options to customize reverb
- Heavy reverb computation is performed on a background thread asynchronously.
- Replace `Source:setEffectEnabled` with individual setters, that allow extra customization:
  - `Source:setAbsorption` -- sources can have custom air absorption coefficients
  - `Source:setCone` -- uses more intuitive angle cone, replaces setDirectivity
  - `Source:setFalloff` -- has configurable min distance and volume limit
  - `Source:setOcclusion` -- can configure number of rays used for occlusion/transmission
  - `Source:setReverb` -- can set reverb volume and choose listener-centric vs. source-centric reverb mode
  - `Source:setSpatialization` -- can blend between spatialized and non-spatialized audio

TODO:

- [x] Add support for ambisonic sound playback with binaural spatialization and rotation


jeveloper:
@bjornbytes  Bjorn, as a sound designer/artist/long time coder , I am applauding for your hard work and I would like to work on using your audio system in the 2026 release and write a few articles about it along the way.

PR #929, opened 2026-02-06
Pass:setQuad and UV2 attribute

bjornbytes:
This tries to solve 2 problems:

- Make it easier to draw a sub-rectangle of a texture.  A frequent pain point is that this currently requires either creating a new `Material` object for each new rectangle, or writing a custom shader.
- Add support for a second set of UVs when importing models, initially attempted in #860

#### Background

We'd like to switch to 16 bit unsigned normalized UVs to save space and keep model vertices at 32 bytes after we add the second UV set.  However, this would mean that we could only support UVs that go from 0-1, which while very common, isn't an acceptable restriction.

To solve this, the model importer normalizes UV data to the 0-1 range and stores the original range of the UVs for each part.  That way it's possible to take a compressed 16-bit 0-1 UV and reconstruct the original UV, even for floating point UVs.  There is a minor precision loss.

However, we need to send this UV rectangle to shaders somehow.  It would be possible to use Material for this, since it already has a UV rectangle, but it would get complicated if 2 meshes shared a material, and we would need to duplicate materials to fix it.

#### Solution

Another solution is to just add per-draw data for a UV rectangle, which 1) makes it possible for models to scale and bias their compressed UVs, and 2) provides a convenient way of adjusting texture UVs on a per-draw basis.  This was added as `Pass:setQuad(x, y, w, h)`.  This is intended to replace Material's uvShift and uvScale properties, eventually.

Note that the quad only applies to the first UV set -- the second set of UVs are not transformed.  UV2 is intended to be used for lightmapping, and it's rare for lightmap UVs to go outside of 0-1.

#### Drawbacks

- This unconditionally adds 16 bytes of data to every draw, raising the total from 64 to 80 bytes, and adds a new uniform buffer.  It may be possible to eliminate this overhead for passes that don't use quads.
- While the precision issues are very minor, they could still cause problems.
- If people are drawing ModelData manually, they now have to remember to plumb through the quad, if they want to support UVs outside 0-1 range.
- The `VertexUV` data isn't the true UV anymore (but, like, normals/tangents are compressed too).

Other stuff:

- `:setQuad` naming
- Units (0-1 uvs vs. pixels, if you know the texture)
- Could add quad as optional arguments after texture/material in :setMaterial


src/modules/data/modelData_gltf.c
```diff
@@ -1129,13 +1161,27 @@ bool lovrModelDataInitGltf(ModelData** result, Blob* source, ModelDataIO* io) {
...
               uint32_t vertexCount = positions->count;
               ModelVertex* vertices = model->vertices + vertexOffset;
-              copyAttribute(vertices, positions, F32, 3, false, 0, sizeof(ModelVertex), vertexCount, 0);
-              copyAttribute(vertices, normals, SN10x3, 1, false, 12, sizeof(ModelVertex), vertexCount, 0);
-              copyAttribute(vertices, uvs, F32, 2, false, 16, sizeof(ModelVertex), vertexCount, 0);
-              copyAttribute(vertices, colors, U8, 4, true, 24, sizeof(ModelVertex), vertexCount, 0xff);
-              copyAttribute(vertices, tangents, SN10x3, 1, false, 28, sizeof(ModelVertex), vertexCount, 0);
+              copyAttribute(vertices, positions, F32, 3, false, 0, sizeof(ModelVertex), vertexCount, 0, false);
+              copyAttribute(vertices, normals, SN10x3, 1, false, 12, sizeof(ModelVertex), vertexCount, 0, false);
+              copyAttribute(vertices, uvs, U16, 2, true, 16, sizeof(ModelVertex), vertexCount, 0, true);
```
bjornbytes, line 1181: Maybe we should only "supernormalize" if the UVs are actually outside of the 0-1 range.  It gives worse precision, but means that the UV data can be used as-is in more cases (quad is optional more often)

src/modules/data/modelData.h
```diff
@@ -43,6 +44,7 @@ typedef struct {
   uint32_t material;
   ModelDrawMode mode;
   float bounds[6];
+  float quad[4];
```
bjornbytes, line 47: Still need to expose this in `ModelData`

src/modules/graphics/graphics.c
```diff
@@ -8749,6 +8791,7 @@ bool lovrPassDrawPart(Pass* pass, Model* model, uint32_t meshIndex, uint32_t par
       .material = model->materials && part->material != ~0u ? model->materials[part->material] : NULL,
       .transform = transform, // TODO fix skinned mesh transforms?
       .bounds = part->bounds,
+      .quad = part->quad,
```
bjornbytes, line 8794: Worth passing NULL here (and in drawNode) if we know the part doesn't have a quad?

src/modules/data/modelData_gltf.c
```diff
@@ -285,7 +285,25 @@ static uint32_t typeSizes[] = {
...
+  if (!accessor || (accessor->hasMin && accessor->hasMax)) {
+    return;
+  }
+
+  char* data = accessor->data;
+  size_t stride = accessor->stride;
+  if (accessor->type == F32 && accessor->components == 2) {
+    for (uint32_t i = 0; i < accessor->count; i++, data += stride) {
+      float* f = (float*) data;
+      accessor->min[0] = MIN(accessor->min[0], f[0]);
```
bjornbytes, line 298: Should initialize the min/max to FLT_MAX/-FLT_MAX

nensanders:
Interesting. I use the uvMatrix technique to to map planetary UVs to quadtree subquads. But for me I need the full resolution and can't even afford to compress the normalized UVs. 
I don't really see that the vertex memory bandwidth is the bottleneck on current standalone headsets. The problem is much more in the fragment stages. I assume people who need this levels of customization are already well equipped to solve for their exact use case. (That's why I wanted to provide my example ;-)

bjornbytes:
Hm, so you need vertices outside the 0-1 range and with more than 65536 unique values at the vertex positions (GPU will still interpolate with 32-bit floats no matter what)?

Enough people have raised concerns here that we might need to add a way to pick between 1 set of 32-bit UVs and 2 sets of 16-bit UVs when importing a Model.

I remember seeing something like a 5-10% difference on GPU times when switching between u16 and u32 indices on Quest, which was surprising, so I don't totally want to write off vertex/index buffer bandwidth.  There are other reasons for wanting to keep power of 2 sizes for the vertex format too: it makes it easier to pack stuff together without padding.

bjornbytes:
I decided to just live with 36 byte model vertices for now.  Added UV2 in 6f60f272f7d2b383158686d65128af7a417515a2.

In the future, I'd like to make a change to support dynamic vertex formats for models.  First, `lovr.data.newModelData` can support compression options:

```lua
lovr.data.newModelData('gltf', {
  compress = {
    position = true, -- un16x4, bake scale/bias into node transforms
    normal = true, -- existing 10 bit compression
    uv1 = true, -- 16 bit compression
    uv2 = true, -- 16 bit compression
    color = true, -- 8 bit srgb
    tangent = true -- existing 10 bit compression
  }
})
```

These can be left as nil and LÖVR will generally do the right thing based on what's in the model.

Also, ModelData will only store data for attributes that are actually in the asset.  If a model doesn't have any tangents, they won't be imported/stored.

Internally these will be tracked with a bunch of flags, like `COMPRESS_POSITIONS`, `HAS_COLORS`, etc.

A single Model still has to use a consistent, lowest-common-denominator format, but different Models can use different vertex formats.

Then, we'll make the following changes to support dynamic vertex formats:

- `ModelData:getMeshVertex` probably still massages everything into a consistent format, returning everything as doubles, returning empty data for missing attributes, etc.
- `Model` builds its own vertex format for its vertex buffer based on the compression/attribute flags.  The renderer already supports buffers with arbitrary vertex formats, so nothing needs to change about the draw calls.
- For the animation compute shaders, we can probably pass some simple vertex format information like the offset of each attribute and whether it's compressed.  That way we can continue to do vertex pulling.  Hopefully isn't too bad.
- A few other stragglers like raytracer builds and GetTriangles might need to incorporate the vertex stride or a position conversion.

This will allow people to customize how model vertices are imported/compressed, and also generally it will greatly reduce the size of model vertices to the minimum amount of stuff that's necessary.  e.g. if a Model only has positions and normals, and is fine with un16x4 and sn10x3 compression for those, it'll only use 12 bytes.

Having a bunch of different vertex formats will increase the amount of pipeline switches, but people can alleviate themselves by passing a consistent set of compression flags when creating their models.

`Pass:setQuad` is deferred for now, but it still seems like a good idea.  It might be easier to do it along with other changes to Material though.

PR #936, opened 2026-03-13
Fix init ordering in Source:clone() for phonon

jmiskovic:
`source:clone()` crashes when LÖVR is built with phonon. Fix order of init operations.


bjornbytes:
Thanks!  I still have more audio testing to do...

PR #937, opened 2026-03-19
Emulated Dpad Buttons

bjornbytes:
Some improvements to #919.

- Adds new `DeviceButton` values for virtual dpads:
  - `thumbstick/up`
  - `thumbstick/down`
  - `thumbstick/left`
  - `thumbstick/right`
  - `touchpad/up`
  - `touchpad/down`
  - `touchpad/left`
  - `touchpad/right`
  - `touchpad/center`
- Adds support for `XR_EXT_dpad_binding`.  Since this extension has poor runtime support, emulate it if it's unsupported.
- Maps hand tracking microgestures to touchpad dpad buttons instead of the existing dpad buttons (so that the center thumb tap can be exposed...).
  - Not sure about this.  Other options might be:
    - Add `dpad/center` and keep it mapped to "dpad".
    - Add its own `swipe/` namespace with `swipe/up`, `swipe/down`, etc.
      - For the central thumb tap: `swipe/tap`?  Or a top-level `thumbtap` button?
- Rename `dpup`, `dpdown`, `dpleft`, `dpright` to `dpad/left`, etc. for consistency.

FWIW I'm still kinda on the fence on exposing these at all, the feature seems a little too high level and can be derived from the axis values pretty easily.  On the other hand, OpenXR does provide it, it would reduce the verbosity of the input code in the new error screen slightly, and mapping microgestures onto a directional touchpad kinda makes sense.  Wanted to sketch out the idea while I think about it more, and allow others to provide feedback.


bjornbytes:
Not exposing emulated dpads for now.

PR #941, opened 2026-05-15
Add no-simulation argument

gareth-morgan-nv:
Disables the simulation option and exits when OXR creation fails if its set. This allows LOVR to be run in automation and detect the FORM_FACTOR_UNAVAILABLE discussed below without a custom LUA script.

### DIFF
--- a/etc/boot.lua
+++ b/etc/boot.lua
@@ -41,6 +41,7 @@ local conf = {
   headset = {
     connect = true,
     start = true,
+    simulatorFallback = true,
     debug = false,
     seated = false,
     mask = true,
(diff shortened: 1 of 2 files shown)


bjornbytes:
If the form factor is unavailable, `lovr.headset.connect` will loop infinitely and never return.  This would prevent apps from calling `lovr.headset.connect` once to check if VR is available (potentially falling back to flatscreen or otherwise showing some kind of error).

Maybe you could do something like this in Lua?

```lua
while not lovr.headset.connect() do
  print('Failed to set up headset, retrying after 1s...')
  lovr.timer.sleep(1)
end
```

You could also use the second return value of `lovr.headset.connect` to specifically match on 'xrGetSystem' or 'FORM_FACTOR_UNAVAILABLE'.

gareth-morgan-nv:
> Maybe you could do something like this in Lua?


Though we have the issue we are running in automation with an arbritary LUA script, so we can't detect the case where the OXR init has failed and its gone into simulation.  How about this?  I added a --no-simulation option that disables simulation and just exits if OXR fails.

bjornbytes:
Closing in favor of #946

PR #944, opened 2026-05-20
Turn Luau into a shared library, make sure all of the libraries are moved correctly

TylkoKuba:
Luau was not built as a shared library, as a result all the plugins instead of linking against it were building against it, additionally move_lib and move_dll function had to be changed as Luau is composed of "Luau.VM", "Luau.Compiler", "Luau.Ast" and "Luau.Common" (which was missing, but seems to be required). I've discovered that when trying to use enet together with Luau version of Lovr which now functions properly :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -686,7 +686,7 @@ if(WIN32)
 
   function(move_dll)
     foreach(LIB IN ITEMS ${ARGV0})
-      if(TARGET ${ARGV0})
+      if(TARGET ${LIB})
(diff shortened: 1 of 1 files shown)


bjornbytes:
Sorry this fell through the cracks so long.  Current status: I think this is a good change, I was just annoyed that Luau creates so many libraries when you tell it to build shared libraries.  I tried to get Luau's CMake to produce 1 library file but didn't have any luck, might need to modify lovr's fork to do it.  But still, it's kind of a separate issue from the fix here, so I should probably merge it.

PR #945, opened 2026-05-20
fix runtime asan error due to storing to and reading from misaligned address

TylkoKuba:
```lua
local blob = lovr.data.newBlob(3)
blob:setU8(0, 0)
blob:setU16(1, 0)
blob:getU8(0)
blob:getU16(1)
```

Results in
`runtime error: store to misaligned address [...] for type 'uint16_t' (aka 'unsigned short'), which requires 2 byte alignment`
and
`runtime error: load of misaligned address [...] for type 'const uint16_t' (aka 'const unsigned short'), which requires 2 byte alignment`

### DIFF
--- a/src/api/l_data_blob.c
+++ b/src/api/l_data_blob.c
@@ -1,6 +1,7 @@
 #include "api.h"
 #include "data/blob.h"
 #include "util.h"
+#include <string.h>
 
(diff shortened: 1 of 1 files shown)


bjornbytes:
Thanks!

PR #946, opened 2026-05-20
Optionally retry on FORM_FACTOR_UNAVAILABLE

gareth-morgan-nv:
Apologies that we are going back and forth on this.   It turns out implementing a retry loop in the LUA gets really complicated due to the graphics init order.  The amount of boiler plate code required just to connect the headset seems excessive, as shown in this Gist:
https://gist.github.com/gareth-morgan-nv/82d9e3b242146f8028e4d60af73c873c

So this PR adds the same retry loop we pushed in the last PR, but this time its behind a command line flag which is false by default.

### DIFF
--- a/etc/boot.lua
+++ b/etc/boot.lua
@@ -41,6 +41,7 @@ local conf = {
   headset = {
     connect = true,
     start = true,
+    retryonunavailable = false,
     debug = false,
     seated = false,
     mask = true,
--- a/etc/nogame/arg.lua
+++ b/etc/nogame/arg.lua
@@ -5,6 +5,7 @@ function lovr.arg(arg)
     console = { long = '--console', help = 'Attach Windows console' },
     debug = { long = '--debug', help = 'Enable debugging checks and logging' },
     simulator = { long = '--simulator', help = 'Force headset simulator' },
+    retryonunavailable = { long = '--retry-on-unavailable', help = 'Retry xrGetSystem when the headset is temporarily unavailable' },
     watch = { short = '-w', long = '--watch', help = 'Watch files and restart on change' }
   }
 
@@ -80,6 +81,10 @@ function lovr.arg(arg)
     if arg.simulator then
       conf.headset.connect = false
       conf.headset.start = false
+    end
+
+    if arg.retryonunavailable then
+      conf.headset.retryonunavailable = true
     end
 
     if arg.watch then
--- a/src/api/l_headset.c
+++ b/src/api/l_headset.c
@@ -1091,6 +1091,7 @@ int luaopen_lovr_headset(lua_State* L) {
 
   HeadsetConfig config = {
     .supersample = 1.f,
+    .retryUnavailable = false,
     .seated = false,
     .mask = true,
     .stencil = false,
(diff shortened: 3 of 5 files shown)


bjornbytes:
Yeah, the boilerplate is unfortunate, but it kinda falls out of the complexity of `XR_KHR_vulkan_enable2`.

If it's too much to stomach, one thing you might be able to do is put the `lovr.headset.connect` retry loop in conf.lua.  This would let the rest of the boot process continue normally, compared to disabling headset/graphics and reimplementing a huge chunk of the boot logic yourself.  I haven't tested it yet, but something like this:

```lua
-- conf.lua
function lovr.conf(t)
  -- ...fill out conf table

  -- Wait until headset is connected before booting lovr!
  local headset = require('lovr.headset') -- or lovr.headset = require 'lovr.headset' is fine too
  local timer = require('lovr.timer')
  while not headset.connect() do
    timer.sleep(1)
  end
end
```

This is still pretty awkward, but it's got a smaller surface area.

(I'm maybe seeing an issue where requiring modules in conf.lua will cause them to not pick up the conf table properly, but that could be fixed)

Regarding the change here (optional flag to enable a retry loop in `xrGetSystem`), my impression is that it still feels a little too "opinionated" or "specific", but I'm still giving it thought.

gareth-morgan-nv:
> Regarding the change here (optional flag to enable a retry loop in `xrGetSystem`), my impression is that it still feels a little too "opinionated" or "specific", but I'm still giving it thought.

Any more thoughts on this?  Anything suggestions to make it less opinionated?  Note this is based on the [spec](https://registry.khronos.org/OpenXR/specs/1.1/man/html/xrGetSystem.html) which says that error can happen temporarily e.g.  because  "connecting or warming up hardware might cause an unavailable form factor to become available."

bjornbytes:
Is there a reason the "wait for connect in conf.lua" approach above doesn't work for you?  It doesn't require reimplementing as much of the boot process, I've confirmed it to be working, and it doesn't require modifying the API.

Beyond that, there's a spectrum of things from this PR that could get added:

- `lovr.headset.connect` takes a `wait` boolean that will do the wait if `xrGetSystem` returns FORM_FACTOR_UNAVAILABLE.  (Or added to `HeadsetConfig` like in this diff).
  - If you called this in main.lua, you'd have to do the graphics setup yourself.
  - If you called this in conf.lua, you wouldn't.
- Add a `t.headset.wait` option to conf.lua, that gets passed to the `lovr.headset.connect` call.
  - More convenient, don't need to call `lovr.headset.connect` yourself.  But you still need to add a conf.lua option.
- Add a `--wait-for-headset` command line option that sets the conf option.  Most convenient, now you don't need to change conf.lua to get the waiting behavior.

It might help to share more about your use case (automated tests?  CloudXR needs to wait to establish a network connection?  do you have a test harness?  can you change conf.lua?) to help inform the design.

Also, as an aside: one thing I didn't realize at first is that `XR_FORM_FACTOR_UNAVAILABLE` is different from `XR_FORM_FACTOR_UNSUPPORTED`.  Still, blocking for `xrGetSystem` to succeed isn't acceptable, since I think most people want LÖVR to fall back to the simulator when their headset is unplugged (instead of just hanging).

PR #970, opened 2026-06-27
Dynamic Resolution

### LINKED ISSUE #956: OpenXR Dynamic Resolution
Use `XR_META_recommended_layer_resolution` to dynamically adjust the headset render resolution, and potentially the resolution of a `Layer`.

- Is this mutually exclusive with `t.headset.supersample`?
- Enabled by default?  Probably
- Does it use viewport/scissor, or texture aliasing?  Viewport/scissor is simpler, but complicates render pipelines that want to sample the rendered result afterwards (need to be aware of the render area).
- Do runtimes report an accurate recommended layer resolution for quad/cylinder layers, or is this a joke

bjornbytes:
Implements dynamic resolution in `lovr.headset`, so the render resolution will scale automatically based on performance to help keep the frame rate high/consistent.

This uses two different OpenXR extensions:

- `XR_EXT_view_configuration_views_change` is an event that the runtime sends if the user's preferred rendering resolution changes, like the "render scale" slider in SteamVR settings.  LÖVR listens to this event now and resizes the headset texture as the slider moves.  (This was a long-standing feature gap between OpenXR and OpenVR).
- `XR_META_recommended_layer_resolution` is "real" dynamic resolution.  LÖVR asks the runtime for a recommended resolution every frame, and it scales dynamically with performance.

Other notes:

- You can control whether dynamic resolution is active with `t.headset.dynamicresolution`, which defaults to true.
- This probably replaces/deprecates `t.headset.supersample`, although currently I have it so the supersample factor is still applied if neither of the extensions are supported.  Maybe it's still useful for the simulator?  It would be nice to reduce the amount of moving parts though.
- `lovr.headset.getDisplayDimensions` returns the dynamic resolution.
- The dynamic resolution updates when `lovr.headset.update` is called.
- The Pass you get from `lovr.headset.getPass` has its viewport/scissor set to the dynamic resolution.
- I have to do some lovely hacks in `lovr.mirror` to render the sub-rectangle of the headset texture.  `Pass:setQuad` can't come soon enough.
- In general it's a huge pain to use the headset texture for anything afterwards, since you have to lug around the viewport size any time you want to sample the texture (as shown in lovr.mirror).  In situations where this is too annoying, you could disable dynamic resolution.

Is this worth it?  I guess so.  It's better than setting `t.headset.supersample` to something arbitrary and hoping for the best.

Closes #956


etc/boot.lua
```diff
@@ -47,6 +47,7 @@ local conf = {
     stencil = false,
     antialias = true,
     supersample = false,
+    dynamicResolution = true,
```
bjornbytes: typo (uppercase R)

bjornbytes:
I don't love the `dynamicresolution` name.  It's too long.  `t.headset.supersample = 'auto'`?  `t.headset.autosize`?  `t.headset.renderscale = 'auto' | number`?

bjornbytes:
Ugh, Meta returns `XR_ERROR_SESSION_NOT_READY` if you query the recommended resolution before the session transitions to READY... In that case we disable dynamic resolution.  We'd have to delay swapchain creation to after receiving the READY event, which is gross.  SteamVR doesn't do this and will happily return a recommended resolution at startup.

bjornbytes:
Going to merge this as-is, for now it will only work on SteamVR (not Meta), I have a separate change that might make the swapchain initialization a little more lazy which may improve that.

PR #971, opened 2026-06-30
Fall back to handheld display form factor

hougantc-nvda:
## Summary

`lovrHeadsetConnect` only requests `XR_FORM_FACTOR_HEAD_MOUNTED_DISPLAY` from `xrGetSystem`. Runtimes that expose only a handheld display (`XR_FORM_FACTOR_HANDHELD_DISPLAY`) therefore fail to connect with `XR_ERROR_FORM_FACTOR_UNSUPPORTED`.

This retries with the handheld form factor when the HMD form factor is unsupported, so both form factors work from the same build.

## Notes

- The view configuration selection already includes `XR_VIEW_CONFIGURATION_TYPE_PRIMARY_MONO`, so handheld (single-view) runtimes are picked up automatically — no other changes are needed.
- HMD remains the preferred form factor; the handheld path is only taken when an HMD is not offered.

### DIFF
--- a/src/modules/headset/headset.c
+++ b/src/modules/headset/headset.c
@@ -583,7 +583,14 @@ bool lovrHeadsetConnect(void) {
     .formFactor = XR_FORM_FACTOR_HEAD_MOUNTED_DISPLAY
   };
 
-  XRG(xrGetSystem(state.instance, &systemInfo, &state.system), "xrGetSystem", fail);
+  // Prefer a head-mounted display, but fall back to a handheld display when the
+  // runtime does not offer an HMD, so both form factors are supported.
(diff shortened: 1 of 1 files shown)


bjornbytes:
If you're able to share, I'm curious which hardware/runtime you're using that supports `HANDHELD_DISPLAY`!

hougantc-nvda:
Thanks. The CloudXR platform supports streaming to iOS and iPadOS devices
https://developer.nvidia.com/topics/ai/xr/cloudxr/apple-platforms

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

PR #973, opened 2026-07-02
Multithreaded Texture Upload

bjornbytes:
This is a less ambitious version of #949.

- It converts core/gpu texture API to take CPU pointers instead of GPU buffers, unblocking WebGPU.
- It uses the `VK_EXT_host_image_copy` extension when available, which reduces peak VRAM usage during texture uploads and is supposedly a big win on Android.
- `lovr.graphics.newTexture` does texture uploads on a background thread.  The host image copy (or staging buffer allocation and memcpy to staging buffer) happens on a worker, and `lovr.graphics.submit` will wait until all texture uploads for the frame are complete before submitting.  This does a great job at both:
  - Hiding latency (`newTexture` returns pretty quickly, and you can use a Texture or do other stuff on the CPU right away, even while pixels are still being copied to VRAM), and
  - Enabling parallelism (you can newTexture multiple textures and they will upload in parallel automatically).
- There are no streamed texture uploads, where the GPU can keep rendering frames while the DMA engine uploads textures in the background.  This ended up being less of a win than I'd hoped, there was a lot of code to add support for the transfer queue, and it "caused other performance issues in some drivers".
  - It's also better to wait here, because VK1.4 mandates either host-image-copy or a transfer queue.  For now we have to handle the case where there's no transfer queue and fall back to the graphics queue.
- There are no Lua API changes.

Simple test uploading 10 4K textures in `lovr.load` on an iGPU with 16 threads and no host-image-copy.  Before: 220ms, After: 97ms.

### DIFF
--- a/src/modules/graphics/graphics.c
+++ b/src/modules/graphics/graphics.c
@@ -613,6 +613,8 @@ static struct {
   gpu_pipeline* pipelines;
   atomic_uint pipelineCount;
   atomic_uint glyphJobs;
+  atomic_uint textureUploads;
+  _Atomic(char*) textureUploadError;
   _Atomic(Layout*) layouts;
   Layout* builtinLayout;
   Layout* materialLayout;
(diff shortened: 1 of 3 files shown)


src/modules/graphics/graphics.c
```diff
@@ -2448,6 +2462,49 @@ bool lovrGraphicsGetWindowTexture(Texture** texture) {
   return true;
 }
 
+void lovrTextureUpload(void* arg) {
```
bjornbytes: Should be static

PR #984, opened 2026-07-20
Added GPU selection

InfernoAmaruq:
in lovr.conf, t.graphics.selectGPU takes a function and lets you choose what GPU you want lovr to use

Only for Vulkan

usage example for selecting discrete GPU:
```lua
-- in lovr.conf
t.graphics.selectGPU = function(gpus)
    for Idx, Device in ipairs(gpus) do
        if Device.discrete then return Idx end
    end
    return 1
end
```

Has fields: `.vendor, .name, .id, .discrete, .renderer`
Defaults to GPU at index 1 if no function provided or no return

### DIFF
--- a/src/api/l_graphics.c
+++ b/src/api/l_graphics.c
@@ -379,6 +379,40 @@ static void luax_writeshadercache(void) {
   lovrFree(data);
 }
 
+
+static int callback_selectGPU(void* luaPtr, GraphicsDevice* gd, uint32_t count){
+    lua_State* L = (lua_State*) luaPtr;
+    
+    // main table
+    lua_createtable(L,count,0);
+
+    for (uint32_t i = 0; i < count; i++){
+        lua_newtable(L);
+
+        GraphicsDevice device = gd[i];
+
+        lua_pushinteger(L, device.deviceId), lua_setfield(L, -2, "id");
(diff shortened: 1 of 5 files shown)


bjornbytes:
For this I was thinking of using the `t.graphics.lowpower` option that LÖVE recently added (true = pick first integrated GPU, false = pick first discrete GPU).  Would this work for your use case?  It's a little bit less powerful than the implementation here, but on the other hand people might get tired of writing a function to filter GPUs if they just wanted to use the integrated GPU.

InfernoAmaruq:
Oh that would also be nice. My goal was making it so you can select a specific GPU since sometimes igpu is chosen when a dgpu is present

If lowpower does the same thing then I see it as a win

bjornbytes:
I guess one other thing to consider is that WebGPU only exposes `low-power` and `high-performance` GPU selection, so lowpower could work on WebGPU whereas specific device selection couldn't map onto that.

bjornbytes:
I added `t.graphics.lowpower` in 75d1d365c7feffec88bbc6e01d5a419091997ed3.  I don't have any hybrid graphics hardware on hand to test, but the change is simple enough.  Thanks for helping to push this forward!

PR #986, opened 2026-07-23
Added :setScale to MeshShape and ConvexShape

### LINKED ISSUE #985: :setScale for ConvexShape and MeshShape
Seems like it could be supported?  Create a new `ScaledShape` and use `lovrShapeReplace`?

InfernoAmaruq:
Seems to work when I tested it


bjornbytes:
Thanks!

PR #996, opened 2026-08-03
Added Blob:setData and Blob:clear

### LINKED ISSUE #958: Copying Between Blobs
Ignoring `ffi.copy`, there isn't an easy way to copy between `Blob`s.

jamesthomasgriffin:
This adds a new method `Blob:setData(blob, destinationOffset, sourceOffset, size)` that copies data between blobs and a new method `Blob:clear()` which zeroes out the whole blob.

Note that setData just copies a signature from `Buffer:setData` and that rather than memcpy I used memmove.

In issue #958, it's suggested that an analogue of memset would be worth adding.  I agree but the API wasn't as obvious and might deserve its own issue.

Closes #958


src/api/l_data_blob.c
```diff
@@ -92,6 +92,38 @@ static int l_lovrBlobSetU32(lua_State* L) { l_lovrBlobSet(L, uint32_t); }
...
+  size_t maxSourceSize = sourceBlob->size - (size_t) sourceOffset;
+  size_t maxDestSize = blob->size - (size_t) destOffset;
+  size_t maxSize = maxSourceSize < maxDestSize ? maxSourceSize : maxDestSize;
+
+  lua_Integer size = luaL_optinteger(L, 5, maxSize);
+  luax_check(L, size >= 0, "Size can not be negative");
+  luax_check(L, (size_t) size <= maxSourceSize, "Blob:setData source range overflows the size of the Blob");
+  luax_check(L, (size_t) size <= maxDestSize, "Blob:setData destination range overflows the size of the Blob");
+
+  memmove((char*)blob->data + destOffset, (char*)sourceBlob->data + sourceOffset, (size_t) size);
```
bjornbytes, line 116: The `memcpy` vs `memmove` choice is kind of interesting

- "Overlapping copy ranges isn't the common case, so should prefer memcpy"
- "Performance difference is probably negligible, so just use memmove to be safe/general"
- "Detect if copy ranges overlap and use memmove only when needed" or "Detect if source == dest and use memmove in that case" (as long as aliasing isn't an issue)

I can't really make a super strong case for any of these, though.

jamesthomasgriffin: My feeling is that in the common case, memmove will check the bounds then perform a memcpy.  The check overhead will only matter if performed many times on small ranges in which case a) the branch predictor will recover the time and b) the other checks and any Lua function call overhead will outweigh the memmove check. 

I don't think using an unchecked memcpy is wise.  It may work on some builds/systems and not others leading to tricky to find bugs.  If it's up to the caller and they have to check then they are paying in Lua time.

bjornbytes:
Thanks for the PR!

PR #998, opened 2026-08-04
Whitespace removal in Font:getLines

### LINKED ISSUE #997: Bug in end of line whitespace removal in Font:getLines
The method `Font:getLines` has two bugs:
* an inconsistency that the final line does not have whitespace stripped from it
* the whitespace-stripping code itself does not work and can remove characters it shouldn't

```lua
function lovr.load()
  local font = lovr.graphics.getDefaultFont()
  
  local lines = font:getLines("123   \n      ", 1000)
  for _, line in ipairs(lines) do
    print("Length: ", #line, "Contents:", '"'..line..'"')
  end
  
  -- Outputs:
  --Length:         0       Contents:       ""
  --Length:         6       Contents:       "      "
end
```

jamesthomasgriffin:
This fixes #997 and adds a test.  It might be worth considering whether removing the whitespace at the end of the line is needed, in which case deleting the relevant lines is an alternative.

[EDIT: 6th Aug - I think I understand this better now, the `lovrFontGetVertices` function ignores spaces as they have no vertices and performs alignment ignoring the spaces.  So to match this behaviour `lovrFontGetLines` should remove all whitespace from the end of a line.  I initially missed the bugged while loop that did this for breaking after wrapping at a space, so added an extra commit fixing it and changed the test that was looking for a space when breaking "hello world".]

### DIFF
--- a/src/modules/graphics/graphics.c
+++ b/src/modules/graphics/graphics.c
@@ -4655,7 +4655,7 @@ void lovrFontGetLines(Font* font, ColoredString* strings, uint32_t count, float 
       continue;
     } else if (codepoint == '\n') {
       size_t length = string - lineStart;
-      while (string[length] == ' ' || string[length] == '\t') length--;
+      while (length > 0 && (lineStart[length - 1] == ' ' || lineStart[length - 1] == '\t')) length--;
       callback(context, lineStart, length);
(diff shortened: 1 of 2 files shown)


bjornbytes:
Makes sense, thanks for the fix!

PR #1000, opened 2026-08-07
CJK character wrapping in lovrFontGetLines and lovrFontGetVertices

### LINKED ISSUE #989: Support text wrapping for CJK characters
Pass:text and Font:getLines do not wrap CJK characters, just treating text as long words.  

Punctuation symbols which contain an implicit space don't break words either.

My understanding of intended behaviour is that text should wrap after characters, unless it has been prohibited (e.g. no punctuation at the start of a line) and that information is contained in the font file.

I took a sample font and sample text from Google Fonts.

jamesthomasgriffin:
Resolves #989

## Summary
Adds CJK-aware line wrapping (break at character boundaries for CJK text, suppressed at certain punctuation) without requiring a full shaping library.

## Changes
- new logic adding word breaks around CJK characters
- this logic needs to know the previous codepoint, in the getLines code the variable `previous` was being set to 0 to avoid kerning across linebreaks.  This was interfering with the CJK logic so a new variable `suppressKerning` now does the job instead.
- `suppressKerning` logic also ported to `lovrFontGetVertices`, because it was not easy to see that they were logically the same.
- this also includes a fix for inconsistency between getLines and getVertices where it comes to the wrap parameter, one used `wrap > 0` and the other `wrap >= 0`; I changed both to `wrap > 0`.

### DIFF
--- a/src/modules/graphics/graphics.c
+++ b/src/modules/graphics/graphics.c
@@ -4625,6 +4625,50 @@ float lovrFontGetWidth(Font* font, ColoredString* strings, uint32_t count) {
   return MAX(maxWidth, x) / font->pixelDensity;
 }
 
+static int isCodepointCJK(uint32_t cp) {
+  // There are many classes of characters, many classes are consecutive, which are merged (godbolt suggested that
+  // the compiler might not do this).  Also merged small gaps containing characters unlikely to
+  // form Latin style words.
+  return cp >= 0x01100 && (
+      (cp <= 0x011FF) ||
+      (cp >= 0x02E80 && cp <= 0x0A4CF) ||  // Gap between these are Lisu characters with specific wrapping rules
+      (cp >= 0x0AC00 && cp <= 0x0D7AF) ||  // Gap contains a block of user definable codepoints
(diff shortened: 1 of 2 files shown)


bjornbytes:
I'm a little worried about implementing more advanced line breaking by hand because it creates a precedent for implementing more and more of UAX14.  In a way, it's nice that the line wrapping algorithm is extremely bad right now, because LÖVR isn't making a promise about its text layout capabilities that it can't keep.

For example, now it seems reasonable to request that LÖVR handle line breaking around hyphens and emdashes properly, whereas before this change it felt somewhat out of scope.  But if that gets added, then it would probably make sense to handle all the funny kinds of dashes, spaces, punctuation marks, currency, etc.  At some point, it just makes sense to switch to the text shaping library, since it can do a better job.

However, maybe there is a way to explicitly draw the line and say that LÖVR handles some subset X of the full text shaping features, but nothing more than that.  Maybe that line can be drawn somewhere beyond where it is today, but not in a place that creates extra expectations?  I'm not too familiar with the landscape here.

PR #1001, opened 2026-08-09
Support for shader_explicit_arithmetic_types and shader_16bit_storage

SK83RJOSH:
These two extensions are useful when working with compute shaders, and by extension raster shaders if they're fed packed data.

My specific use case was doing some mesh generation in a compute pass where I needed atomics to write packed data. Using `shader_8bit_storage: require` allows me to remove all atomics by ensuring each byte is only touched by up to at most one thread. So it's a fairly good win, especially considering 8-bit / 16-bit storage has been fairly well supported for the last ~10 years.

Some early feedback received, but not yet address:
1. `vec3` types are not widely supported for vertex attributes, and are not supported by wgpu at all
2. It may be worthwhile merging `storage` / `uniform` features

Please note, I have not tested the wgpu backend yet, but I imagine it works just fine since the changes are fairly mechanical. One thing to call out there is the explicit errors added if you manage to break alignment rules for vertex attributes. Ostensibly this means, you're fine to use types like `u16vec3` if you're not using `scalar` / `packed` rules - the padding saves your bacon. Otherwise, you'll hit these new errors.

### DIFF
--- a/src/core/gpu_vk.c
+++ b/src/core/gpu_vk.c
@@ -244,6 +244,9 @@ typedef struct {
   bool hostImageCopy;
   bool atomicFloat;
   bool calibratedTimestamps;
+  bool float16Int8;
+  bool storage8;
+  bool storage16;
 } gpu_extensions;
 
 // State
@@ -3125,7 +3128,10 @@ bool gpu_init(gpu_config* config) {
       { "VK_EXT_pipeline_creation_cache_control", true, &state.extensions.pipelineCacheControl },
       { "VK_EXT_memory_budget", true, &state.extensions.memoryBudget },
       { "VK_EXT_host_image_copy", true, &state.extensions.hostImageCopy },
-      { "VK_EXT_shader_atomic_float", true, &state.extensions.atomicFloat }
+      { "VK_EXT_shader_atomic_float", true, &state.extensions.atomicFloat },
+      { "VK_KHR_shader_float16_int8", true, &state.extensions.float16Int8 },
+      { "VK_KHR_8bit_storage", true, &state.extensions.storage8 },
+      { "VK_KHR_16bit_storage", true, &state.extensions.storage16 }
     };
 
     uint32_t extensionCount = 0;
@@ -3228,6 +3234,9 @@ bool gpu_init(gpu_config* config) {
     VkPhysicalDeviceRayQueryFeaturesKHR rayQueryFeatures = { .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_RAY_QUERY_FEATURES_KHR };
     VkPhysicalDeviceHostImageCopyFeaturesEXT hostImageCopyFeatures = { .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_HOST_IMAGE_COPY_FEATURES_EXT };
     VkPhysicalDeviceShaderAtomicFloatFeaturesEXT atomicFloatFeatures = { .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_ATOMIC_FLOAT_FEATURES_EXT };
+    VkPhysicalDeviceShaderFloat16Int8Features float16int8 = { .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_SHADER_FLOAT16_INT8_FEATURES };
+    VkPhysicalDevice8BitStorageFeatures storage8 = { .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_8BIT_STORAGE_FEATURES };
+    VkPhysicalDevice16BitStorageFeatures storage16 = { .sType = VK_STRUCTURE_TYPE_PHYSICAL_DEVICE_16BIT_STORAGE_FEATURES };
 
     if (state.extensions.foveation) {
       CHAIN(supported, fragmentDensityMapFeatures);
@@ -3235,6 +3244,18 @@ bool gpu_init(gpu_config* config) {
 
     if (state.extensions.atomicFloat) {
       CHAIN(supported, atomicFloatFeatures);
+    }
+
+    if (state.extensions.float16Int8) {
+      CHAIN(supported, float16int8);
+    }
+
+    if (state.extensions.storage8) {
+      CHAIN(supported, storage8);
+    }
+
+    if (state.extensions.storage16) {
+      CHAIN(supported, storage16);
     }
 
     vkGetPhysicalDeviceFeatures2(state.adapter, &supported);
@@ -3315,6 +3336,18 @@ bool gpu_init(gpu_config* config) {
       CHAIN(enabled, atomicFloatFeatures);
     }
 
+    if (state.extensions.float16Int8) {
+      CHAIN(enabled, float16int8);
+    }
+
+    if (state.extensions.storage8) {
+      CHAIN(enabled, storage8);
+    }
+
+    if (state.extensions.storage16) {
+      CHAIN(enabled, storage16);
+    }
+
     if (config->features) {
       config->features->textureBC = enabled.features.textureCompressionBC;
       config->features->textureASTC = enabled.features.textureCompressionASTC_LDR;
@@ -3335,8 +3368,10 @@ bool gpu_init(gpu_config* config) {
       config->features->subgroupQuad = subgroupProperties.supportedOperations & VK_SUBGROUP_FEATURE_QUAD_BIT;
       config->features->float32AtomicAdd = atomicFloatFeatures.shaderBufferFloat32AtomicAdd && atomicFloatFeatures.shaderImageFloat32AtomicAdd;
       config->features->float64 = enabled.features.shaderFloat64;
+      config->features->float16 = float16int8.shaderFloat16 && storage16.storageBuffer16BitAccess;
       config->features->int64 = enabled.features.shaderInt64;
-      config->features->int16 = enabled.features.shaderInt16;
+      config->features->int16 = enabled.features.shaderInt16 && storage16.storageBuffer16BitAccess;
+      config->features->int8 = float16int8.shaderInt8 && storage8.storageBuffer8BitAccess;
 
       // Formats
       for (uint32_t i = 0; i < GPU_FORMAT_COUNT; i++) {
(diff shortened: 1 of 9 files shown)


SK83RJOSH:
On 1:
I believe `rgb8` is widely supported for vertex attributes. `rgb16` on the other hand, is not, but should we remove support altogether?

`storage8/16` support is fairly universal, so they still have wide applicability outside that use case, though `VK_BUFFER_USAGE_VERTEX_BUFFER_BIT | VK_BUFFER_USAGE_STORAGE_BUFFER_BIT` are used in tandem so maybe this is moot if this will block their use regardless.

On 2:
While `storage8/16` is near universal on desktop, `uniform8/16` is not. I suspect tethering the two would mean you don't get to benefit from these types in a lot of scenarios.

Perhaps `storage8/16` could be merged, and `uniform8/16` as well? I would presume these two features coincide with one another a great deal so that may be a win - but I'd want to get some numbers on that from somewhere.

Alternatively, would you be opposed to a bitmask for `Float`, `Int`, and `Storage` / `Uniform` support? Something like 1/2/4 signal 8/16/32 support? They could use the same underlying type / enum and we can avoid bit field support for now just to keep things simple.

```c
enum {
   GPU_FEATURE_SUPPORT_8 = (1 << 0),
   GPU_FEATURE_SUPPORT_16 = (1 << 1),
   GPU_FEATURE_SUPPORT_32 = (1 << 2),
   GPU_FEATURE_SUPPORT_64 = (1 << 3),
};

// All support 32-bit, and you would never query this, but nice from a "correctness" POV
uint8 integer = GPU_FEATURE_SUPPORT_32;
uint8 float = GPU_FEATURE_SUPPORT_32;
uint8 storage = GPU_FEATURE_SUPPORT_32;
uint8 uniform = GPU_FEATURE_SUPPORT_32;
```

Of course, with bit fields, this becomes _more_ compact and all four of these fit in 2 bytes.

src/core/gpu_web.c
```diff
@@ -750,6 +764,10 @@ bool gpu_pipeline_init_graphics(gpu_pipeline* pipeline, gpu_pipeline_info* info,
 
     for (uint32_t j = 0; j < info->vertex.attributeCount; j++) {
       if (info->vertex.attributes[j].buffer == i) {
+        if (info->vertex.attributes[j].offset % 4 != 0) {
+          return setError("WebGPU vertex attribute offset must be a multiple of 4");
```
bjornbytes: I don't think WebGPU requires this:

> attrib.offset is a multiple of the minimum of 4 and byteSize(attrib.[format]).

src/modules/graphics/graphics.c
```diff
@@ -9878,8 +9908,15 @@ static bool checkShaderFeatures(uint32_t* features, uint32_t count) {
...
+      case 4434: lovrCheck(state.features.uniform16, "GPU does not support shader feature #%d: %s", features[i], "16 bit uniform/storage buffers"); break;
+      case 4435: return lovrSetError("Shader uses unsupported feature #%d: %s", features[i], "16 bit push constants");
+      case 4436: return lovrSetError("Shader uses unsupported feature #%d: %s", features[i], "16 bit shader inputs/outputs");
       case 4437: return lovrSetError("Shader uses unsupported feature #%d: %s", features[i], "multigpu");
       case 4439: lovrCheck(state.limits.renderSize[2] > 1, "GPU does not support shader feature #%d: %s", features[i], "multiview"); break;
+      case 4448: lovrCheck(state.features.storage8, "GPU does not support shader feature #%d: %s", features[i], "8 bit storage buffers"); break;
+      case 4449: lovrCheck(state.features.uniform8, "GPU does not support shader feature #%d: %s", features[i], "8 bit uniform/storage buffers"); break;
+      case 4450: return lovrSetError("Shader uses unsupported feature #%d: %s", features[i], "8 bit push constants");
```
bjornbytes: I don't think it's worth explicitly naming the 16 bit push constants / 8 bit push constants features, since LÖVR doesn't support push constants and people aren't expected to be using them in their shaders (the shader will fail to compile if they use them, at least in graphics shaders, before this code can ever check the features).

src/core/gpu_web.c
SK83RJOSH: Ah right, this can probably go if we're dropping the vec3 types – was mostly to guard against someone using packed layout. Didn't put much thought into it besides assume minimum of 4 as alignment was *probably* correct. :)

src/modules/graphics/graphics.c
SK83RJOSH: Ah wasn't aware of that, happy to kill it in that case

src/modules/graphics/graphics.h
```diff
@@ -131,17 +131,25 @@ bool lovrGraphicsWait(void);
...
+  TYPE_I8,
+  TYPE_I8x2,
+  TYPE_I8x3,
   TYPE_I8x4,
+  TYPE_U8,
+  TYPE_U8x2,
+  TYPE_U8x3,
   TYPE_U8x4,
   TYPE_SN8x4,
   TYPE_UN8x4,
```
bjornbytes, line 147: It probably makes sense to also add `un8`, `un8x2`, `sn8`, `sn8x2` for consistency.  `sn16`, and `un16` could be supported as well, but it's not directly related to what was added in this PR.

src/core/gpu_vk.c
```diff
@@ -3315,6 +3336,18 @@ bool gpu_init(gpu_config* config) {
...
+    if (state.extensions.float16Int8) {
+      CHAIN(supported, float16int8);
+    }
+
+    if (state.extensions.storage8) {
+      CHAIN(supported, storage8);
+    }
+
+    if (state.extensions.storage16) {
+      CHAIN(supported, storage16);
```
bjornbytes: These should be `CHAIN(enabled, ...)`

bjornbytes:
One idea to simplify the graphics features a bit might be to document them as like:

> All the "type" features, like float16, int8, etc. indicate support for using the type as variables in shader functions and in storage buffers.  Uniform buffers and in/out don't support these types.

That way there can just be top-level `float16` and `int8` features, similar to WebGPU.  They indicate that you can use the types for most of the common uses, without bundling them with the other features that have less-broad support.

SK83RJOSH:
> All the "type" features, like float16, int8, etc. indicate support for using the type as variables in shader functions and in storage buffers.  Uniform buffers and in/out don't support these types.

I'm okay with this, but for clarity, this feature in WGPU essentially checks for `storageBuffer16BitAccess` + `uniformAndStorageBuffer16BitAccess`.

So I guess for true parity, we may as well do the same. Bit of a shame, since I think this limits 100% support to Nvidia RTX and forward or ~8 years vs. ~11 years.

SK83RJOSH:
All feedback addressed, only change worth noting, is I made int16 align with the rest here, which is a breaking change. I don't know how many people use lovr on mobile, but support for i16 types _and_ storage/uniform buffer support is probably the worst on those devices.

I presume this wouldn't be the case on stuff like the Quest and other standalone headsets though. So it may be worth regression testing if you've got anything to test on (I've a Quest 3 lying around, but no projects to test with).

Granted, I suspect not many people were using that anyway.

bjornbytes:
It's fine to make int16 a little more restrictive.

Since we get to decide between:

1. Make `float16` etc. also indicate support for uniform buffers.  Match WebGPU, but feature is unsupported more.
2. Make `float16` etc. only indicate support for storage buffers, and document it as not working for uniform buffers.

I think I'd probably pick (2) here?  It seems better to advertise the feature more (esp. on Quest) rather than strictly follow WebGPU's semantics.  Either way, it's definitely niche.

src/core/gpu_vk.c
```diff
@@ -4349,9 +4384,17 @@ static VkFormat convertFormat(gpu_texture_format format, int colorspace) {
...
 static VkFormat convertAttributeType(gpu_attribute_type type) {
   static const VkFormat types[] = {
+    [GPU_TYPE_I8] = VK_FORMAT_R8_SINT,
+    [GPU_TYPE_I8x2] = VK_FORMAT_R8G8_SINT,
     [GPU_TYPE_I8x4] = VK_FORMAT_R8G8B8A8_SINT,
+    [GPU_TYPE_U8] = VK_FORMAT_R8_UINT,
+    [GPU_TYPE_U8x2] = VK_FORMAT_R8G8_UINT,
     [GPU_TYPE_U8x4] = VK_FORMAT_R8G8B8A8_UINT,
+    [GPU_TYPE_SN8] = VK_FORMAT_R8_UNORM,
+    [GPU_TYPE_SN8x2] = VK_FORMAT_R8G8_UNORM,
```
bjornbytes: Small typo here, should be `SNORM` instead of `UNORM`.  Hm, also SN formats are duplicated

bjornbytes:
`l_lovrGraphicsGetFeatures` in `l_graphics.c` should also add the new `float16` and `int8` features to the table.

src/core/gpu_vk.c
SK83RJOSH: Whoops, yeah that's my bad, I've been copy pasting and multi-line editing these to make doing the work in triplicate a bit less tedious. :)

SK83RJOSH:
> I think I'd probably pick (2) here? It seems better to advertise the feature more (esp. on Quest) rather than strictly follow WebGPU's semantics. Either way, it's definitely niche.

Here's what support looks like right now:
#### Oculus Go:
shaderInt16
#### Oculus Quest 2 / Pro:
shaderInt8, shaderInt16, shaderFloat16, storageBuffer16BitAccess
#### Oculus Quest 3 / 3S:
shaderInt8, shaderInt16, shaderInt64, shaderFloat16, storageBuffer8BitAccess, storageBuffer16BitAccess
#### Apple Vision Pro (M2):
shaderInt8, shaderInt16, shaderFloat16, storageBuffer8BitAccess, uniformAndStorageBuffer8BitAccess, storageBuffer16BitAccess, uniformAndStorageBuffer16BitAccess
#### Apple Vision Pro (M5):
shaderInt8, shaderInt16, shaderInt64, shaderFloat16, storageBuffer8BitAccess, uniformAndStorageBuffer8BitAccess, storageBuffer16BitAccess, uniformAndStorageBuffer16BitAccess

Interestingly Qualcomm GPUs do not support uniformAndStorageBuffer at all - so if mobile is a big area for y'all I think that seals the deal as the minimum target.

src/core/spv.c
```diff
@@ -733,7 +733,13 @@ static spv_result spv_parse_field(spv_context* spv, const uint32_t* word, spv_fi
     }
   }
 
-  if (OP_CODE(word) == 22 && word[2] == 32) { // OpTypeFloat
+  if (OP_CODE(word) == 22 && word[2] == 16) { // OpTypeFloat 16-bit
+    if (columnCount == 1 && componentCount <= 4 && (componentCount & (componentCount - 1)) == 0) {
```
SK83RJOSH: In hindsight, I probably could have checked `componentCount != 3` here + just used `+ componentCount / 2`

Not sure if it's worth fixing this up, or not, really. Easier to read for the less bit fiddle minded. :)

SK83RJOSH:
I don't think you need to run checks for this one, but decided to resolve that / code style issues in `spv.c`, and I updated the errors in `graphics.c` to handle storage/uniform correctly.

Think this is ready to merge, so I'll restrain myself from further code golfing. :)

bjornbytes:
Looks good, thanks for the PR!

bjornbytes:
(Added support for reading back the new formats in `Buffer:getData` + friends, and golfed the spir-v parsing a bit more)

PR #1003, opened 2026-08-14
Add Meta full-body support to lovr.headset.getSkeleton('body')

jsepulveda-nvidia:
Previously getSkeleton('body') only ever queried the BD tracker (XR_BD_body_tracking, 24 fixed joints), ignoring the FB/Meta tracker entirely even when XR_META_body_tracking_full_body was active. It now prefers the Meta full-body tracker (84 joints, positions + orientations) when active, falling back to BD otherwise, and reports the actual joint count and source ("bd"/"meta") back to Lua instead of assuming a fixed 24.

### DIFF
--- a/src/api/l_headset.c
+++ b/src/api/l_headset.c
@@ -149,6 +149,8 @@ static int l_lovrHeadsetGetFeatures(lua_State* L) {
   lua_pushboolean(L, features.handTracking), lua_setfield(L, -2, "handTracking");
   lua_pushboolean(L, features.handTrackingElbow), lua_setfield(L, -2, "handTrackingElbow");
   lua_pushboolean(L, features.bodyTracking), lua_setfield(L, -2, "bodyTracking");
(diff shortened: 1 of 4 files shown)


bjornbytes:
Thanks for the PR.  I don't think it's very good to return 3 different kinds of skeletons from `lovr.headset.getSkeleton('body')`.  This forces people to look at the headset/extension in use and remap joints themselves, which is too unfriendly compared to the rest of the API.

Instead, we probably need to massage the underlying data into a consistent format, either by picking a lowest-common-denominator skeleton or returning empty data for unsupported joints.  Alternatively, we could continue the current strategy of doing nothing and waiting for an EXT extension :smile:

It's probably worth opening an issue to discuss what the universal body skeleton should look like.  I'd be interested in contributing to it when I have time.

bjornbytes:
There's also this old body tracking issue that could get resurrected https://github.com/bjornbytes/lovr/issues/840

